PR #139, opened 2012-07-12
New feature: extra_headers argument for ngx.location.capture

vadim-pavlov:
Hello, this feature allows to set custom http headers for sub request using an 'extra_headers' argument.
I would appreciate it if you could run tests and merge this feature with your branch :)

### DIFF
--- a/src/ngx_http_lua_subrequest.c
+++ b/src/ngx_http_lua_subrequest.c
@@ -24,32 +24,140 @@ ngx_str_t  ngx_http_lua_delete_method =
 ngx_str_t  ngx_http_lua_options_method =
         ngx_http_lua_method_name("OPTIONS");
 
-static ngx_str_t  ngx_http_lua_content_length_header_key =
+static ngx_str_t ngx_http_lua_content_length_header_key =
         ngx_string("Content-Length");
 
-static ngx_int_t ngx_http_lua_is_content_length_string(ngx_str_t *str);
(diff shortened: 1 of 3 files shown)


surfire91:
Hi, guys,
The feature is useful for me, but it seems like that have not be done?

agentzh:
@surfire91 Not yet though it's been on my TODO list.

BTW, there exists work-arounds:

1. use `ngx.req.set_header` and `ngx.req.clear_header` in the parent request's Lua code, and
2. use `proxy_set_header` and nginx variables as the header names or values in the subrequest location if `ngx_proxy` is used and pass values from the parent to the subrequest via nginx variables or something like that.

Because of the existence of these work-arounds, the priority of this PR has been relatively low. But still I'd like to see this to get reviewed and eventually merged.

surfire91:
@agentzh Thanks for reply.
I just use the `ngx.req.set`_header on my current work, but I think  the TODO **new feature** is more flexible.
Excepting merged.

agentzh:
@surfire91 Yeah, I know :)

RoryCrispin:
It'd be great to see this feature merged - the workaround seems clumsy especially for use cases such as making an upstream json POST: 
copy the original content-type header into a temp variable
set the content type text/json header
make request
set the content type header back to it's original value

Thanks

Naidile-P-N:
Hi, Is this feature available now ? Would be of great help

tanguilp:
Would still be useful in 2020 :)

PR #256, opened 2013-07-03
API change: change setkeepalive to gotopool

guanlan:


doujiang24:
it's not a good choice to change it now since `setkeepalive` is already used for a long time.
so, closing it.

PR #265, opened 2013-08-02
New Feature: getlifetime API

guanlan:


zhuizhuhaomeng:
We need to re-implement this feature via ffi if there are someone interested in it

PR #289, opened 2013-10-07
Subrequest API - method can be passed as a string

aviramc:
In the ngx.location.capture and ngx.localtion.capture_multi APIs, the method option can now also be a string. If the string is an unknown method, the subrequest's method field is set to NGX_HTTP_UNKNOWN, and the method_name field is set to the given method name. The previous numeric method still works.


aviramc:
Will open a new tidy pull request

Issue #241, opened 2013-05-27
Can't read/write to the same cosocket from different threads


AdallomRoy:
It's impossible with the current implementation to read from thread 1 and write from thread 2 to the same socket.
This causes issues in tunnel cases - if I wanted to create a tunnel between 2 connections I'd run 2 threads, 1 reading from socket A and writing to socket B, the other reading from socket B and writing to socket A - which is currently impossible ("socket busy") - even though this supported by nginx itself.
This is caused by the fact that there isn't a separation of read & write state for each socket -
ngx_http_lua_co_ctx_t           *co_ctx;
and
unsigned                         waiting:1;
are for both read & write operations - They should probably be separated.

Issue #178, opened 2012-11-17
SSL/TLS support for cosocket API?


Doridian:
So far all I needed to do SSL with has been HTTPS so I settled with using ngx.location.capture and internal proxy locations.
But now I have come accross an issue.
I need to send E-Mail over a server which requires TLS (Amazon SES), which seems undoable at the moment as LuaSec does not work (and prolly would ruin cosocket's asynchronity anyway).
Is there any possible way I could establish communication with the server?

PR #290, opened 2013-10-07
Major cosockets patch

aviramc:
This includes the following changes:

 - Added support for SSL sockets (includes the verification of the remote end's certificate). Related to issue #178
 - Setting different timeouts for receive and send.
 - Now read and write can be done from different Lua threads. Relates to issue #241.
 - Added fake_close so that we won't be able to read again from the client socket after it is closed.
 - receive API - added the option 'bsd_read', which enables to receive the maximum number of bytes given (like the BSD API, not LuaSocket API).


aviramc:
Will reopen new pull requests for each of the required features

PR #350, opened 2014-03-22
add a ngx.req.is_internal() api.

ruoshan:
ngx.req.is_internal() return `true` when current request is an internal
request, say a request been rewrited.


agentzh:
@ruoshan Sorry for the long delay on my side. I'm looking into this patch.

agentzh:
@ruoshan Just committed slightly modified version of your patch to git master. Thank you for your contribution and I'm sorry for the long delay on my side.

PR #420, opened 2014-09-17
update example code:ngx.shared.cats.set

toontong:
cats.set --> cats:set


zhuizhuhaomeng (commented):
the origin is right.
if want to change to cat: then it should be  cats:set("Marry", "it is a nice cat!")
so reject the PR

doujiang24:
yeah, it's an example for `The first argument to this method must be the dictionary object itself` in the doc.
so, closing it.

PR #423, opened 2014-09-22
Exposing ngx_decode_base64url and ngx_encode_base64url to Lua scripts.

ghost:


felixbuenemann:
I think it's strange that core provides base64 encode/decode but not the urlsafe variants.

I also think putting them somewhere else would be confusing.

If I look at docs of resty.core, I would ask myself: "Why are there no url variants?" and if looking at resty.string I would ask myself: "Why does it only implement url safe variants?"

felixbuenemann:
@agentzh Would you accept a patch that adds `urlsafe` as a boolean param to the existing `encode_base64` and `decode_base64` functions? That way no new methods would have to be added.

I have [patched 1.9.15.1rc0](https://gist.github.com/bf300482a2df89970c5b5d7e3191b464) with urlsafe support and it gives a nice performance boost over the gsub code proposed by @bungle.

I have a location that uses `access_by_lua_file` to verifiy a HMAC-SHA1 signature passed in the query string as urlsafe Base64 and after applying the patch I'm seeing a boost from 29k rps to 33k rps.

bungle:
@felixbuenemann I have also `lua-resty-nettle` library that thas url-safe base64 encoding:
https://github.com/bungle/lua-resty-nettle/blob/master/lib/resty/nettle/base64.lua

Urlsafe Base64 encoding is BTW specified in RFC 4648:
https://www.ietf.org/rfc/rfc4648.txt

alubbe:
I also second having a base64url function either in core or resty-string. It is a use case that does pop up often enough to be solved by something "official" and having a C implementation would help greatly with performance.
@bungle In my benchmarks (which of course may be wrong), resty-nettle's base64url functions are 3x slower than the one in resty-core and 1.5x slower than using the core functions without padding and two consecutive ngx.re.gsub's to turn base64 to base64urlsafe.

bungle:
@alubbe, yeah, I haven't personally benchmarked them, and I don't know if they even do the same thing.

bungle:
If you want absolute best performance, I think the whole Base64 thing should probably be implemented with LuaJIT optimized code without going to C.

bungle:
This looks pretty fast as well:
https://github.com/aklomp/base64

alubbe:
It does indeed. In terms of "pure" luajit, I also found (but haven't benchmarked) https://github.com/jsolman/luajit-mime-base64/blob/master/mime_base64.lua.
Maybe we could allow different 64-length alphabets here https://github.com/aklomp/base64/blob/master/lib/lib.c#L16, write a luajit ffi wrapper and be good to go?

bungle:
I also found out this:
https://github.com/kengonakajima/luvit-base64/issues/1

But well, I still think OpenResty should expose Nginx's `ngx_decode_base64url` to Lua land. At least the FFI version of it. I mean why not really. For the best performance we can still look at other implementations that can be used where it matters.

alubbe:
Actually, I agree. And I think that should be part of either core or resty-string.
It might then be interesting to ask the nginx dev team what they think of using a simd-based implementation for base64.

agentzh:
Well, patches for lua-resty-string are welcome :) I don't mind adding simple pure C wrappers in the ngx_lua core for FFI.

felixbuenemann:
@agentzh I'm not sure I understand you correctly, because you are talking about lua-resty-string and ngx_lua core in the same sentence.

In the patch I posted above I patched lua-resty-core and the code re-uses functions that already exist in lua-resty-core. So, if I understand you correctly you would prefer these new wrappers to be added to lua-resty-string and not to lua-resty-core?

bungle:
You may implement this binding without C but the binding will be a bit dirty as you have to fiddle with all those `ngx_str_t´s. But it is doable. But it differs a lot from that normal base64 a lot, for no obvious reason IMHO. I'm also against polluting Lua Nginx Module too much, but this is something that should be in there, or else you need to have C code in say lua-resty-string.

BTW, meanwhile I did wrote test bindings to this:
https://github.com/aklomp/base64

The bindings are here (there is only one premade output buffer that is 8192 bytes, if you need bigger, please change that – longer ones are created as needed):
https://github.com/bungle/lua-resty-libbase64/blob/master/lib/resty/libbase64.lua

@alubbe, if you do benchmark, please let me know how it went (this is just a quick and dirty example).

felixbuenemann:
@bungle I couldn't get the base64 library to link on my mac, cause they're calling `ld` in a way that doesn't seem to work on Mac OS X.

bungle:
@alubbe 

I did change `Makefile` a little bit:

```diff
-lib/libbase64.o: $(OBJS)
-       $(LD) --relocatable -o $@ $^
+lib/libbase64.so: $(OBJS)
+       $(LD) -dylib  -o $@ $^
```

And then (you can use OSX default make as well, here I'm using brew version of it):

```bash
AVX2_CFLAGS=-mavx2 SSSE3_CFLAGS=-mssse3 gnumake lib/libbase64.so
```

And then:

```bash
cp libbase64.so /usr/local/lib/libbase64.dylib
```

alubbe:
I cannot get it to work either, I keep getting 
```
../base64/lib/libbase64.o: only ET_DYN and ET_EXEC can be loaded
```
Using your patch and running make I get
```
$ make lib/libbase64.so 
cc -std=c99 -O3 -Wall -Wextra -pedantic -o lib/lib.o -c lib/lib.c
cc -std=c99 -O3 -Wall -Wextra -pedantic  -o lib/codec_avx2.o -c lib/codec_avx2.c
cc -std=c99 -O3 -Wall -Wextra -pedantic -o lib/codec_choose.o -c lib/codec_choose.c
cc -std=c99 -O3 -Wall -Wextra -pedantic  -o lib/codec_neon32.o -c lib/codec_neon32.c
cc -std=c99 -O3 -Wall -Wextra -pedantic  -o lib/codec_neon64.o -c lib/codec_neon64.c
cc -std=c99 -O3 -Wall -Wextra -pedantic -o lib/codec_plain.o -c lib/codec_plain.c
cc -std=c99 -O3 -Wall -Wextra -pedantic  -o lib/codec_ssse3.o -c lib/codec_ssse3.c
ld -dylib  -o lib/libbase64.so lib/lib.o lib/codec_avx2.o lib/codec_choose.o lib/codec_neon32.o lib/codec_neon64.o lib/codec_plain.o lib/codec_ssse3.o
ld: warning: cannot find entry symbol _start; defaulting to 00000000004000f0
lib/lib.o: In function `base64_encode':
lib.c:(.text+0x21a): undefined reference to `__stack_chk_fail'
lib/lib.o: In function `base64_decode':
lib.c:(.text+0x2b7): undefined reference to `__stack_chk_fail'
Makefile:49: recipe for target 'lib/libbase64.so' failed
make: *** [lib/libbase64.so] Error 1
```
I am on 64-bit ubuntu 16.04.

alubbe:
Just benchmarked https://github.com/jsolman/luajit-mime-base64/blob/master/mime_base64.lua, it runs a lot slower than ngx.b64.
```
luajit b64 (not urlsafe) runtime: 151400
ngx.b64 (not urlsafe) runtime: 28969
```
@bungle If you can help me get this new library to work/compile, I'd love to see how it fares.

felixbuenemann:
@bungle Thanks, I got it to work after some fiddling. The system clang miscompiled the avx2 codec, but gcc 5.3 worked fine and far outperformed clang-omp 3.5 in OpenMP mode (added `-liomp5 -lSystem`).

In my case it doesn't really give a performance boost over the lua core C version, because my payload is only 27 chars (hmac-sha1 urlencoded w/o padding). My main optimization room would be to use a shared hmac digest context instead of creating it on each hit (I'm using api-gateway-hmac atm.).

This discussion has become a bit off-topic though. The intent was to enable access to the native base64url en/decoder in nginx, not add the fastest possible base64 implementation, which should probably live in its own library.

felixbuenemann:
@alubbe You probably want to use `gcc -shared` instead of `ld -dylib` and add `-fpic` to CFLAGS.

alubbe:
@felixbuenemann That got it to work, it does indeed run a little faster than ngx's own b64 implementation, but not by too much. 
```
luajit b64 (not urlsafe) runtime: 117196
fastb64 (not urlsafe) runtime: 28387
ngx.b64 (not urlsafe) runtime: 35658
```
I agree that we've gone off on a tangent, so let's return to the topic at hand.
@agentzh Could we convince you to get base64url either in core or resty.string?

alubbe:
> Well, patches for lua-resty-string are welcome :) I don't mind adding simple pure C wrappers in the ngx_lua core for FFI.

@agentzh do you have lua-nginx-module or resty.string in mind for this simple wrapper? How does this PR compare?

bungle:
@agentzh, is it okay to remove Lua C API versions from here and only add Lua FFI version C wrappers to `lua-nginx-module`, and then add Lua APIs for them to either to `lua-resty-string` or `lua-resty-core`?

E.g. what's proposed in this comment:
https://github.com/openresty/lua-nginx-module/pull/423#issuecomment-65509507

alubbe:
Any news on this? https://github.com/openresty/lua-nginx-module/pull/423#issuecomment-220127818 sounds like a good way forward.

russmac:
I also need this.

PR #450, opened 2015-01-15
Wait for fd API

daurnimator:
Following on from https://groups.google.com/d/msg/openresty-en/T81159uj0y4/ZDckUrpUH30J

The short term goal here is a simple API that allows an advanced user to yield their request until an fd polls ready, or a timeout elapses. An fd polling ready is a primitive used across the majority of unix APIs, and is commonly exposed by (e.g.) database client libraries.

The single function added is: `ngx.wait(fd, events, timeout)`
- `fd` is the numeric file descriptor to wait on
- `events` is a string containing "r" and/or "w" indicating to wait for readability vs writability
- `timeout` is the maximum amount of time (in seconds) to wait.


agentzh:
it did not address the problem i pointed out earlier :)
that is, it requires unnecessary epoll_ctl syscalls when ET is used.
better make it stateful rather than stateless.

agentzh:
Regarding your offline comments, this is not a premature optimization because system calls are almost always expensive, not to mention the frequency of the wait() calls in reality. Besides, we are designing an API here and this API should make room for important optimizations (even in the future) because one we add an API, we have to maintain backward compatibility.

daurnimator:
> it requires unnecessary epoll_ctl syscalls when ET is used.

I feel like that would be premature optimisation.

This is a first step to make a new style of programming *possible* (i.e. use cqueues).
If it works out well, I hope that a cqueues-like object can provide that stateful api you're talking about
(a cqueues object is essentially an object that contains an epoll fd and a self-pipe to wake itself up)

daurnimator:
> this is not a premature optimization because system calls are almost always expensive, not to mention the frequency of the wait() calls in reality. Besides, we are designing an API here and this API should make room for important optimizations (even in the future) because one we add an API, we have to maintain backward compatibility.

If everything works out, this API would be kept for one-shots, where the stateful abstraction would not be helpful. One-shots are quite common, and enough as in a applications an RPC call will be retrieved in a single packet.
A new api that is stateful would be added in parallel.

agentzh:
BTW, I'm against merging quick hacks just to make certain things possible because we have to eventually do it right and we'll almost always pay a bigger price when fixing hacks. Also, I'm very serious about adding new APIs to the ngx_lua core because I'd like to keep the core minimal :)

agentzh:
Having said that, I'd love to see this pull request evolves (here or somewhere else) to a state where it is good to merge.

daurnimator:
> BTW, I'm against merging quick hacks just to make certain things possible because we have to eventually do it right and we'll almost always pay a bigger price when fixing hacks.

An extra function useful in some (uncommon) circumstances isn't a quick hack IMO.

> Also, I'm very serious about adding new APIs to the ngx_lua core because I'd like to keep the core minimal :)

One of the reasons I would like this to be added in this state is that the functionality provided cannot be done in an external module.
If you could provide a way to have this code outside the core lua-nginx-module I'd be happy to keep things seperate and maintain it myself.
I'd love to start writing+distributing modules that can rely on this type of function.

> I'd love to see this pull request evolves

Would you like me to rebase onto master as you work on other things?

> a state where it is good to merge.

Could you define what that entails?

agentzh:
> Could you define what that entails?

Pass my review :)

agentzh:
@daurnimator Instead of making it a stateless function, I suggest you turn it into something stateful, like the standard Lua io module's file objects.

agentzh:
@daurnimator In addition, we may want to make room for the support of waiting on multiple fds at the same time.

agentzh:
@daurnimator Also, please strictly follow the nginx coding style in your patch. For example, 2 blank lines are used to separate function definitions.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+                return NGX_DONE;
+            }
+            return rc;
+    }
+}
+
+static void ngx_http_lua_fd_rev_handler(ngx_event_t *ev) {
+    ngx_connection_t *conn = ev->data;
+    ngx_http_lua_udata_t *u = conn->data;
+    ngx_http_request_t *r = u->request;
```
agentzh: Please avoid using such complicated initializers becuse it's against the nginx coding style.

Also, please always align up the variables in their definitions.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+#include "ngx_http.h" /* ngx_http_request_t, ngx_http_run_posted_requests */
+#include "ngx_http_lua_common.h" /* ngx_http_lua_module, ngx_http_lua_co_ctx_t */
+
+#include "ngx_http_lua_util.h" /* ngx_http_lua_get_lua_vm, ngx_http_lua_run_thread, ngx_http_lua_finalize_request */
+#include "ngx_http_lua_contentby.h" /* ngx_http_lua_content_wev_handler */
+
+typedef struct {
+    ngx_http_request_t *request;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_connection_t *conn;
```
agentzh: Style: align up the member variable names.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+} ngx_http_lua_udata_t;
+
```
agentzh: Style: two blank lines required here.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+            return ngx_http_lua_run_posted_threads(r->connection, vm, r, ctx);
+        default: /* NGX_ERROR: coroutine failed */
+            if (ctx->entered_content_phase) {
+                ngx_http_lua_finalize_request(r, rc);
+                return NGX_DONE;
+            }
+            return rc;
+    }
+}
+
```
agentzh: Ditto and other places.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+static void ngx_http_lua_fd_rev_handler(ngx_event_t *ev) {
```
agentzh: Please follow the nginx coding style strictly here: we put the function name at column 1.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+            ngx_del_timer(u->conn->read);
+        }
+
+        if (u->conn->fd != -1) {
+            /* remove from mainloop; do not pass CLOSE_SOCKET flag */
+            ngx_del_conn(u->conn, 0);
+        }
+
+        /* delete any pending but not handled events */
+#if (NGX_THREADS)
```
agentzh: Please remove the `NGX_THREADS` things because they're obsolete and never worked.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+    if ((r = ngx_http_lua_get_req(L)) == NULL) {
+        return luaL_error(L, "no request found");
+    }
+    if ((ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module)) == NULL) {
+        return luaL_error(L, "no request ctx found");
+    }
+    if ((co_ctx = ctx->cur_co_ctx) == NULL) {
+        return luaL_error(L, "no co ctx found");
+    }
+    if ((u = calloc(1, sizeof(ngx_http_lua_udata_t))) == NULL) {
```
agentzh: Please avoid using the libc allocator functions directly. Use the nginx wrappers instead.

daurnimator:
> @daurnimator Instead of making it a stateless function, I suggest you turn it into something stateful, like the standard Lua io module's file objects.

This seems like a much more complex api that I don't expect to get right on the first go.... 

What bits of state would you like to keep? the connection object seems to be the only thing possible?
And it comes out of an nginx internal pool anyway....

> @daurnimator In addition, we may want to make room for the support of waiting on multiple fds at the same time.

This is the end goal: but I was going to provide it via `epoll`, `kqueue`, `ports`, etc. in a library.
These allow for a single file descriptor to represent/poll many.

> @daurnimator Also, please strictly follow the nginx coding style in your patch. For example, 2 blank lines are used to separate function definitions.

I don't suppose there is an 'astyle' incarnation that will make it suit your style?

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+        return luaL_error(L, "no mem");
+    }
+    /* always create a connection object (even if fd is < 0) */
+    if ((u->conn = ngx_get_connection(fd, r->connection->log)) == NULL) {
```
agentzh: Calling `ngx_get_connection` and `ngx_free_connection` upon every `wait()` is also too expensive IMHO.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+    ngx_http_lua_udata_t *u = conn->data;
+    ngx_http_request_t *r = u->request;
+    ngx_http_lua_co_ctx_t *co_ctx = u->co_ctx;
+    ngx_http_lua_ctx_t *ctx;
+
+
+    ngx_http_lua_cleanup_pending_operation(co_ctx);
+    ev = NULL, conn = NULL, u = NULL; /* now invalidated */
+
+    if ((ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module)) != NULL) {
```
agentzh: Style: better avoid using the side-effect of C assignments according to nginx's coding style.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+static void ngx_http_lua_fd_rev_handler(ngx_event_t *ev) {
+    ngx_connection_t *conn = ev->data;
+    ngx_http_lua_udata_t *u = conn->data;
+    ngx_http_request_t *r = u->request;
+    ngx_http_lua_co_ctx_t *co_ctx = u->co_ctx;
+    ngx_http_lua_ctx_t *ctx;
+
+
+    ngx_http_lua_cleanup_pending_operation(co_ctx);
+    ev = NULL, conn = NULL, u = NULL; /* now invalidated */
```
agentzh: Style: one assignment per line please.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+    ngx_http_request_t *request;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_connection_t *conn;
+} ngx_http_lua_udata_t;
+
+static ngx_int_t ngx_http_lua_fd_resume_request(ngx_http_request_t *r) {
+    ngx_http_lua_ctx_t *ctx;
+    lua_State *vm;
+    ngx_int_t rc;
+    if ((ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module)) == NULL) {
```
agentzh: Please use blank lines wherever appropriate.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+    co_ctx->data = NULL;
+}
+
+static int ngx_http_lua_fd_wait(lua_State *L) {
+    ngx_http_request_t *r;
+    ngx_http_lua_ctx_t *ctx;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_http_lua_udata_t *u;
+
+    ngx_socket_t fd = luaL_optint(L, 1, -1); /* -1 is invalid fd */
```
agentzh: Style: no function calls in the initializer.

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,221 @@
...
+    const char *events = luaL_optstring(L, 2, "");
+    int poll_mask = 0;
+    double timeout = luaL_optnumber(L, 3, HUGE_VAL); /* default to infinite timeout */
+    while (*events) {
+        if (*events == 'r')
```
agentzh: Style: always use `{}` with `if` statements. This is required by the nginx coding style.

agentzh:
Also, among other things, please add corresponding unit tests to the test suite (but I'm afraid the current API deserves a redesign, so better get the API straight first).

daurnimator:
^^ pushed fixes for your minor comments

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,229 @@
...
+        }
+#else
+        if (u->conn->read->prev) {
+            ngx_delete_posted_event(u->conn->read);
+        }
+        if (u->conn->write->prev) {
+            ngx_delete_posted_event(u->conn->write);
+        }
+#endif
+        /* not sure what this line does, the 0 means non-reusable */
```
daurnimator: @agentzh are you able to explain what this does? I pulled it out of [ngx_close_connection](http://lxr.nginx.org/source/src/core/ngx_connection.c#0969) but maybe I should have marked it reusable? what are the criteria for this?

agentzh: In this case, you should not make it reusable. "reusable" means that the connection can be grabbed (and closed) immediately when nginx is running out of the `worker_connections` limit configured in `nginx.conf`. See the C functions `ngx_drain_connections` and `ngx_reusable_connection` in the nginx core for more details.

daurnimator: thankyou for the description. nginx source is sorely lacking in documentation :(

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    u->request = r;
+
+    u->co_ctx = co_ctx;
+    co_ctx->cleanup = (ngx_http_cleanup_pt)&ngx_http_lua_fd_cleanup;
+    co_ctx->data = u;
+
+    if (timeout != HUGE_VAL) { /* no point adding an infinite timeout */
+        ngx_add_timer(u->conn->read, (ngx_msec_t)(timeout * 1000));
```
daurnimator: Is it okay to use the timer in `u->conn->read` even `ngx_handle_read_event` has not been called with it?

agentzh: @daurnimator Better only use the timer in `u->conn->read` when `u->conn` is actively waiting for the next read event.

daurnimator: > when u->conn is actively waiting for the next read event.

But can you still use `conn->read`'s timer object when the connection is only waiting for a write event?
Reading through the source it seems okay to do.

Otherwise I'll need to add this patch:
```diff
diff --git a/src/ngx_http_lua_fd.c b/src/ngx_http_lua_fd.c
index d5dc6db..f46aa36 100644
--- a/src/ngx_http_lua_fd.c
+++ b/src/ngx_http_lua_fd.c
@@ -99,6 +99,9 @@ ngx_http_lua_fd_cleanup(ngx_http_lua_co_ctx_t *co_ctx)
         if (u->conn->read->timer_set) {
             ngx_del_timer(u->conn->read);
         }
+        if (u->conn->write->timer_set) {
+            ngx_del_timer(u->conn->read);
+        }
 
         if (u->conn->fd != -1) {
             /* remove from mainloop; do not pass CLOSE_SOCKET flag */
@@ -207,7 +210,7 @@ ngx_http_lua_fd_wait(lua_State *L)
     co_ctx->data = u;
 
     if (timeout != HUGE_VAL) { /* no point adding an infinite timeout */
-        ngx_add_timer(u->conn->read, (ngx_msec_t)(timeout * 1000));
+        ngx_add_timer((poll_mask & POLLIN) ? u->conn->read : u->conn->write, (ngx_msec_t)(timeout * 1000));
     }
 
     /* make sure nginx knows what to do next */
```

agentzh: Hello!

On Thu, Feb 5, 2015 at 12:33 PM, daurnimator wrote:
> But can you still use conn->read's timer object when the connection is only waiting for a write event?

That's confusing and error prone. Don't do that.

> Reading through the source it seems okay to do.
>

Even though you can do it, you shouldn't do that.

-agentzh

daurnimator: So I should commit that patch?
I figured the extra logic was just as confusing, but I'm happy to add it if it makes you happy :P

daurnimator:
Have you had any more thoughts about this? Should I rebase the PR?

agentzh:
@daurnimator Though I admit this PR's goal is meaningful this PR still requires quite some treatment before it can be merged.

lordnynex:
@daurnimator as a side note, I greatly appreciate your efforts on this PR. If merged, this PR will drastically change the openresty ecosystem. Personally I've been waiting on zmq abilities for quite awhile. I certainly hope you keep up the good work.

lordnynex:
@agentzh Do you have any open notes or comments on this PR to help keep dev moving forward? At this point, the efforts can not even be sponsored because the criteria for PR acceptance is unclear.

agentzh:
@lordnynex I agree. Yeah, I have something in my mind. I'll try drafting an alternative API design for open discussions soon (sorry, I've been too busy with those ngx_stream_* module development and new *_by_lua directives in ngx_lua atm). Thanks for asking.

daurnimator:
I just rebased this PR.

daurnimator:
@agentzh any progress?

agentzh:
@daurnimator Not yet, sorry. Been busy with the upcoming OpenResty 1.9.7.1 release.

bungle:
I talked shortly with @daurnimator on IRC about this pull request. I have been following this, because I have been looking at building some LuaJIT bindings and Lua APIs for things like nanomsg (using it just as an example), and maybe others as well. Right now, to do it in a non-blocking way, you have to kinda start from the scratch (and then have to maintain the lower level network things as well). With this patch it is obvious that I could leverage the existing work done in the official nanomsg project, and I could just write LuaJIT bindings to that, and a nice Lua API on top of it. I think that this way we could grow OpenResty ecosystem to a new heights. And we could also leverage many existing solutions as well – in that sense this adds somewhat a compatibility layer as well (or at least makes it easier).

I remember that on OpenResty Con I was asked about the problems of making OpenResty to talk with ZooKeeper. A quick look at ZooKeeper's C-client shows that this patch could make writing bindings to ZooKeeper quite straightforward.

My knowledge on the lower level things like merging this patch is limited, but I would like to have discussion going on here. I'd like to hear about the concerns @agentzh is having about this, and the ideas for the alternative design. I understand that this is a rather low level library, and maybe somewhat dangerous in wrong hands (?), but it is quite simple as well. Moving it below `ngx` namespace somewhere like `ngx.fd.wait(...)` could be a better place for it (?).

Not trying to push or anything, well maybe I am ;-). I agree that we shall get this as correct as possible before releasing. If the API is alright we can possible change the internals later on. But before this is decided, some things cannot be moved forward (like building the actual libraries on top of this).

agentzh:
@bungle Thanks for following this discussion.

Basically I'm thinking about a more OO-like API for this. Maybe something like this?

```lua
local connection = require "ngx.connection"
local c = connection:new(fd, options)
local ok, err = c:wait_for_read(timeout)
local ok, err = c:wait_for_write(timeout)
local ok, err = c:wait_for_read_write(timeout)
```

I suggest putting the `ngx.connection` module in lua-resty-core and uses LuaJIT's FFI to call into the C API (or ABI) provided by the ngx_lua core. In the same spirit of the ngx.semaphore API in development:

https://github.com/openresty/lua-nginx-module/pull/584

https://github.com/openresty/lua-resty-core/pull/5

Just for the reference :)

To be honest, I myself am also eager to have this in OpenResty now :)

src/ngx_http_lua_fd.c
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /* make sure nginx knows what to do next */
```
daurnimator: Is this bit required? I was browsing through the semaphore pull request and they don't do it.

agentzh: @daurnimator This just depends on what you are trying to do. The cosocket API needs to maintain a lot of its own state data which is inappropriate to get included in the co_ctx_t struct itself while the semaphore API can reuse existing fields in the co_ctx_t struct.

daurnimator:
@agentzh just pushed a refactor that has the OO form you asked for.

agentzh:
@daurnimator Thanks. I'll look into it as soon as I can manage.

daurnimator:
One issue: If you try and use this in an `init_by_lua` context, it fails with a segfault.
Tracking it down: `ngx_get_connection` calls `ngx_drain_connections` which tries to dereference `ngx_cycle->reusable_connections_queue` which is `NULL`.

The easy fix would be now allowing use of `ngx.connection.new` in an init context.
But is this an issue that should be fixed elsewhere?

agentzh:
@daurnimator `init_by_lua` runs in the nginx master context which requires some kind of emulation there. The same applies to the existing cosocket API.

daurnimator:
> @daurnimator init_by_lua runs in the nginx master context which requires some kind of emulation there. The same applies to the existing cosocket API.

Okay; I'll add a context check then :)

daurnimator:
It's not working in a timer thread: the call to `ngx_http_run_posted_requests` fails due to `r->connection->data` being `NULL`.
If I add guard the call like:
```C
    if (r->connection->data)
        ngx_http_run_posted_requests(r->connection);
```
it works. However I think this is just an insufficient fake request object elsewhere?

agentzh:
@daurnimator cosockets support the timer callback context. I suggest you check out how cosockets do here.

daurnimator:
> @daurnimator cosockets support the timer callback context. I suggest you check out how cosockets do here.

I'm looking in there, but I'm not even sure which bit of code I'm meant to be similar to.
Am I even meant to be calling `ngx_http_run_posted_requests`? It does after activating the request `write_handler` in `ngx_http_lua_socket_udp_handler` and `ngx_http_lua_socket_resolve_handler`, but not in `ngx_http_lua_socket_udp_handle_success`

daurnimator:
> Am I even meant to be calling ngx_http_run_posted_requests?

As far as my testing has gone: no.
Adding commit to remove it now.

daurnimator:
Cool little example of using this PR to create a REPL: https://gist.github.com/daurnimator/984d4ec27b03ba6de6bf

agentzh:
@daurnimator `ngx_http_run_posted_requests` (or something similar) is required for subrequest scheduling. Without it, the subrequest may hang forever.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+
+int
+ngx_http_lua_connection_prep(ngx_http_request_t *r, ngx_connection_t *conn,
+    int poll_mask, ngx_msec_t wait_ms, const char **err)
+{
+    ngx_http_lua_ctx_t    *ctx;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_http_lua_udata_t  *u;
+    if ((ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module)) == NULL) {
```
agentzh: Style: we need a blank line before this line.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        *err ="no request ctx found";
+        return NGX_ERROR;
+    }
+    if ((co_ctx = ctx->cur_co_ctx) == NULL) {
```
agentzh: Style: we need a blank line before this line. Please fix other similar places.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        return NGX_ERROR;
+    }
+    conn->data = u;
+    u->request = r;
+    u->co_ctx = co_ctx;
+    u->conn = conn;
+    co_ctx->cleanup = (ngx_http_cleanup_pt)&ngx_http_lua_fd_cleanup;
+    co_ctx->data = u;
+
+    if (wait_ms != (ngx_msec_t)-1) {
```
agentzh: Style: need a space after the type cast expression `(...)`.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        ngx_add_timer(conn->read, wait_ms);
+    }
+
+    /* make sure nginx knows what to do next */
+    if (ctx->entered_content_phase) {
+        r->write_event_handler = ngx_http_lua_content_wev_handler;
+    } else {
```
agentzh: Style: need a blank line before this `} else {` line.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    lua_setmetatable(L, -2);
+    if (NGX_ERROR == ngx_http_lua_connection_init(pconn, fd, &errmsg)) {
+        return luaL_error(L, errmsg);
+    }
+    (*pconn)->data = NULL;
+    return 1;
+}
+
+
+static ngx_connection_t*
```
agentzh: Style: need a space before `*`.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    ngx_connection_t **pconn = luaL_checkudata(L, idx, NGX_HTTP_LUA_CONNECTION_KEY);
+    luaL_argcheck(L, *pconn != NULL, idx, "ngx_connection_t has been freed");
+    return *pconn;
+}
+
+
+static int
+ngx_http_lua_connection_gc(lua_State *L)
+{
+    ngx_connection_t **pconn = luaL_checkudata(L, 1, NGX_HTTP_LUA_CONNECTION_KEY);
```
agentzh: We should avoid such C99 feature that can upset older compilers.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        while (*events) {
+            if (*events == 'r') {
+                poll_mask |= POLLIN;
+            } else if (*events == 'w') {
+                poll_mask |= POLLOUT;
+            }
+            events++;
+        }
+        break;
+    case LUA_TNUMBER:
```
agentzh: Style: need a blank line before each `case` statement.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+/* need to include DDEBUG so that ngx_http_lua_util.h works */
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+#include "ngx_http_lua_connection.h"
+
+#include <math.h> /* HUGE_VAL */
+#include <poll.h> /* POLLIN, POLLOUT */
```
agentzh: We shouldn't rely on poll since the underlying events type can be arbitrary.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+#include <lua.h>
+#include <lauxlib.h>
```
agentzh: We don't need to include these headers here since `ngx_http_lua_common.h` already has them.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+#include "ngx_core.h"
+#include "ngx_alloc.h" /* ngx_alloc, ngx_free */
+#include "ngx_http.h" /* ngx_http_request_t, ngx_http_run_posted_requests */
```
agentzh: Ditto.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+#include "ngx_http_lua_util.h" /* ngx_http_lua_get_lua_vm, ngx_http_lua_run_thread, ngx_http_lua_finalize_request */
+#include "ngx_http_lua_contentby.h" /* ngx_http_lua_content_wev_handler */
+
+
+typedef struct {
+    ngx_http_request_t    *request;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_connection_t      *conn;
+} ngx_http_lua_udata_t;
```
agentzh: The `ngx_http_lua_udata_t` is not very descriptive. Will you be a bit more specific here?

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_http_lua_ctx_t    *ctx;
+
+    conn = ev->data;
+    u = conn->data;
+    r = u->request;
+    co_ctx = u->co_ctx;
+
+    ngx_http_lua_cleanup_pending_operation(co_ctx);
+    ev = NULL, u = NULL; /* now invalidated */
```
agentzh: Style: we usually avoid such comma expressions in contexts outside the `for (;;)` loop in the NGINX world.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+void
+ngx_http_lua_inject_connection(lua_State *L)
+{
+    luaL_newmetatable(L, NGX_HTTP_LUA_CONNECTION_KEY);
+    lua_pushcfunction(L, ngx_http_lua_connection_gc);
+    lua_setfield(L, -2, "__gc");
+    lua_pushcfunction(L, ngx_http_lua_connection_tostring);
+    lua_setfield(L, -2, "__tostring");
+    lua_newtable(L);
+    lua_pushcfunction(L, ngx_http_lua_connection_wait);
```
agentzh: lua_CFunction cannot be JIT compiled. I suggest we expose pure C API in the ngx_lua core and do the Lua wrapper API via FFI in lua-resty-core instead. I don't want to support the standard Lua interpreter for these new APIs.

agentzh: Please take a look at how the `ngx.semaphore` API does this.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+static int
+ngx_http_lua_connection_wait(lua_State *L)
+{
+    ngx_connection_t   *conn = ngx_http_lua_connection_check(L, 1);
+    int                 poll_mask;
+    double              timeout;
+    const char         *events, *err;
+    ngx_http_request_t *r;
+
+    switch(lua_type(L, 2)) {
```
agentzh: Such tests are a bit wasteful. I suggest we expose specialized `wait_read` and `wait_write` methods to save code branches.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        /* fall through on invalid poll_mask */
+    default:
+        return luaL_argerror(L, 2, "expected bitwise 'or' of POLLIN|POLLOUT or characters from set 'rw'");
+    }
+
+    timeout = luaL_optnumber(L, 3, HUGE_VAL); /* default to infinite timeout */
+    if (!poll_mask && timeout == HUGE_VAL) {
+        return luaL_error(L, "must provide a valid events mask or finite timeout");
+    }
+    if ((r = ngx_http_lua_get_req(L)) == NULL) {
```
agentzh: Style: please avoid using assignments in `if` conditions.

src/ngx_http_lua_connection.c
daurnimator: What is the C99 feature used here?

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    return lua_yield(L, 0);
+}
+
+
+void
+ngx_http_lua_inject_connection(lua_State *L)
+{
+    luaL_newmetatable(L, NGX_HTTP_LUA_CONNECTION_KEY);
```
agentzh: I suggest hooking the metatable onto the `ngx.connection` module table. So the user can speed up her code by writing something like this:

```lua
local ngx_connection = require "ngx.connection"
local wait_read = ngx_connection.wait_read
...
local ok, err = wait_read(c)
```

src/ngx_http_lua_util.c
```diff
@@ -49,7 +49,7 @@
 #include "ngx_http_lua_config.h"
 #include "ngx_http_lua_worker.h"
 #include "ngx_http_lua_socket_tcp.h"
-
+#include "ngx_http_lua_connection.h"
```
agentzh: Style: please preserve the original 2 blank lines here.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+int
+ngx_http_lua_connection_init(ngx_connection_t **p, ngx_socket_t fd, const char **err)
+{
+    ngx_connection_t *conn;
+    if ((conn = ngx_get_connection(fd, ngx_cycle->log)) == NULL) {
+        *err = "unable to get nginx connection";
+        return NGX_ERROR;
+    }
+    conn->data = NULL;
```
agentzh: Style: need a blank line before this line. Please fix other similar places as well.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    }
+    /* restore normal resume handler */
+    ctx->resume_handler = ngx_http_lua_wev_handler;
+    /* resume lua thread */
+    vm = ngx_http_lua_get_lua_vm(r, ctx);
+    lua_pushboolean(ctx->cur_co_ctx->co, 1);
+    /* resume coroutine */
+    rc = ngx_http_lua_run_thread(vm, r, ctx, 1 /*nret*/);
+    switch (rc) {
+        case NGX_DONE: /* coroutine finished */
```
agentzh: Style: `case` statement should align up with `switch`.

src/ngx_http_lua_connection.c
daurnimator: I have made the functions available for ffi. Notice all the functions that were not declared `static` (are are in headers)

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+    timeout = luaL_optnumber(L, 3, HUGE_VAL); /* default to infinite timeout */
+    if (!poll_mask && timeout == HUGE_VAL) {
+        return luaL_error(L, "must provide a valid events mask or finite timeout");
+    }
+    if ((r = ngx_http_lua_get_req(L)) == NULL) {
+        return luaL_error(L, "no request found");
+    }
+
+    if (NGX_ERROR == ngx_http_lua_connection_prep(r, conn, poll_mask,
```
agentzh: Style: please put the constant on the right hand side of comparison operators. (I do understand the rationale before this style but 1) modern C compilers already emit warnings for such misuses and 2) we should follow nginx's coding style).

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+        (timeout < 0)? 0 : (timeout == HUGE_VAL)? ((ngx_msec_t)-1) : timeout*1000, &err)) {
```
agentzh: Style: `{` should be in its own line when the if condition has been splitted into multiple lines. Please fix other similar places (if any).

src/ngx_http_lua_connection.c
daurnimator: most libraries return values POLLIN/POLLOUT. This is for convenience so that you don't need to transfrom from OS standard constants to nginx specific ones.

src/ngx_http_lua_connection.c
daurnimator: I usually like being explicit here: If you use a declaration in a header; always include that directly (otherwise you get into a mess of implementation specific details).

I can remove it if it's against your style though :(

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+    ngx_http_lua_ctx_t *ctx;
+    lua_State          *vm;
+    ngx_int_t           rc;
+    if ((ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module)) == NULL) {
+        return NGX_ERROR;
+    }
+    /* restore normal resume handler */
+    ctx->resume_handler = ngx_http_lua_wev_handler;
+    /* resume lua thread */
+    vm = ngx_http_lua_get_lua_vm(r, ctx);
```
agentzh: How will you handle repeated event firing when the user code does not completely consume the events under LT (Level-Triggered) mode? You should call `ngx_handle_read_event` and `ngx_handle_write_event` at right places I think.

src/ngx_http_lua_connection.c
```diff
@@ -0,0 +1,353 @@
...
+
+typedef struct {
+    ngx_http_request_t    *request;
+    ngx_http_lua_co_ctx_t *co_ctx;
+    ngx_connection_t      *conn;
+} ngx_http_lua_udata_t;
+
+
+static ngx_int_t
+ngx_http_lua_fd_resume_request(ngx_http_request_t *r)
```
agentzh: I think C function prefixes like `ngx_http_lua_conn_` makes more sense than `ngx_http_lua_fd` for the current file name, `ngx_http_lua_connection.c`.

src/ngx_http_lua_connection.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (C) 2014 Daurnimator
+ */
+
+#ifndef _NGX_HTTP_LUA_CONNECTION_H_INCLUDED_
+#define _NGX_HTTP_LUA_CONNECTION_H_INCLUDED_
+
+#include <lua.h>
```
agentzh: Again, no need to explicitly include `lua.h` here.

src/ngx_http_lua_connection.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (C) 2014 Daurnimator
+ */
+
+#ifndef _NGX_HTTP_LUA_CONNECTION_H_INCLUDED_
+#define _NGX_HTTP_LUA_CONNECTION_H_INCLUDED_
+
```
agentzh: Style: need 2 blank lines here. Please fix other similar places. Thank you.

src/ngx_http_lua_connection.c
daurnimator: It already is... see the following ~10 lines.

agentzh:
@daurnimator Please address my comments during the review. Thank you :) And we need a paired PR for lua-resty-core that does the Lua API in FFI. After we have a Lua API we both agree on, you're recommended to also add some tests for this :) Thanks!

src/ngx_http_lua_connection.c
daurnimator: The whole point of this API is that it will only *wait* for the event. it's up to the libraries to consume the event.

It's cleaned up eventually by calling `ngx_del_con` in the cleanup handler.

src/ngx_http_lua_connection.c
daurnimator: Looks like this once missed the rename spree: will rename to `ngx_http_lua_connection_resume_request`

daurnimator:
> @daurnimator Please address my comments during the review. Thank you :) And we need a paired PR for lua-resty-core that does the Lua API in FFI. After we have a Lua API we both agree on

I figured we'd agree on the API first and merge this.
Then once it's merged we can code up the luajit/ffi speedup for it.

> you're recommended to also add some tests for this 

I could never get the openresty test suite to run on my machine. I have 0 skill at debugging (or writing) perl, so I don't think I'll be able to do this. Can someone else step in?

src/ngx_http_lua_connection.c
agentzh: @daurnimator Sorry, it's not really a C standard thing. It's really an NGINX coding style thing. We should use constant expressions in variable initializations.

src/ngx_http_lua_connection.c
agentzh: @daurnimator I don't want to keep compatibility with the standard Lua interpreter for new nontrivial APIs now since it's quite a maintenance burden.

src/ngx_http_lua_connection.c
agentzh: @daurnimator Use of these constants make the reader think we are actually using poll(). Also please keep in mind we need to support non-POSIX systems where no poll() exists, like Windows.

src/ngx_http_lua_connection.c
agentzh: @daurnimator I like shorter code so that we can easily override things in standard headers (if necessary) in a central place.

src/ngx_http_lua_connection.c
daurnimator: I need them myself.... e.g. I debug using standard lua; then once it works there, I go debug with luajit.
Please reconsider this.

src/ngx_http_lua_connection.c
agentzh: @daurnimator The libraries do not always "consume" the event, in which case, it is our responsibility to de-activate the event for LT events otherwise we're looping here like a hell and there's no way for the "libraries" to manipulate the events directly since its agnostic.

src/ngx_http_lua_connection.c
daurnimator: > please keep in mind we need to support non-POSIX systems where no poll() exists, like Windows.

There is a poll on windows. and it uses the same constants: https://msdn.microsoft.com/en-us/library/windows/desktop/ms740094(v=vs.85).aspx

src/ngx_http_lua_connection.c
agentzh: @daurnimator The `ngx_del_con` thing is not enough for such purposes since you should re-create and destroy the `ngx.connection` object upon every event, which is just too expensive.

src/ngx_http_lua_connection.c
agentzh: @daurnimator Ah, I prefer something shorter like `conn` since seeing long prefixes like `ngx_http_lua_conneciton_` all over the place would make me crazy ;)

agentzh:
@daurnimator The test scaffold does not use Perl to code up test cases. We use our own (minimal) declarative DSL. Just take a look at one of the `.t` file and count the number of Perl code yourself :) If you have any problems with the testing toolchain, please let us know.

agentzh:
@daurnimator I don't want to merge the classic Lua C API implementation since I don't want to maintain two parallel implementations in ngx_lua for such relatively big new features. Sorry. The `ngx.semaphore` API does not have a classic C API implementation either, for example.

src/ngx_http_lua_connection.c
daurnimator: >  seeing long prefixes like ngx_http_lua_conneciton_ all over the place would make me crazy ;)

All the other functions in this file have `ngx_http_lua_connection_`. I like it to match the filename. Would you like me to rename the file to `ngx_http_lua_conn.c`?

Aside: `ngx_http_lua` really seems like the long prefix here: if you're worried about long prefixes, why not rename to e.g. `nhl_` across all files?

src/ngx_http_lua_connection.c
agentzh: @daurnimator Are you sure? There's not even a `poll.h` there on Windows :)

agentzh: @daurnimator The link you referenced is not a POSIX-compliant thing but Window's own creature. Similar but not really compatible.

daurnimator:
> @daurnimator I don't want to merge the classic Lua C API implementation since I don't want to maintain two parallel implementations in ngx_lua for such relatively big new features. Sorry. 

I can remove it from my PR... I can load+use it as a shared library maintained elsewhere for debugging purposes.

> The ngx.semaphore API does not have a classic C API implementation either, for example.

Yeah.... I don't think thats a good thing.

src/ngx_http_lua_connection.c
agentzh: @daurnimator I'm fine with the file name change.

src/ngx_http_lua_connection.c
daurnimator: > The link you referenced is not a POSIX-compliant thing but Window's own creature. Similar but not really compatible.

Correct.
But compatible enough that it defines POLLIN and POLLOUT, which is all I need.

src/ngx_http_lua_connection.c
daurnimator: Should I rename the library namespace to `ngx.conn.new()`? I do enjoy having them all match up!

agentzh: Your `nhl_` suggestion is too extreme to be constructive. You know that's not readable at all. We always need to balance things in engineering.

src/ngx_http_lua_connection.c
agentzh: @daurnimator I think you've already got my points so I won't repeat here :)

src/ngx_http_lua_connection.c
agentzh: @daurnimator I prefer longer and more descriptive names on the end user Lua module API side. We can be a bit more flexible in our internal implementations and ABI though.

daurnimator: > Your nhl_ suggestion is too extreme to be constructive. You know that's not readable at all.

Considering we're within the `lua-nginx-module` codebase, I think that a prefix only needs to be enough to namespace ourselves/disambiguate from other libraries. something short like `nhl_` does this sufficiently.

agentzh: @daurnimator Are you serious or just trolling?

src/ngx_http_lua_connection.c
agentzh: @daurnimator That's a reasonable approach but unfortunately we are accumulating features exclusive to LuaJIT.

src/ngx_http_lua_connection.c
daurnimator: FWIW, I just looked it up, and NGX_READ_EVENT is a #define to POLLIN on linux+osx anyway.

The problem is that NGX_READ_EVENT is not safe to be used as a bitmask.
e.g. see the `select` `#ifdef`:
``` C
#define NGX_READ_EVENT     0
#define NGX_WRITE_EVENT    1
```

agentzh:
@daurnimator I'm sorry that I'm picky when reviewing PRs. I am always picky and you can check out all the existing PRs. Please don't take it personal. I hope we can keep this discussion civil and friendly :)

I always appreciate and value your contributions a lot.

src/ngx_http_lua_connection.c
daurnimator: I'm serious. I value consistency over terseness. I was suggesting a way to be both consistent *and* terse.

src/ngx_http_lua_connection.c
agentzh: @daurnimator That's why NGINX introduced such indirections in the first place so that nginx's build system can point them to the right things on different platforms :)

src/ngx_http_lua_connection.c
agentzh: @daurnimator Lua module names should be a bit more descriptive. I think `ngx.connection` is better than `ngx.conn` given that we already have `ngx.semaphore`.

agentzh: @daurnimator I think it's fine to use shorter names for different contexts. Huffman's principle are in play :)

agentzh: @daurnimator For instance, NGINX does not use `nginx_http_blah_blah_blah` but use `ngx_http_blah_blah_blah` itself. So we are consistent with NGINX itself here anyway. It's all about balance.

src/ngx_http_lua_connection.c
daurnimator: It sounds like in this case I'll need to invent my own identifiers then? (because there is no `NGX_NO_EVENT` or `NGX_READ_WRITE_EVENT`).

src/ngx_http_lua_connection.c
daurnimator: > It's all about balance.

I don't think I'll rename the file then.
If there's going to be inconsistency; lets just shorten the prefix but leave it in connection.c

src/ngx_http_lua_connection.c
agentzh: @daurnimator Yes, that sounds better to me.

src/ngx_http_lua_connection.c
agentzh: @daurnimator Fine.

src/ngx_http_lua_connection.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (C) 2014 Daurnimator
```
agentzh: @daurnimator Will you please state that "I hereby assign copyright in this code to the lua-nginx-module project, to be licensed under the same terms as the rest of the code"? This is just for legal resolution. Thanks for  your understanding.

bungle:
I think I could use this with libmodbus:
http://libmodbus.org/docs/v3.1.4/modbus_get_socket.html

I already started writing FFI bindings to that library (still WIP, though):
https://github.com/bungle/lua-resty-modbus

Would be great if we could make OpenResty as a non-blocking modbus server. I have by the way already connected this to a solar panel controller :-).
 
Is this still considered to be included an/or worked on, @daurnimator / @agentzh?

agentzh:
@bungle Yeah, it's still on the official roadmap though not of the highest priority atm.

daurnimator:
Hey @agentzh, how can we proceed with an API for waiting for a file descriptor?
Forgive me, but I can't remember the next steps we laid out.

agentzh:
@daurnimator Me either. I need some time to recap. Will look into this as soon as I can manage.

guanlan:
@daurnimator 
Here is a good example of testing:
https://github.com/openresty/lua-nginx-module/blob/master/t/077-sleep.t

lePereT:
Did this PR ever get anywhere?

bungle:
@lePereT, this PR has never got merged, but there is something:
https://github.com/openresty/lua-nginx-module#ngxrun_worker_thread

zhuizhuhaomeng:
This PR hasn't been updated in a long time, and based on previous discussions, I'm definitely closing this PR.

PR #451, opened 2015-01-20
Update README.markdown

lneto:
s/cococket/cosocket


agentzh:
@lneto Applied a slightly modified version of your patch to git master. Thanks!

PR #456, opened 2015-02-12
readme: add note on possible uninitialized variables in rewrite phase

Sirupsen:
We noticed in production testing that some Nginx variables would be empty in some cases in the logging phase. Through investigation this seemed to be due to the possibility of Nginx short circuiting requests in `ngx_http_request.c`, meaning the rewrite phase would be skipped completely.

If we send a bad request, for example with an empty host header (`curl -H "Host: " localhost:80`), to a configuration that had something like:

```nginx
set $animal "walrus";
log_by_lua 'ngx.log(ngx.ERR, ngx.var.animal)';
```

We would get a warning that `ngx.var.animal` is uninitialized.

This is fairly painful to track down and not immediately obvious, so I thought it'd save someone some time by mentioning it in the `log_by_lua` documentation.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -41,6 +41,7 @@ Table of Contents
     * [Special PCRE Sequences](#special-pcre-sequences)
     * [Mixing with SSI Not Supported](#mixing-with-ssi-not-supported)
     * [SPDY Mode Not Fully Supported](#spdy-mode-not-fully-supported)
+    * [Missing data on short circuited requests](#missing-data-on-short-circuited-requests)
 * [TODO](#todo)
 * [Changes](#changes)
 * [Test Suite](#test-suite)
@@ -850,6 +851,27 @@ SPDY Mode Not Fully Supported
 -----------------------------
 
 Certain Lua APIs provided by ngx_lua do not work in Nginx's SPDY mode yet: [ngx.location.capture](#ngxlocationcapture), [ngx.location.capture_multi](#ngxlocationcapture_multi), and [ngx.req.socket](#ngxreqsocket).
+
+[Back to TOC](#table-of-contents)
+
+Missing data on short circuited requests
+----------------------------------------
+
+Nginx may terminate a request early with (at least):
+
+* `400 (Bad Request)`
+* `405 (Not Allowed)`
+* `408 (Request Timeout)`
+* `414 (Request URI Too Large)`
+* `494 (Request Headers Too Large)
+* `499 (Client Closed Request)`
+* `500 (Internal Server Error)`
+* `501 (Not Implemented)`
+
+This means that phases that normally run are skipped, such as the rewrite or
+access phase. This also means that later phases that are run regardless, e.g.
+`log_by_lua`, will not have access to information that is normally set in those
+phases.
 
 [Back to TOC](#table-of-contents)
 
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -700,6 +700,24 @@ Mixing SSI with ngx_lua in the same Nginx request is not supported at all. Just 
 
 Certain Lua APIs provided by ngx_lua do not work in Nginx's SPDY mode yet: [[#ngx.location.capture|ngx.location.capture]], [[#ngx.location.capture_multi|ngx.location.capture_multi]], and [[#ngx.req.socket|ngx.req.socket]].
 
+== Missing data on short circuited requests ==
+
+Nginx may terminate a request early with (at least):
+
+* `400 (Bad Request)`
+* `405 (Not Allowed)`
+* `408 (Request Timeout)`
+* `414 (Request URI Too Large)`
+* `494 (Request Headers Too Large)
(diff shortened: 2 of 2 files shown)


README.markdown
```diff
@@ -1904,6 +1904,8 @@ Note that the following API functions are currently disabled within this context
 * Subrequest API functions (e.g., [ngx.location.capture](#ngxlocationcapture) and [ngx.location.capture_multi](#ngxlocationcapture_multi))
 * Cosocket API functions (e.g., [ngx.socket.tcp](#ngxsockettcp) and [ngx.req.socket](#ngxreqsocket)).
 
+Note that Nginx may reject a request early with `ngx.HTTP_BAD_REQUEST`, `ngx.HTTP_REQUEST_TIMEOUT`, `ngx.HTTP_INTERNAL_SERVER_ERROR`, `ngx.HTTP_REQUEST_URI_TOO_LARGE`, `ngx.HTTP_NOT_IMPLEMENTED`, `ngx.HTTP_NOT_ALLOWED`, `ngx.HTTP_NOT_IMPLEMENTED` or `ngx.HTTP_REQUEST_HEADER_TOO_LARGE`. This causes variables set during the rewrite phase to be uninitialized in the logging phase.
```
agentzh: Better avoid using ngx_lua's own constants here. Please use `400 (Bad Request)` and etc here instead.

agentzh: BTW, this README.markdown file is automatically generated from `doc/HttpLuaModule.wiki` via the `ngx-releng` tool from the `nginx-devel-utils` github repos. So better change the `.wiki` file instead.

agentzh:
@Sirupsen This does not only affect nginx variables but also any other request-related data like request headers and etc.

Sirupsen:
Yeah, I'll update the documentation to include that, change to not use `ngx_lua` constants and do it in the wiki instead.

agentzh:
@Sirupsen Cool, thanks :) Maybe we can put this under the "Known Issues" section?

Sirupsen:
I was having that internal debate with myself. Known issues to me implies something fixable, this "problem" however makes a lot of sense if you know how Nginx internals work and isn't something that's going to be 'fixed' but is actually expected if you think about it. On the other hand, it may be where people might go to look for something like this. You make the decision :)

agentzh:
@Sirupsen Not all "known issues" are fixable anyway :) But I agree we might need a separate section like "Pitfalls" or "Common Mistakes" to avoid potential confusions. What do you think?

Sirupsen:
I honestly think in cases where they can be under a call or directive they should be there. It's be the first place I would look and would in many cases mean I'd know it before I went down the miserable debugging road. A section like that I'd only check once I've been questioning my sanity for long enough. Obviously it's better than nothing, but when they can be attached to something I think they should. What do you think?

agentzh:
@Sirupsen The problem is that this issue affects a *lot* of directives and Lua API functions, not just your `rewrite_by_lua` case. And it's actually a general nginx behavior (for natural reasons) which is not even specific to ngx_lua at all :)

Sirupsen:
@agentzh I tried to add it under `Known Issues`.

Sirupsen:
I'm having trouble phrasing it the most accurately. I haven't tracked down every single case of this, e.g. I'm sure the SSL module short-circuits requests too. I just ran:

```bash
$ grep -A 1 -P "(ngx_http_close_request|ngx_http_finalize_request)" src/http/ngx_http_request.c | grep -Eo "NGX_HTTP_\w+" | sort | uniq
NGX_HTTP_BAD_REQUEST
NGX_HTTP_CLIENT_CLOSED_REQUEST
NGX_HTTP_INTERNAL_SERVER_ERROR
NGX_HTTP_NOT_ALLOWED
NGX_HTTP_NOT_IMPLEMENTED
NGX_HTTP_REQUEST_HEADER_TOO_LARGE
NGX_HTTP_REQUEST_TIME_OUT
NGX_HTTP_REQUEST_URI_TOO_LARGE
NGX_HTTP_TO_HTTPS
```

These are the only ones I've seen in production too.

agentzh:
@Sirupsen No need to rephrase it accurately because it is hairy in the nginx core per se.

agentzh:
@Sirupsen Just applied a slightly modified version of your patch. Thanks!

Sirupsen:
I'll close this since you already put it in master.

yeah it is.

dreddi:
Leaving a comment here for posterity, after tracing the [Missing data on short circuited requests](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#missing-data-on-short-circuited-requests) section in the readme back to this PR.

I find that when exiting early from an access_by_lua_block with `return ngx.exit(429)` (NGX_HTTP_TOO_MANY_REQUESTS), any ngx.var.VAR variables that are set during the block, are unset by the time the nginx access log is prepared for that request, and will be missing from the logs.

This differs from the README, in that terminating early with a 400 or 500, or other codes (403), will populate the access log correctly, but returning 429 will not.

PR #464, opened 2015-03-03
missing use lib

maage:
Some tests are missing use lib mantra.

You can check if use lib is missing:

```
grep -cr 'use lib' t|grep '\.t:'|grep -v :1
```

### DIFF
--- a/t/034-match.t
+++ b/t/034-match.t
@@ -1,5 +1,5 @@
 # vim:set ft= ts=4 sw=4 et fdm=marker:
-
+use lib 'lib';
 use Test::Nginx::Socket::Lua;
 
 #worker_connections(1014);
--- a/t/089-phase.t
+++ b/t/089-phase.t
@@ -1,5 +1,5 @@
 # vim:set ft= ts=4 sw=4 et fdm=marker:
-
+use lib 'lib';
 use Test::Nginx::Socket::Lua;
 
 #worker_connections(1014);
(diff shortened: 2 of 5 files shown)


agentzh:
@maage Thank you for looking but I think the `use lib 'lib'` is useless, no?

maage:
Main problem here is inconsistency. `use lib 'lib'` is not needed if you use `prove -l` to run tests. So far almost all tests at lua-nginx-module have it. When I forget to add `-l` some tests failed and some did not.

I  setup modules to have lib symlink to my current Test::Nginx directory. This way I do not get any hassle when I want to modify Test::Nginx. Also at Maxim Dounins http://mdounin.ru/hg/nginx-tests there is different implementation of Test::Nginx and if I install your version globally, there might be some problems.

It might be better to remove every `use lib 'lib'`, less code. Then you always need to remember add `-l` when running prove.

I run tests currently like this:
```
cd <root_of_module>
prove -l -r --directives t
```

agentzh:
@maage Let's just remove all the `use lib 'lib'` for the sake of inconsistency :)

Usually I use the following bash script named `./go` for running tests.

```bash
#!/usr/bin/env bash

prove -I../test-nginx/lib "$@"
```

And then just call it directly from vim:

```
:!./go %
```

where `%` expands by vim to the current file being edited.

Regards,
-agentzh

agentzh:
Okay, I've just removed all the `use lib ...` craps from the `.t` files. Thanks for the report.

PR #465, opened 2015-03-03
lua tables are not sorted

maage:
See for example: http://www.lua.org/pil/19.3.html

Two strategies to solve this, either sort tables explicitly if it is
possible. When sorting table is not possible, allow result in any order
by using regexp.

Regexp are mainly done with following script. Just drop outer (?^: ... ). Might be better just to use less items as generated regexps are not easy to follow anyways after there is more than 2 iterms.

 ```
#!/usr/bin/perl
use Algorithm::Permute;
use Regexp::Assemble;
my @arr = map { split } @ARGV;
my $ra = Regexp::Assemble->new;
Algorithm::Permute::permute { $ra->add( '^\Q'.join('&', @arr).'\E'.q{$} ) } @arr;
print $ra->re;
print "\n";
```

### DIFF
--- a/t/030-uri-args.t
+++ b/t/030-uri-args.t
@@ -655,7 +655,12 @@ args:
 --- request
 GET /lua
 --- response_body_like
-^args:(?:a=9&a=2&b=3|b=3&a=9&a=2)$
+(?x) ^args:
+    (?= .*? \b a=9 \b )  # 3 chars
+    (?= .*? \b a=2 \b )  # 3 chars
+    (?= .*? \b b=3 \b )  # 3 chars
+    (?= (?: [^&]+ & ){2} [^&]+ $ )  # requires exactly 2 &'s
+    (?= .{11} $ )  # requires for total 11 chars (exactly) in the string
 
 
 
@@ -700,8 +705,13 @@ args: foo=3
     }
 --- request
 GET /lua
---- response_body
-foo=3&bar=32&bar
+--- response_body_like
+(?x) ^
+    (?= .*? \b bar=32 \b )     # 6 chars
+    (?= .*? \b bar (?!=) \b )  # 3 chars
+    (?= .*? \b foo=3 \b )      # 5 chars
+    (?= (?: [^&]+ & ){2} [^&]+ $ )  # requires exactly 2 &'s
+    (?= .{16} $ )  # requires for total 16 chars (exactly) in the string
 --- no_error_log
 [error]
 
@@ -1129,7 +1139,12 @@ b = foo
         content_by_lua '
             local s = "f+f=bar&B=foo"
             args = ngx.decode_args(s)
+            local arr = {}
             for k, v in pairs(args) do
+                table.insert(arr, k)
+            end
+            table.sort(arr)
+            for i, k in ipairs(arr) do
                 ngx.say("key: ", k)
             end
             ngx.say("s = ", s)
@@ -1138,8 +1153,8 @@ b = foo
 --- request
 GET /lua
 --- response_body
-key: f f
 key: B
+key: f f
 s = f+f=bar&B=foo
 --- no_error_log
 [error]
@@ -1170,8 +1185,13 @@ s = f+f=bar&B=foo
             for _, command in pairs(commands) do
                 --command = ngx.unescape_uri(command)
                 local request_args = ngx.decode_args(command, 0)
-                for key, value in pairs(request_args) do
-                    ngx.say(key, ": ", value)
+                local arr = {}
+                for k, v in pairs(request_args) do
+                    table.insert(arr, k)
+                end
+                table.sort(arr)
+                for i, k in ipairs(arr) do
+                    ngx.say(k, ": ", request_args[k])
                 end
                 ngx.say(" ===============")
             end
@@ -1181,20 +1201,20 @@ s = f+f=bar&B=foo
 POST /t
 method=zadd&key=User%3A1227713%3Alikes%3Atwitters&arg1=1356514698&arg2=780984852||method=zadd&key=User%3A1227713%3Alikes%3Atwitters&arg1=1356514698&arg2=780984852||method=zadd&key=User%3A1227713%3Alikes%3Atwitters&arg1=1356514698&arg2=780984852
 --- response_body
-arg2: 780984852
-method: zadd
-key: User:1227713:likes:twitters
 arg1: 1356514698
+arg2: 780984852
+key: User:1227713:likes:twitters
+method: zadd
  ===============
-arg2: 780984852
-method: zadd
-key: User:1227713:likes:twitters
 arg1: 1356514698
+arg2: 780984852
+key: User:1227713:likes:twitters
+method: zadd
  ===============
-arg2: 780984852
-method: zadd
-key: User:1227713:likes:twitters
 arg1: 1356514698
+arg2: 780984852
+key: User:1227713:likes:twitters
+method: zadd
  ===============
 --- no_error_log
 [error]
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: t/099-c-api.t [hunk 1]
```diff
@@ -229,9 +229,8 @@ bar: rc=0, type=1, val=0
     }
 --- request
 GET /test
---- response_body
-foo: rc=-5
-bar: rc=-5
+--- response_body eval
+qr/foo: rc=-5\nbar: rc=-5\n|bar: rc=-5\nfoo: rc=-5\n/
```
agentzh: @maage The `foo`/`bar` order is deterministic in the test script. Why do we need to take into account the reversed order case? Isn't that impossible?

t/055-subreq-vars.t
```diff
@@ -65,8 +65,8 @@ M(http-subrequest-start) {
 --- request
 GET /lua
 --- response_body_like: 500 Internal Server Error
---- error_log chop
-variable "dog" cannot be assigned a value (maybe you forgot to define it first?)
+--- error_log eval
+qr/variable "(dog|cat)" cannot be assigned a value \(maybe you forgot to define it first\?\)/
```
agentzh: I don't think we need this because the `$dog` variable is always accessed first before `$cat`, hence the error message complaining about `$dog` instead of `$cat`.

t/030-uri-args.t
```diff
@@ -1298,8 +1312,8 @@ foo=3&a%20b=32&bar=5
     }
 --- request
 GET /lua
---- response_body
-foo=3&a%20b=32&a%20b&bar=5
+--- response_body_like
+^(?:a%20b(?:&(?:bar=5&(?:a%20b=32&foo=3|foo=3&a%20b=32)|foo=3&(?:a%20b=32&bar=5|bar=5&a%20b=32)|a%20b=32&(?:bar=5&foo=3|foo=3&bar=5))|=32&(?:a%20b&(?:bar=5&foo=3|foo=3&bar=5)|bar=5&(?:a%20b&foo=3|foo=3&a%20b)|foo=3&(?:a%20b&bar=5|bar=5&a%20b)))|bar=5&(?:a%20b(?:&(?:a%20b=32&foo=3|foo=3&a%20b=32)|=32&(?:a%20b&foo=3|foo=3&a%20b))|foo=3&a%20b(?:&a%20b=32|=32&a%20b))|foo=3&(?:a%20b(?:&(?:a%20b=32&bar=5|bar=5&a%20b=32)|=32&(?:a%20b&bar=5|bar=5&a%20b))|bar=5&a%20b(?:&a%20b=32|=32&a%20b)))$
```
agentzh: Ah, this is ugly. I think we need a better way here. But I admit we need to do something here :)

t/030-uri-args.t
```diff
@@ -1280,14 +1293,15 @@ foo=3&a=32&bar=5
     }
 --- request
 GET /lua
---- response_body
-foo=3&a%20b=32&bar=5
+--- response_body_like
+^(?:bar=5&(?:a%20b=32&foo=3|foo=3&a%20b=32)|foo=3&(?:a%20b=32&bar=5|bar=5&a%20b=32)|a%20b=32&(?:bar=5&foo=3|foo=3&bar=5))$
```
agentzh: Same as below.

t/030-uri-args.t
agentzh: Maybe using the following (a little bit weeker) regex?

```
(?x) ^
    (?= .*? \b foo=3 \b )
    (?= .*? \b a%20b=32 \b )
    (?= .*? \b a%20b \b )
    (?= .*? \b bar=5 \b )
    (?= (?: [^\&]+ \& ){3} [^\&]+ )  # requires exactly 3 &'s
    (?= .{26} $ )  # requires for total 26 chars (exactly) in the string
```

t/099-c-api.t
maage: Yes, you are right. In a hindsight I do not know why I added this. I had problems with ffi.

t/030-uri-args.t
maage: Yes this is much better way. Easy to understand and quite simple to update.

& test should have $ at the end.

t/055-subreq-vars.t
maage: When I use Lua 5.1 from RHEL-6 I get this:
```
t/055-subreq-vars.t .. 17/54 
#   Failed test 'TEST 1: set non-existent variables via "vars" option - pattern "variable "dog" cannot be assigned a value (maybe you forgot to define it first?)" matches a line in error.log (req 0)'
#   at lib/Test/Nginx/Socket.pm line 1105.

#   Failed test 'TEST 1: set non-existent variables via "vars" option - pattern "variable "dog" cannot be assigned a value (maybe you forgot to define it first?)" matches a line in error.log (req 1)'
#   at lib/Test/Nginx/Socket.pm line 1105.
```
If I modify test to have "cat" it works. I'll look into this some more.

maage: I've modified test:
```
use lib 'lib';
use Test::Nginx::Socket::Lua;

plan tests => repeat_each() * (blocks() * 3);
run_tests();

__DATA__

=== TEST 1: set non-existent variables via "vars" option
--- config
    location /other {
        echo Ok;
    }

    location /lua {
        content_by_lua '
            res = ngx.location.capture("/other",
                { vars = { dog = "hello", cat = 32 }});

            ngx.print(res.body)
        ';
    }

--- request
GET /lua
--- no_error_log
[error]
[alert]
```
And this shows:
```
% prove -r -l --directives t/055-subreq-vars-1.t
t/055-subreq-vars-1.t .. 1/3 
#   Failed test 'TEST 1: set non-existent variables via "vars" option - status code ok'
#   at lib/Test/Nginx/Socket.pm line 894.
#          got: '500'
#     expected: '200'

#   Failed test 'TEST 1: set non-existent variables via "vars" option - pattern "[error]" should not match any line in error.log but matches line "2015-03-04T14:04:30.674+00:00 [error] 16147\#0: *1 variable \"cat\" cannot be assigned a value (maybe you forgot to define it first?) , client: 127.0.0.1, server: localhost, request: \"GET /lua HTTP/1.1\", host: \"localhost\"" (req 0)'
#   at lib/Test/Nginx/Socket.pm line 1160.

#   Failed test 'TEST 1: set non-existent variables via "vars" option - pattern "[error]" should not match any line in error.log but matches line "2015-03-04T14:04:30.674+00:00 [error] 16147\#0: *1 lua entry thread aborted: runtime error: content_by_lua(nginx.conf:53):2: failed to adjust the subrequest: -1" (req 0)'
#   at lib/Test/Nginx/Socket.pm line 1160.
t/055-subreq-vars-1.t .. Failed 3/3 subtests 

Test Summary Report
-------------------
t/055-subreq-vars-1.t (Wstat: 0 Tests: 4 Failed: 4)
  Failed tests:  1-4
  Parse errors: Bad plan.  You planned 3 tests but ran 4.
Files=1, Tests=4,  0 wallclock secs ( 0.04 usr  0.01 sys +  0.20 cusr  0.07 csys =  0.32 CPU)
Result: FAIL
```
So my patch is correct. Error message is generated when vars table is assigned and the table is unordered.

t/030-uri-args.t
agentzh: @maage Yeah, right.

t/055-subreq-vars.t
agentzh: @maage Oh yeah, I was wrong. The variable sets happen when the subrequest is created instead of first being referenced in the subrequest running phases. I forgot the implementation details. Sorry. You're right.

agentzh:
@maage Please just commit your updates to the same branch (you could rewrite your previous commit and push by force to your branch if desired). No need to close the current PR and create a new one ;) Thanks!

maage:
I've updated branch. All complex (more than two elements) regexp are now using your base plate. As there was some substrings I added negative or positive lookaheads too.

agentzh:
@maage Thanks for your contribution! Just applied to master.

PR #466, opened 2015-03-03
fix env tests

maage:
In my nginx implementation only explicitly set envs are visible.

Other strategy is to handle this is to add these at Test::Nginx::Util write_config_file lines 814-815.


agentzh:
@maage Merged. Thanks!

PR #468, opened 2015-03-06
util: don't set content type header on 304

Sirupsen:
This is a violation of HTTP. 304 doesn't allow a body, and therefore should not specify the `Content-Type`. [`nginx` handles this situation correctly](https://github.com/nginx/nginx/blob/aa248e453fd0df9623d5da0f13727049e69e503d/src/http/modules/ngx_http_not_modified_filter_module.c#L96), however `ngx-lua` does not. It always sets it, even on empty responses. Note also that `nginx` [does the same thing in `ngx_http_special_response `](https://github.com/nginx/nginx/blob/aa248e453fd0df9623d5da0f13727049e69e503d/src/http/ngx_http_special_response.c#L402).

Some clients cope with this, others do not. Our symptom in production was that some users would hit a cached page and be prompted to download a file with a dialog (because our default type: `Content-Type: application/octet-stream`). We could only reproduce this for users behind the Squid proxy.

For now, we've just disabled `lua_use_default_type`. However, it seems to me that this is a bug in the `ngx-lua` defaults.

@agentzh

### DIFF
--- a/src/ngx_http_lua_util.h
+++ b/src/ngx_http_lua_util.h
@@ -379,7 +379,9 @@ ngx_http_lua_set_content_type(ngx_http_request_t *r)
     ngx_http_lua_loc_conf_t     *llcf;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (llcf->use_default_type) {
+    if (llcf->use_default_type
+        && r->headers_out.status != NGX_HTTP_NOT_MODIFIED)
+    {
         return ngx_http_set_content_type(r);
     }
 
--- a/t/118-use-default-type.t
+++ b/t/118-use-default-type.t
@@ -120,3 +120,22 @@ hello
 --- no_error_log
 [error]
 
+
+
+=== TEST 6: lua_use_default_type on does not set content type on 304
+--- config
+    lua_use_default_type on;
+    location /lua {
+        default_type text/plain;
+        content_by_lua '
+            ngx.status = ngx.HTTP_NOT_MODIFIED
+        ';
+    }
+--- request
+GET /lua
+--- response_body
+--- response_headers
+!Content-Type
+--- no_error_log
+[error]
+--- error_code: 304


src/ngx_http_lua_util.h
```diff
@@ -69,6 +69,10 @@ extern char ngx_http_lua_headers_metatable_key;
 #define NGX_HTTP_SWITCHING_PROTOCOLS 101
 #endif
 
+#ifndef NGX_HTTP_NOT_MODIFIED
```
agentzh: We don't need this because the macro is always there (as far as the minimum nginx core version that ngx_lua supports, 0.8.54).

agentzh:
@Sirupsen Looking mostly good to me. But will you please add some test cases for it to the existing test suite? Thanks!

Sirupsen:
I'll remove the macro and add tests.

src/ngx_http_lua_util.h
Sirupsen: Oh, so that's the only reason you had it for `101`. Gotcha.

agentzh: @Sirupsen Correct :)

agentzh:
@Sirupsen Thanks. Please drop me a note when you're ready :)

Sirupsen:
@agentzh I got the test suite running and added a test! I think this is the only test we need.

t/118-use-default-type.t
```diff
@@ -120,3 +120,20 @@ hello
 --- no_error_log
 [error]
```
agentzh: Style: 3 blank lines are required between test blocks. BTW, we usually use the `reindex` script to handle such things automatically for us: https://raw.githubusercontent.com/agentzh/old-openresty/master/bin/reindex

src/ngx_http_lua_util.h
```diff
@@ -379,7 +379,7 @@ ngx_http_lua_set_content_type(ngx_http_request_t *r)
     ngx_http_lua_loc_conf_t     *llcf;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (llcf->use_default_type) {
+    if (llcf->use_default_type && r->headers_out.status != NGX_HTTP_NOT_MODIFIED) {
```
agentzh: Style: lines should not exceed 80 columns :)

Sirupsen:
@agentzh fixed

src/ngx_http_lua_util.h
```diff
@@ -379,7 +379,8 @@ ngx_http_lua_set_content_type(ngx_http_request_t *r)
     ngx_http_lua_loc_conf_t     *llcf;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (llcf->use_default_type) {
+    if (llcf->use_default_type
+            && r->headers_out.status != NGX_HTTP_NOT_MODIFIED) {
```
agentzh: Style: nginx usually aligns `&&` with the first character after `(` in the previous line. Also, for splitted `if` statements, nginx usually puts the `{` on a separate line (and aligns it up with `if`'s beginning column).

Sirupsen:
Checked with an example in `src/ngx_http_lua_control.c`. Should be the same now. :)

src/ngx_http_lua_util.h
```diff
@@ -379,7 +379,9 @@ ngx_http_lua_set_content_type(ngx_http_request_t *r)
     ngx_http_lua_loc_conf_t     *llcf;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (llcf->use_default_type) {
+    if (llcf->use_default_type
+        && r->headers_out.status != NGX_HTTP_NOT_MODIFIED)
```
agentzh: Style: a redundant line-trailing space here.

Sirupsen:
haha, now!

agentzh:
@Sirupsen Patch already applied to master. Thank you for your effort and patience!

Sirupsen:
Of course! Thanks for the pointers, there'll likely be a next time. You should add these tips to the README under `Test Suite`, or even better—add a `Contribution` section.

agentzh:
@Sirupsen Patches welcome :)

PR #469, opened 2015-03-14
add no_padding option to ngx_http_lua_ffi_encode_base64

yangshuxin:


agentzh:
Thanks for your patch! Already merged :)

PR #473, opened 2015-03-18
fix ssl_cert_by_lua_file can't work correctly issue

friendwu:
ssl_certificate_by_lua_file directive can't work correctly in server {...}

the error log is like this:
```
2015/03/13 17:42:13 [emerg] 5271#0: "ssl_certificate_by_lua_file" directive is not allowed here 
```


agentzh:
@friendwu Looking good to me. Will run some local tests and then merge it if nothing comes up. Thanks!

agentzh:
@friendwu Just applied your patch to the `ssl-cert-by-lua` branch with a corresponding test case. Thanks a lot!

PR #476, opened 2015-03-30
ngx.thread.kill - give all threads the permissions to kill each other

aviramc:
Removed the restriction that only a parent of a thread can kill it, making the ngx.thread.kill API a bit more like the UNIX-style interface. In my code, I have a case in which a thread should kill another (non child) thread.

### DIFF
--- a/src/ngx_http_lua_uthread.c
+++ b/src/ngx_http_lua_uthread.c
@@ -242,6 +242,12 @@ ngx_http_lua_uthread_kill(lua_State *L)
         return 2;
     }
 
+    if (ctx->cur_co_ctx == sub_coctx) {
+        lua_pushnil(L);
+        lua_pushliteral(L, "not allowed to kill self");
+        return 2;
+    }
+
     if (sub_coctx->pending_subreqs > 0) {
         lua_pushnil(L);
         lua_pushliteral(L, "pending subrequests");
--- a/src/ngx_http_lua_uthread.c
+++ b/src/ngx_http_lua_uthread.c
@@ -209,7 +209,7 @@ ngx_http_lua_uthread_kill(lua_State *L)
     lua_State                   *parent_co;
     ngx_http_request_t          *r;
     ngx_http_lua_ctx_t          *ctx;
-    ngx_http_lua_co_ctx_t       *coctx, *sub_coctx;
+    ngx_http_lua_co_ctx_t       *sub_coctx;
     ngx_http_lua_co_ctx_t       *parent_coctx;
 
     r = ngx_http_lua_get_req(L);
@@ -226,8 +226,6 @@ ngx_http_lua_uthread_kill(lua_State *L)
                                | NGX_HTTP_LUA_CONTEXT_ACCESS
                                | NGX_HTTP_LUA_CONTEXT_CONTENT
                                | NGX_HTTP_LUA_CONTEXT_TIMER);
-
-    coctx = ctx->cur_co_ctx;
 
     sub_co = lua_tothread(L, 1);
     luaL_argcheck(L, sub_co, 1, "lua thread expected");
--- a/src/ngx_http_lua_uthread.c
+++ b/src/ngx_http_lua_uthread.c
@@ -242,12 +242,6 @@ ngx_http_lua_uthread_kill(lua_State *L)
         return 2;
     }
 
-    if (sub_coctx->parent_co_ctx != coctx) {
-        lua_pushnil(L);
-        lua_pushliteral(L, "killer not parent");
-        return 2;
-    }
-
     if (sub_coctx->pending_subreqs > 0) {
         lua_pushnil(L);
         lua_pushliteral(L, "pending subrequests");
--- a/src/ngx_http_lua_uthread.c
+++ b/src/ngx_http_lua_uthread.c
@@ -89,7 +89,7 @@ ngx_http_lua_uthread_spawn(lua_State *L)
 
     ctx->cur_co_ctx->thread_spawn_yielded = 1;
 
-    if (ngx_http_lua_post_thread(r, ctx, ctx->cur_co_ctx) != NGX_OK) {
+    if (ngx_http_lua_post_thread(r, ctx, ctx->cur_co_ctx, 0) != NGX_OK) {
         return luaL_error(L, "no memory");
     }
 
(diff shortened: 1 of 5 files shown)


agentzh:
@aviramc Care must be take in the case when, for example, the parent is currently waiting on the thread being killed by a 3rd party. Will you add some test cases for such corner cases? I usually dare not merge anything that does not come with tests. Thanks!

aviramc:
@agentzh I will tests, but this may take some time. If you can review the changes beforehand, that would be great. This applies to all my open pull requests (there'll be more really soon)

agentzh:
@aviramc Yeah, sure. Thank you :)

agentzh:
@aviramc When you kill a thread being waited by its parent, I think you should notify the sleeping parent somehow.

aviramc:
@agentzh when you say a sleeping parent, do you mean a parent that is waiting on the thread?

agentzh:
@aviramc Yes, via `ngx.thread.wait()`.

aviramc:
@agentzh , please review commit  4ca58fe. In order to notify the parent thread that it the child thread has been killed, I've used posted `ngx_http_lua_post_thread()`. In order to be able to pass parameters to the posted thread, I've added the nrets attribute to the posted thread object. If you approve this fix, I'll starting writing some test cases for the feature.

src/ngx_http_lua_uthread.c
```diff
@@ -273,6 +273,15 @@ ngx_http_lua_uthread_kill(lua_State *L)
         ngx_http_lua_del_thread(r, L, ctx, sub_coctx);
         ctx->uthreads--;
 
+        /* notify the parent that the thread was killed */
+        parent_coctx = sub_coctx->parent_co_ctx;
+        if (sub_coctx->waited_by_parent && parent_coctx) {
+            parent_co = parent_coctx->co;
+            lua_pushboolean(parent_co, 0);
+            lua_pushstring(parent_co, "user thread killed");
+            ngx_http_lua_post_thread(r, ctx, parent_coctx, 2);
```
agentzh: Hmm, I'm a bit worried about potential race conditions here. Is it possible that the parent coroutine has already been posted at this point?

aviramc: @agentzh , currently, parents are posted only when their children die naturally. However, in order to protect such future cases, we can add a check in `ngx_http_lua_post_thread` itself, so that a thread won't be posted twice anyway (I don't see why that would be useful), or to check that here. What do you think?

agentzh: @aviramc When a light thread kills several siblings of it in series (without yielding in between) and their parent thread wait on all of them at the same time, then the parent thread will be posted here for multiple times, right? We need a test case for this corner case.

src/ngx_http_lua_uthread.c
```diff
@@ -273,6 +273,15 @@ ngx_http_lua_uthread_kill(lua_State *L)
         ngx_http_lua_del_thread(r, L, ctx, sub_coctx);
         ctx->uthreads--;
 
+        /* notify the parent that the thread was killed */
+        parent_coctx = sub_coctx->parent_co_ctx;
+        if (sub_coctx->waited_by_parent && parent_coctx) {
```
agentzh: @aviramc I think there is a corner case that a child light thread kills its parent and the parent is waiting on this child. We need to clear the `sub_coctx->waited_by_parent` field here at least, right? Since nobody is waiting on this killer thread any more. Thoughts?

src/ngx_http_lua_uthread.c
agentzh: @aviramc Also, maybe we should avoid the overhead of posting the parent thread if the killer is indeed the parent.

t/127-uthread-kill.t
```diff
@@ -507,3 +507,112 @@ thread created: zombie
...
+               ngx.sleep(1)
+               ngx.thread.kill(thread_to_kill)
+               ngx.say("Killing thread ended")
+            end
+            
+            function sleepy_thread()
+               local i
+               for i = 1, 3 do
+                  ngx.say("Sleepy thread going to sleep")
+                  ngx.sleep(3)
```
agentzh: @aviramc Can we use smaller sleeping intervals so that the tests can run much faster? Tests are meant to run over and over again, making them as fast as possible can be a big win :) I think times like `0.3` and `0.2` (seconds) should be good enough.

t/127-uthread-kill.t
```diff
@@ -507,3 +507,112 @@ thread created: zombie
 lua tcp socket abort resolver
 --- error_log
 
+=== TEST 10: non parent kill, parent waits for the killed
+--- config
+    location = /t {
+        content_by_lua '
+            function killing_thread(thread_to_kill)
+               ngx.sleep(1)
```
agentzh: @aviramc Can we use smaller sleeping intervals so that the tests can run much faster? Tests are meant to run over and over again, making them as fast as possible can be a big win :) I think times like 0.1 and 0.2 (seconds) should be long enough for our purposes.

t/127-uthread-kill.t
```diff
@@ -507,3 +507,112 @@ thread created: zombie
...
+[error]
+[alert]
+
+
+=== TEST 11: non parent kill, parent waits for the killer
+--- config
+    location = /t {
+        content_by_lua '
+            function killing_thread(thread_to_kill)
+               ngx.sleep(3)
```
agentzh: Let's use short sleeping times. 3 seconds in a single case are way too long for unit tests.

src/ngx_http_lua_uthread.c
```diff
@@ -242,9 +242,9 @@ ngx_http_lua_uthread_kill(lua_State *L)
         return 2;
     }
 
-    if (sub_coctx->parent_co_ctx != coctx) {
+    if (ctx->cur_co_ctx == sub_coctx) {
         lua_pushnil(L);
-        lua_pushliteral(L, "killer not parent");
+        lua_pushliteral(L, "not allowed to kill self");
```
agentzh: "not allowed to commit suicide" sounds better to me :)

agentzh:
@aviramc Sorry for the long delay on my side! I've made some more comments. Will you please look into them? Thanks a lot!

t/127-uthread-kill.t
```diff
@@ -507,3 +507,112 @@ thread created: zombie
...
+Sleepy thread going to sleep
+Parent now waits
+Killing thread ended
+Result: false user thread killed
+Finished
+
+--- no_error_log
+[error]
+[alert]
+
```
agentzh: @aviramc Will you please run the [reindex](https://github.com/agentzh/old-openresty/blob/master/bin/reindex) tool against the `.t` files you edited to ensure the formatting is good? Thanks!

Roycohen:
@agentzh can there be a progress on that matter.
It will be great if Openresty will close automatically all threads.

PR #480, opened 2015-04-01
Lua cosocket - added APIs to set each of the timeouts

aviramc:
Added APIs to set each of the socket timeouts separately. Today, only the method sock:settimeout exists, so there is no way to change the send, receive and connect timeouts separately (it is possible in the configuration).


src/ngx_http_lua_socket_tcp.c
```diff
@@ -2673,6 +2694,111 @@ ngx_http_lua_socket_tcp_settimeout(lua_State *L)
...
+
+    if (n != 2) {
+        return luaL_error(L, "ngx.socket set_receive_timeout: expecting at least 2 "
+                          "arguments (including the object) but seen %d",
+                          lua_gettop(L));
+    }
+
+    timeout = (ngx_int_t) lua_tonumber(L, 2);
+
+    lua_rawseti(L, 1, SOCKET_TIMEOUT_INDEX);
```
agentzh: This does not seem quit right, how will you define the relationship between the current `settimeout` and `setreadtimeout`, `setsendtimeout`, and `setconnecttimeout`? My hunch is that `settimeout` is just a shorthand of setting all 3 specialized timeout thresholds. What do you think?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2673,6 +2694,111 @@ ngx_http_lua_socket_tcp_settimeout(lua_State *L)
...
+
+    lua_rawgeti(L, 1, SOCKET_CTX_INDEX);
+    u = lua_touserdata(L, -1);
+
+    if (u) {
+        if (timeout > 0) {
```
agentzh: Maybe we should use the more compact `u->read_timeout = timeout > 0 ? (ngx_msec_t) timeout : u->conf->read_timeout` form here?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2673,6 +2694,111 @@ ngx_http_lua_socket_tcp_settimeout(lua_State *L)
...
+            u->read_timeout = (ngx_msec_t) timeout;
+        } else {
```
agentzh: Style: need a blank line before `} else {`. Nevertheless, I still think we can get rid of `if ... else ...` here via the ternary operator.

agentzh:
Again, we need some test cases to exercise various interesting corner cases and combinations :)

Thanks!

src/ngx_http_lua_socket_tcp.c
aviramc: @agentzh thanks for the feedback, I'll change that. Don't you think that `setreceivetimeout` is better than `setreadtimeout`?

agentzh: @aviramc Nope. We already have `receive` and `receiveuntil` methods and we may add more reading methods like those in the future. Don't want to give the user the impression that this timeout is only for the `receive()` method.

Also, we already have the `lua_socket_read_timeout` directive BTW :)

aviramc: @agentzh, I'm looking at the use of `SOCKET_TIMEOUT_INDEX` in order to revise that. It seems that the line `lua_rawgeti(L, 1, SOCKET_TIMEOUT_INDEX)` appears only when using `connect()`, but it is used anyway. I think it should be used on `connect` only when creating a new `ngx_http_lua_socket_tcp_upstream_t` object, meaning in line 573. In that case, I think it would be correct to split `SOCKET_TIMEOUT_INDEX` so that there would be an index for each of the connect, read and write actions. Do you agree?

aviramc:
@agentzh , I've changed the code according to your remarks, and added some test cases

thibaultcha:
This would be nice to have. Any plans to merge it?

agentzh:
@thibaultcha I've changed my mind a bit. Maybe it's better to add a `set_timeouts(conn_timeout, send_timeout, read_timeout)` similar to `ngx.balancer` module's `set_timeouts` method below?

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#set_timeouts

Will you work on a PR for this? Many thanks!

agentzh:
@thibaultcha It would be great if you can reuse @aviramc's changes and tests in this PR wherever possible.

thibaultcha:
Yes, this looks like a better API indeed.

> Will you work on a PR for this?

Will do.

agentzh:
@thibaultcha Awesome. Thanks a lot!

doujiang24:
@agentzh @thibaultcha Actually, I have already created a PR: https://github.com/openresty/lua-nginx-module/pull/850

spacewander:
@agentzh @doujiang24 
Maybe we could close this one because #850 is merged?

agentzh:
@spacewander Yes, right.

Issue #478, opened 2015-03-31
Possible SEGFAULT when using raw downstream socket in two threads


aviramc:
I'm currently reviewing the changes that @agentzh and I made simultaneously in order to start using the mainline ngx_lua. One of the changes in the full duplex cosocket objects, a feature that I use for the raw downstream socket in my code.

I have had the following race condition in my code (which I resolved), and it seems that it happens in the mainline code as well:
* There is a thread reading from the raw downstream socket.
* There is a thread writing to the raw downstream socket.
* The two threads are blocking, but the reading thread started to block **after** the writing thread did. This would set `ctx->downstream_co_ctx` to the reading thread.
* The write handler of the request runs (`ngx_http_lua_wev_handler`). Since there is data to write, it fetches the reading thread from `ctx->downstream_co_ctx` and tries to run it.

In this case, the wrong code to run (i.e. the reading thread instead of the writing), and this may cause SEGFAULTs. The proper fix is using two co_ctx pointers for the downstream socket - `downstream_read_co_ctx` for reading and `downstream_write_co_ctx` for writing.

@agentzh , I will post a pull request soon, but if you have time, I'd love to hear what you think about it. The case itself is hard to reproduce (has happened to me in a load test about a year ago).

agentzh:
@aviramc Yeah, I totally agree we need separate co_ctx pointers for the downstream socket as well. Patches welcome :)

aviramc:
@agentzh , after researching this a bit, I've seen that the above scenario is not possible, but a similar race condition is:
* There is a thread reading from the raw downstream socket.
* There is a thread writing to the raw downstream socket.
* The two threads are blocking, but the reading thread started to block **after** the writing thread did. This would set `ctx->downstream_co_ctx` to the reading thread.
* The reading thread ends its receive before the writing thread does, and starts blocking on something else. This makes `coctx->data` to be changed to something that is not the downstream socket, but a different object.
* The write handler of the request runs (`ngx_http_lua_wev_handler`). Since there is data to write, it assumes that `ctx->downstream_co_ctx->data` is the downstream socket, but it isn't since `ctx->downstream_co_ctx` is pointing to the receiving thread. It tries to run `u->write_event_handler(r, u)`, which would probably cause a segfault.

The fix to this race condition is the same fix I've applied in the pull request. The test that reproduces this will be committed soon.

agentzh:
@aviramc Excellent! Thanks for the hard work!

[...]

PR #481, opened 2015-04-01
Downstream socket full duplex bugfix

aviramc:
The suggested patch for issue #478

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -382,8 +382,10 @@ typedef struct ngx_http_lua_ctx_s {
 
     ngx_int_t                exit_code;
 
-    ngx_http_lua_co_ctx_t   *downstream_co_ctx; /* co ctx for the coroutine
-                                                   reading the request body */
+    ngx_http_lua_co_ctx_t   *downstream_read_co_ctx; /* co ctx for the coroutine
+                                                        reading the request body */
+    ngx_http_lua_co_ctx_t   *downstream_write_co_ctx; /* co ctx for the coroutine
+                                                        writing the request body */
 
     ngx_uint_t               index;              /* index of the current
                                                     subrequest in its parent
--- a/t/131-duplex-req-socket.t
+++ b/t/131-duplex-req-socket.t
@@ -1,0 +1,95 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+BEGIN {
+    if (!defined $ENV{LD_PRELOAD}) {
+        $ENV{LD_PRELOAD} = '';
+    }
+
+    if ($ENV{LD_PRELOAD} !~ /\bmockeagain\.so\b/) {
+        $ENV{LD_PRELOAD} = "mockeagain.so $ENV{LD_PRELOAD}";
+    }
+
+    if ($ENV{MOCKEAGAIN} eq 'r') {
+        $ENV{MOCKEAGAIN} = 'rw';
+
+    } else {
+        $ENV{MOCKEAGAIN} = 'w';
+    }
+
+    $ENV{TEST_NGINX_EVENT_TYPE} = 'poll';
+    $ENV{MOCKEAGAIN_WRITE_TIMEOUT_PATTERN} = 'slow';
+}
+
+use lib 'lib';
+use Test::Nginx::Socket::Lua;
+
+log_level('debug');
+
+repeat_each(1);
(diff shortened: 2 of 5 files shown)


agentzh:
@aviramc Will you mind adding one or a few test cases to the current test suite to exercise this bug?

aviramc:
@agentzh , I've added the required test case (verified that it fails without the fix)

src/ngx_http_lua_common.h
```diff
@@ -382,8 +382,10 @@ typedef struct ngx_http_lua_ctx_s {
 
     ngx_int_t                exit_code;
 
-    ngx_http_lua_co_ctx_t   *downstream_co_ctx; /* co ctx for the coroutine
-                                                   reading the request body */
+    ngx_http_lua_co_ctx_t   *downstream_read_co_ctx; /* co ctx for the coroutine
+                                                        reading the request body */
+    ngx_http_lua_co_ctx_t   *downstream_write_co_ctx; /* co ctx for the coroutine
```
agentzh: Hmm, seems like we can save some space by reusing the existing `read_co_ctx` and `write_co_ctx` fields here? We already have the `body_downstream` and `raw_downstream` flags to identify downstream sockets anyway :)

t/131-duplex-req-socket.t
```diff
@@ -0,0 +1,95 @@
...
+               local data, err, partial
+               -- First we receive in a blocking fashion so that ctx->downstream_co_ctx will be changed
+               data, err, partial = req_socket:receive(1)
+               if err ~= "timeout" then
+                  ngx.log(ngx.ERR, "Did not get timeout in the receiving thread!")
+                  return
+               end
+            
+               -- Now, sleep so that coctx->data is changed to sleep handler
+               ngx.sleep(3)
```
agentzh: Hmm, 3 seconds is quite some time. Can we reduce it a bit? Like 1 or 2 seconds?

t/131-duplex-req-socket.t
```diff
@@ -0,0 +1,95 @@
...
+repeat_each(1);
+
+plan tests => repeat_each() * (blocks() * 2);
+
+no_shuffle();
+run_tests();
+
+__DATA__
+
+=== TEST 18: downstream cosocket used in two different threads. See issue #481
```
agentzh: Test number is wrong. "18" looks like a copy&paste mistake :)

t/131-duplex-req-socket.t
```diff
@@ -0,0 +1,95 @@
...
+            
+            writer_thread = ngx.thread.spawn(writer, req_socket)
+            reader_thread = ngx.thread.spawn(reader, req_socket)
+            
+            ngx.thread.wait(writer_thread)
+            ngx.thread.wait(reader_thread)
+            ngx.log(ngx.INFO, "The two threads finished")
+';
+        }
+--- raw_request eval
```
agentzh: Maybe we can just use the following instead?

```
--- request
POST /test
--- more_headers
Content-Length: 1
```

It looks simpler to me :)

agentzh:
@aviramc Please look at my comments above. Thank you very much for your hard work. And I'm sorry for the long delay on my side.

t/131-duplex-req-socket.t
```diff
@@ -0,0 +1,95 @@
...
+               local bytes, err
+               -- send in a slow manner with a low timeout, so that the timeout handler will be
+               bytes, err = req_socket:send("slow!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+               if err ~= "timeout" then
+                  ngx.log(ngx.ERR, "Did not get timeout in the sending thread!")
+                  return
+               end
+            end
+            
+            local req_socket, err = ngx.req.socket(true)
```
agentzh: This test case the "raw" downstream cosocket. Can we add another test case for the more common "body" downstream cosocket? Thanks!

agentzh:
@aviramc Never mind, I've just committed a massively modified version of your patch to git master as commit 9e0503b03. Will you review and try it out on your side? I'd like to make a new release of ngx_lua very soon. Many thanks!

aviramc:
@agentzh , sorry for the late reply. Looks like an elegant solution for the problem. Soon I'll incorporate this into my code, but this looks like this would work with no problems.

agentzh:
@aviramc Cool, thanks!

Issue #482, opened 2015-04-02
Some Nginx built-in headers not set/unset properly


aviramc:
I've noticed that the header `If-Match` doesn't get properly unset by the lua module. It seems that it is because it is not a considered a built in header in the `ngx_http_lua_headers_in.c` module (the array `ngx_http_lua_set_handlers`).

In addition to that, comparing the built-in headers in the Nginx code (`src/http/ngx_http_request.c:80` in Nginx 1.7.10) to this repository's code, it seems that there are a lot of headers that aren't considered as built-ins as well (e.g. `If-Unmodified-Since`, `Accept-Language` and some more). Is there a particular reason for that or should that be updated? To me it seems like some kind of a bug.

agentzh:
@aviramc I think these were added to the nginx care after I implemented those parts in ngx_lua. That's why they are missing. Will you consider contributing a patch? ;) Thanks!

aviramc:
@agentzh , working on it. Are there any built-in headers that you explicitly want *not* not handle?

agentzh:
@aviramc Not any off the top of my head :)

agentzh:
Consider it resolved.

PR #491, opened 2015-04-14
Update builtin nginx headers

aviramc:
The following built-in headers will now be set in Nginx - `If-Unmodified-Since`, `If-Match`, `If-None-Match`, `Upgrade` and `X-Forwarded-For` (if X-Forwarded-For is enabled in compile time).

Note that the following built-in headers will still not be set by this patch:
* `Accept`, `Accept-Language` - it seems that there is not use in the built-in field in Nginx's source code, and so there's no way to test that (and it seems irrelevant anyway).
* The WebDAV headers `Depth`, `Destination`, `Overwrite` and `Date`.

@agentzh , notice that this patch includes a change in the function that sets the cookies header - it has been generalized to support all multiline built-in headers. I've verified that all the cookies header test pass.

Also, added tests to set and clear the headers that were added.

This patch should resolve issue #482.


agentzh:
Patch applied to git master in several separate commits. Please let me know if I've missed anything. Thanks for your contribution and sorry for the long delay on my side!

PR #493, opened 2015-04-18
Discard request-body if it's not read at all.

yangshuxin:
This is necessary in the case of pipelined request; the un-discarded request-body would otherwise linger to the next request, and will be mistakenly recognized as part of the header of the next pipelined request.

This change is to explicitly discard request-body for two directives: access_by_lua and content_by_lua. It leave out two situations without fixed: when the directive handler returns NGX_DONE and NGX_EAGAIN. Quite honestly, I don't know for sure how to fix right at this moment. But I think I catch the most common cases.

### DIFF
--- a/src/ngx_http_lua_accessby.c
+++ b/src/ngx_http_lua_accessby.c
@@ -230,9 +230,8 @@ ngx_http_lua_access_handler_file(ngx_http_request_t *r)
     return ngx_http_lua_access_by_chunk(L, r);
 }
 
-
 static ngx_int_t
-ngx_http_lua_access_by_chunk(lua_State *L, ngx_http_request_t *r)
+ngx_http_lua_access_by_chunk_helper(lua_State *L, ngx_http_request_t *r)
 {
     int                  co_ref;
     ngx_int_t            rc;
@@ -372,6 +371,22 @@ ngx_http_lua_access_by_chunk(lua_State *L, ngx_http_request_t *r)
 #endif
 
     return NGX_DECLINED;
+}
+
+static ngx_int_t
+ngx_http_lua_access_by_chunk(lua_State *L, ngx_http_request_t *r)
+{
+    ngx_int_t   rc, rc2;
+    rc = ngx_http_lua_access_by_chunk_helper(L, r);
+
+    if (r->keepalive &&
+        (rc >= NGX_HTTP_OK && rc < NGX_HTTP_SPECIAL_RESPONSE)) {
+        rc2 = ngx_http_discard_request_body(r);
+        if (rc2 != NGX_OK)
+            return rc2;
+    }
+
+    return rc;
 }
 
 /* vi:set ft=c ts=4 sw=4 et fdm=marker: */
--- a/t/044-req-body.t
+++ b/t/044-req-body.t
@@ -8,7 +8,7 @@ log_level('warn');
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 4 + 35);
+plan tests => repeat_each() * (blocks() * 4 + 33);
 
 #no_diff();
 no_long_string();
@@ -159,11 +159,8 @@ hiya, world"]
 hello, world",
 "POST /bar
 hiya, world"]
---- response_body eval
-["body: nil\n",
-qr/400 Bad Request/]
 --- error_code eval
-[200, 400]
+[200, 200]
 --- no_error_log
 [error]
 [alert]
@@ -1481,4 +1478,65 @@ Will you change this world?
 --- no_error_log
 [error]
 [alert]
+
+
+
+=== TEST 45: content_by_lua quit normally
+--- config
+    location /t {
+        content_by_lua '
+            ngx.exit(200)
+        ';
+    }
+
+--- pipelined_requests eval
+[
+"POST /t
+hello",
+"POST /t
+world"
+]
+--- response_body eval
+["",
+""]
+
+=== TEST 46: content_by_lua quit normally
+--- config
+    location /t {
+        content_by_lua '
+            return
+        ';
+    }
+
+--- pipelined_requests eval
+[
+"POST /t
+hello",
+"POST /t
+world"
+]
+--- response_body eval
+["",
+""]
+
+=== TEST 47: access_by_lua early abort
+-- ONLY
+--- config
+    location /t {
+        access_by_lua '
+            --ngx.req.read_body()
+            ngx.exit(200)
+        ';
+    }
+
+--- pipelined_requests eval
+[
+"POST /t
+hello",
+"POST /t
+world"
+]
+--- response_body eval
+["",
+""]
 
--- a/src/ngx_http_lua_contentby.c
+++ b/src/ngx_http_lua_contentby.c
@@ -103,6 +103,14 @@ ngx_http_lua_content_by_chunk(lua_State *L, ngx_http_request_t *r)
     rc = ngx_http_lua_run_thread(L, r, ctx, 0);
 
     if (rc == NGX_ERROR || rc >= NGX_OK) {
+        if (r->keepalive &&
+            (rc == NGX_OK ||
+            (rc < NGX_HTTP_SPECIAL_RESPONSE && rc >= NGX_HTTP_OK))) {
+            /* need to discard the request body (if any) if the return code
+             * indicates the request was successfully fullfilled.
+             */
+            return ngx_http_discard_request_body(r);
+        }
         return rc;
     }
 


jgrahamc:
Can you explain what will happen with NGX_DONE and NGX_EAGAIN?

yangshuxin:
Sorry for the late reply. I somehow missed your response yesterday.

They are two possible return values from the handler's helper functions.
NGX_AGAIN indicates things are not yet completely done.  Don't know for sure
in what situations, content_by_lua and access_by_lua's helper function 
return NGX_DONE.

In both cases, the handler call a "post-thread".  I have hard time to 
figure out where the
directives' handler completely finish. I will continue to dig with the 
testing cases shipped
with the lua_nginx_module.

However, as far as guess, often time the handler either return a value 
indicating something
wrong take place or they returns NGX_OK, http code (200, 301, 302 etc).  
If something wrong
take place, downstream code will discard request body (if any), so we 
don't need to worry about
for this situation.

Best Regards and Thank you for your response.

Shuxin

src/ngx_http_lua_accessby.c
```diff
@@ -374,4 +373,20 @@ ngx_http_lua_access_by_chunk(lua_State *L, ngx_http_request_t *r)
     return NGX_DECLINED;
 }
 
+static ngx_int_t
+ngx_http_lua_access_by_chunk(lua_State *L, ngx_http_request_t *r)
```
agentzh: This is too specific. And we should not handle this in specific phase handlers. Along this route, we'll also have to handle the rewrite phase.

yangshuxin: Never mind, the problem I ran into before is not about lua-nginx-module handler do not explictly discard request body; it is about Nginx itself does not discard the request body in the presence of custom error handler.

agentzh: @yangshuxin In that case, it's intentional. Because nginx delegates the decision into the content handler serving that location targeted by the custom error page (internal) redirection. And for ngx_lua, it's still the normal "fall-through" case :)

t/044-req-body.t
```diff
@@ -159,11 +159,8 @@ hiya, world"]
 hello, world",
 "POST /bar
 hiya, world"]
---- response_body eval
-["body: nil\n",
-qr/400 Bad Request/]
```
agentzh: It's bad to remove existing checks from the test case. We should fix the checks instead. I'll do that.

agentzh:
@yangshuxin How about the following patch instead?

```diff
diff --git a/src/ngx_http_lua_util.c b/src/ngx_http_lua_util.c
index 238e90e..0d83e93 100644
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -487,6 +487,16 @@ ngx_http_lua_send_chain_link(ngx_http_request_t *r, ngx_http_lua_ctx_t *ctx,
     }
 
     if (in == NULL) {
+        /* last buf to be sent */
+
+#if 1
+        if (!r->request_body && r == r->main) {
+            if (ngx_http_discard_request_body(r) != NGX_OK) {
+                return NGX_ERROR;
+            }
+        }
+#endif
+
         if (ctx->buffering) {
             rc = ngx_http_lua_send_http10_headers(r, ctx);
             if (rc == NGX_ERROR || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
diff --git a/t/023-rewrite/req-body.t b/t/023-rewrite/req-body.t
index dec093a..c085c0f 100644
--- a/t/023-rewrite/req-body.t
+++ b/t/023-rewrite/req-body.t
@@ -8,7 +8,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 2 + 1);
+plan tests => repeat_each() * (blocks() * 4 + 5);
 
 #no_diff();
 #no_long_string();
@@ -100,6 +100,9 @@ hello, world
 --- response_body
 hello, world
 sub: foo
+--- no_error_log
+[error]
+[alert]
 
 
 
@@ -123,3 +126,99 @@ Expect: 100-Continue
 [error]
 http finalize request: 500, "/test?" a:1, c:0
 
+
+
+=== TEST 6: not discard body (exit 200)
+--- config
+    location = /foo {
+        rewrite_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(200)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 7: not discard body (exit 201)
+--- config
+    location = /foo {
+        rewrite_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(201)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 8: not discard body (exit 302)
+--- config
+    location = /foo {
+        rewrite_by_lua '
+            -- ngx.req.discard_body()
+            -- ngx.say("body: ", ngx.var.request_body)
+            ngx.redirect("/blah")
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+[qr/302 Found/,
+"body: hiya, world\n",
+]
+--- error_code eval
+[302, 200]
+--- no_error_log
+[error]
+[alert]
+
diff --git a/t/024-access/req-body.t b/t/024-access/req-body.t
index e26a79c..4d8eb78 100644
--- a/t/024-access/req-body.t
+++ b/t/024-access/req-body.t
@@ -8,7 +8,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 2 + 1);
+plan tests => repeat_each() * (blocks() * 2 + 19);
 
 #no_diff();
 #no_long_string();
@@ -123,3 +123,99 @@ Expect: 100-Continue
 [error]
 http finalize request: 500, "/test?" a:1, c:0
 
+
+
+=== TEST 6: not discard body (exit 200)
+--- config
+    location = /foo {
+        access_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(200)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 7: not discard body (exit 201)
+--- config
+    location = /foo {
+        access_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(201)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 8: not discard body (exit 302)
+--- config
+    location = /foo {
+        access_by_lua '
+            -- ngx.req.discard_body()
+            -- ngx.say("body: ", ngx.var.request_body)
+            ngx.redirect("/blah")
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+[qr/302 Found/,
+"body: hiya, world\n",
+]
+--- error_code eval
+[302, 200]
+--- no_error_log
+[error]
+[alert]
+
diff --git a/t/044-req-body.t b/t/044-req-body.t
index 8c309a0..0246265 100644
--- a/t/044-req-body.t
+++ b/t/044-req-body.t
@@ -8,7 +8,7 @@ log_level('warn');
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 4 + 35);
+plan tests => repeat_each() * (blocks() * 4 + 51);
 
 #no_diff();
 no_long_string();
@@ -140,7 +140,7 @@ hiya, world"]
 
 
 
-=== TEST 6: not discard body
+=== TEST 6: not discard body (content_by_lua falls through)
 --- config
     location = /foo {
         content_by_lua '
@@ -161,9 +161,10 @@ hello, world",
 hiya, world"]
 --- response_body eval
 ["body: nil\n",
-qr/400 Bad Request/]
+"body: hiya, world\n",
+]
 --- error_code eval
-[200, 400]
+[200, 200]
 --- no_error_log
 [error]
 [alert]
@@ -1482,3 +1483,127 @@ Will you change this world?
 [error]
 [alert]
 
+
+
+=== TEST 45: not discard body (content_by_lua exit 200)
+--- config
+    location = /foo {
+        content_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(200)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 46: not discard body (content_by_lua exit 201)
+--- config
+    location = /foo {
+        content_by_lua '
+            -- ngx.req.discard_body()
+            ngx.say("body: ", ngx.var.request_body)
+            ngx.exit(201)
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+["body: nil\n",
+"body: hiya, world\n",
+]
+--- error_code eval
+[200, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 47: not discard body (content_by_lua exit 302)
+--- config
+    location = /foo {
+        content_by_lua '
+            -- ngx.req.discard_body()
+            -- ngx.say("body: ", ngx.var.request_body)
+            ngx.redirect("/blah")
+        ';
+    }
+    location = /bar {
+        content_by_lua '
+            ngx.req.read_body()
+            ngx.say("body: ", ngx.var.request_body)
+        ';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /bar
+hiya, world"]
+--- response_body eval
+[qr/302 Found/,
+"body: hiya, world\n",
+]
+--- error_code eval
+[302, 200]
+--- no_error_log
+[error]
+[alert]
+
+
+
+=== TEST 48: not discard body (custom error page)
+--- config
+    error_page 404 = /err;
+
+    location = /foo {
+        content_by_lua '
+            ngx.exit(404)
+        ';
+    }
+    location = /err {
+        content_by_lua 'ngx.say("error")';
+    }
+--- pipelined_requests eval
+["POST /foo
+hello, world",
+"POST /foo
+hiya, world"]
+--- response_body eval
+["error\n",
+"error\n",
+]
+--- error_code eval
+[404, 404]
+--- no_error_log
+[error]
+[alert]
```

yangshuxin:
Being a newbie of lua-nginx-module,  I can tell if this change fix the 
root cause.
However, I'm happy to see the TEST-48 which is to case we should avoid 
to discarding request body.


Thanks
Shuxin

agentzh:
@yangshuxin I've committed my modified version of the patch to master. Thanks!

PR #519, opened 2015-05-26
using luaL_Buffer on TCP cosocket instead of allocating new memory

lneto:


agentzh:
@lneto I originally tried using luaL_Buffer here but I had to remove it due to its severe limitation: "During its normal operation, a string buffer uses a variable number of stack slots. So, while using a buffer, you cannot assume that you know where the top of the stack is. You can use the stack between successive calls to buffer operations as long as that use is balanced; that is, when you call a buffer operation, the stack is at the same level it was immediately after the previous buffer operation."

This limitation is really a showstopper in this context

See http://www.lua.org/manual/5.1/manual.html#luaL_Buffer for more details.

lneto:
@agentzh this patch always finalizes the buffer on every call. Also, the only call that manipulates the stack in between of `luaL_buffinit` and `luaL_pushresult` calls is to `luaL_addlstring`. Thus, I don't see how one could mess (or even use) the stack among the buffer operations. Could you please show an example?

Thanks!

agentzh:
@lneto Okay, I see now. But still I'm not sure if using luaL_Buffer here is a good idea. The Lua buffer thing may still allocate new memory anyway and it has its own complications. And the direction is to move onto LuaJIT FFI so as to make cosocket calls JIT compiled.

lneto:
@agentzh `luaL_Buffer` would just allocate the definitive Lua string, avoding unecessary allocations. Also, it would prevent us from allocating a big buffer, pushing a big Lua string (which would also allocate memory) and copying data more than once.

> (...) and it has its own complications

Which complications? IMHO, it makes this function simpler (also smaller and more efficient =).

> And the direction is to move onto LuaJIT FFI so as to make cosocket calls JIT compiled.

I don't see why it should block this patch (while it hasn't moved yet).

agentzh:
@lneto Do you have any actual benchmark examples and results that can support your "more efficient" argument? I'd like to see them :)

lneto:
Unfortunately, I don't have =(. I can provide it later if you think it is worth. However, please notice that I'm not arguing that this patch adds significant performance gains. I meant _"more efficient"_ as _"it reduces the amount of allocated memory and data copying"_.

agentzh:
@lneto I've changed my mind. I'll look into your patch shortly :) Thanks!

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4908,65 +4901,30 @@ ngx_http_lua_socket_push_input_data(ngx_http_request_t *r,
     nbufs = 0;
     ll = NULL;
 
+    luaL_buffinit(L, &luabuf);
     for (cl = u->bufs_in; cl; cl = cl->next) {
+        ptrdiff_t chunk_size;
```
doujiang24: it's better to  use `size_t chunck_size`, and move after `size_t size`

doujiang24:
@agentzh this patch works fine on my side, and passed all of the test case in this repository

src/ngx_http_lua_socket_tcp.c
agentzh: Agreed. This is according to the nginx coding style (though declaring the local in the inner-most scope using it is actually more efficient from the compiler's perspective).

lneto: Could you clarify why it's better to use `size_t` to compute a difference between two pointers?

Could you point this style rule?

Thanks.

agentzh: @lneto Just look at the source code in the nginx core for both of these :) We just need to be consistent here.

lneto: hey @agentzh, IMHO, type choosing and variable scoping is not a matter of coding style. Honestly, I don't see how it is better to store the difference of two pointers in something else than `ptrdiff_t` (specially, using an unsigned type). Also, I don't see see how declaring a variable in the begin of a block is inconsistent with nginx style. Do you mean that we can use only the ANSI C subset that was previously used on the nginx core? (Sorry, but that doesn't make much sense to me =(

agentzh: @lneto I mean we need to be consistent with nginx's code convention. Being creative and distinguished here does more harm than good :)

agentzh: @lneto And BTW, I don't see why these two points cannot be part of "coding style".

lneto: > @lneto I mean we need to be consistent with nginx's code convention wherever possible. Being creative and distinguished here does more harm than good, no matter how much you like it personally.

There is nothing of creativity in using `ptrdiff_t` for pointers difference or on specifying the scope of a variable. Also, I don't see how you have concluded that it is a convention.

>@lneto And BTW, I don't see why these two points cannot be part of "coding style".

Simply because coding style is about syntax, not semantics.

agentzh: @lneto Well, just point me to a place in the nginx core that uses `ptrdiff_t` for similar usage. Also point me to a place where the nginx core declares a local variable in an inner code block scope (inside a C function).

agentzh: @lneto By convention, I mean we should code the same way as the original author and maintainers of the nginx core would code in this project.

agentzh: @lneto Well, you don't have to update the patch yourself. I can change it and merge this.

lneto: hey @agentzh, feel free to update the patch as you like..

agentzh:
A slightly modified version of this patch has been applied to git master. Thank you for your contribution.

lneto:
@agentzh you are welcome; btw, just be aware of:

>"When two pointers are subtracted, both shall point to elements of the same array object,
or one past the last element of the array object; the result is the difference of the
subscripts of the two array elements. The size of the result is implementation-defined,
and its type (a signed integer type) is ptrdiff_t defined in the <stddef.h> header.
*If the result is not representable in an object of that type, the behavior is undefined*.
(...)"  [ISO/IEC9899]

agentzh:
@lneto In this case, `b->last` is always smaller than `b->pos` so the result is always positive and always representable in the `size_t` type.

agentzh:
@lneto And in your own patch, you also unconditionally coalesce the `ptrdiff_t` type into `size_t` in the `luaL_addlstring` call anyway.

lneto:
hi @agentzh, both `size_t` and `ptrdiff_t` are implementation-defined; my point is that, although unlikely, the pointer difference can lead to an *unexpected behavior*. Differently from type casting two integer types, which is a well-defined operation.

PR #529, opened 2015-06-19
Fix built-in response headers not being accessible via lowercase (ngx.resp.get_headers)

GUI:
While most response headers were accessible via the original case or normalized lowercase (for example, `ngx.resp.get_headers()["Foo"]` and `ngx.resp.get_headers()["foo"]`), there were several built-in headers that were not available when trying to access the normalized, lowercase version of these headers. These special built-in headers were:

- Content-Length
- Content-Type
- Connection
- Transfer-Encoding

This fixes the issue so these response headers are available via both title case and now lowercase to match the behavior of the other response headers.

I wasn't able to get the full openresty test suite running locally, so hopefully this hasn't inadvertently affected anything else, but I was at least able to get the `t/016-resp-header.t` test suite running and passing with the additional tests added for this.

And this is based on the assumption that these headers should be accessible via lowercase or underscored names, as the [`ngx.req.get_headers`](https://github.com/openresty/lua-nginx-module#ngxreqget_headers) docs indicate all headers should be (but if this assumption is incorrect, just let me know).

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -488,7 +488,7 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
 
 #if 1
     if (r->headers_out.content_type.len) {
-        lua_pushliteral(L, "Content-Type");
+        lua_pushliteral(L, "content-type");
         lua_pushlstring(L, (char *) r->headers_out.content_type.data,
                         r->headers_out.content_type.len);
         lua_rawset(L, -3);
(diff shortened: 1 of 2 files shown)


agentzh:
@GUI Looks good to me. I'll take care of it tomorrow. Thanks a lot!

agentzh:
Patch applied to git master. Thanks!

GUI:
Thanks for the quick review and merge!

PR #538, opened 2015-07-27
Fix typo: use a double quote instead of a single quote

tatsuyafw:


agentzh:
Thanks for the catch!

PR #542, opened 2015-08-05
bugfix:ngx.encode_args() not escaped "|" to "%7c". see: https://en.wi…

goecho:
…kipedia.org/wiki/Percent-encoding


src/ngx_http_lua_util.c
```diff
@@ -1909,9 +1929,8 @@ ngx_http_lua_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
                     /* mail_auth is the same as memcached */
 
-    static uint32_t  *map[] =
-        { uri, args, html, refresh, memcached, memcached };
-
+	static uint32_t  *map[] =
```
agentzh: Please fix the indentation.

agentzh:
@goecho From what I can see. You're escaping not only `|` but also quite some others like `@`, `*`, `(`, `)`, `!`, and etc. So your commit log message is inaccurate.

It's worth mentioning that even Google Chrome's JavaScript API function encodeURIComponent() does not escape all your new candidates:

```
> encodeURIComponent(",*()'$!@`|")
"%2C*()'%24!%40%60%7C"
```

We needn't be harsher than Google Chrome I think :)

src/ngx_http_lua_util.c
```diff
@@ -2424,7 +2443,7 @@ ngx_http_lua_process_args_option(ngx_http_request_t *r, lua_State *L,
 
             if (total_escape) {
                 p = (u_char *) ngx_http_lua_escape_uri(p, key, key_len,
-                                                       NGX_ESCAPE_URI);
+                                                       NGX_ESCAPE_URI_COMPONENT);
```
agentzh: This line exceeds 80 columns (so do the following changes from you). Please fix them. Thanks.

agentzh:
@goecho Will you add a simple test case to the existing test case to cover these changes? Thanks!

goecho:
@agentzh I had added a simple test case to the existing test case to cover these changes. 

100% compatible with What Google Chrome's JavaScript API function encodeURIComponent() does. 

see:https://tools.ietf.org/html/rfc2396.

```javascript
>encodeURIComponent(",$@|`")
"%2C%24%40%7C%60"
>encodeURIComponent("-_.!~*'()")
"-_.!~*'()"
```

agentzh:
@goecho Cool, thanks. I'll look into this when I have a chance.

Tieske:
What's the status on this? We still got workarounds in place waiting for this to gets fixed....

toruneko:
2 years ago...
What's the status on this?

toruneko:
It seems that the same code with function `ngx_escape_uri` from file `ngx_string.c`.

agentzh:
OK, I've merged a slightly modified version of this patch. Sorry for the long delay on my side.

snpcp:
Hello, I using the **ngx.escape_uri** and **ngx.encode_args** function doesn't look right.

------
This is RFC 3986 : https://datatracker.ietf.org/doc/html/rfc3986#appendix-A
[[
reserved      = gen-delims / sub-delims
   gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
]]
-------

This is my code:
print(ngx.encode_args({a="1,2!3^4_5;6"}));
print(ngx.escape_uri("1,2!3^4_5"));
-----

This is output:
a=1%2C2!3%5E4_5%3B6
a%3D1%2C2!3%5E4_5%3B6
-----

The expected output is '1,2!3%5E_5;6' by RFC 3986, however that's not true.

zhuizhuhaomeng:
@snpcp where did you get this example?

zhuizhuhaomeng:
@snpcp  according to the rfc you given, the "," should be escaped.
but the expected output is not escaped. They are contradictory

snpcp:
In the fact, Many software doesn't strictly follow this standard(about gen-delims and sub-delims),  And the Openresty look like the same strategy as encodeURIComponent of JavaScript.
That being the case, I accepted. thanks for your response to me.

PR #544, opened 2015-08-17
optimize: reduce memory alloc when get peer from socket_pool

doujiang24:
1. reuse cleanup handler for upstream ctx in tcp socket
2. move ngx_palloc for host after ngx_http_lua_get_keepalive_peer

I'm not sure if this is an good idea, but usefull in my case:

```lua
local sock = ngx.socket.tcp()
-- in a long time loop
while true do
    sock:connect()
    -- do some socket communicate
    sock:setkeepalive()
    -- other thing for a period time
end
```

Alloc memory in lua space instead request pool if can not get peer from socket_pool, will be a more bigger change, I'm not sure if it's worth to do.
@agentzh Could you give me some advice? If yes, I will have a try :)

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -418,6 +418,7 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
     int                          key_index;
     const char                  *msg;
     ngx_http_lua_co_ctx_t       *coctx;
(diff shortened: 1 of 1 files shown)


agentzh:
@doujiang24 Thanks for the patch. I'll look into this as soon as I can manage.

doujiang24:
@agentzh Sorry, this patch may cause core dump in some case. I create a new pr when I fixed it

PR #555, opened 2015-08-28
Ngx shared rbtree

helloyi:
# Introduction

**directive:** *ngx_shared_rbtree <name> <size>;*

**get in ngx_lua:** *rbtree = ngx.shared.RBTREE*
**get in ngx_lua:** *rbtree = ngx.shared\[name_var\]*

**API:**

* insert
* get
* delete

**features:**

* support table as key of red-black tree node.
* support table as value of red-black tree node.
* support base store and access.

**how to use:**

* write a compare function and to compare your keys.
  **convention:** two argument of your key type,
  if first argument > second argument return 1;
  if first argument < second argument return -1;
  if first argument == second argument return 0;

* call the API to do what you want.

Here is an example:

```nginx
   http {
       lua_shared_rbtree rbtree 10m;
       server {
           location = /test {
               content_by_lua '
                   local cmp = function(a, b)
                       local x
                       if type(a) == "table" then
                           x = a[1]
                       else
                           x= a
                       end

                       if x > b[2] then
                           return 1

                       elseif x < b[1] then
                           return -1

                       else
                           return 0
                       end
                   end

                   local rbtree = ngx.shared.rbtree
                   local node = {}

                   node[1] = {1, 3}
                   node[2] = {1, 2, 3, "4", "5"}
                   node[3] = cmp
                   rbtree:insert(node)

                   node[1] = {4, 6}
                   node[2] = {k1 = "v1", k2="v2", k3="v3"}
                   node[3] = cmp
                   rbtree:insert(node)

                   node[1] = {7, 9}
                   node[2] = {1, "2", {3, "4", 5}}
                   node[3] = cmp
                   rbtree:insert(node)

                   node[1] = {11, 19}
                   node[2] = {{1, 2, 3}, {"1", "2", "3"}}
                   node[3] = cmp
                   rbtree:insert(node)

                   local val
                   val = rbtree:get{3, 1, cmp}
                   ngx.say(val, " ", type(val))
                   val = rbtree:get{3, 4,  cmp}
                   ngx.say(val, " ", type(val))

                   val = rbtree:get{5, "k1", cmp}
                   ngx.say(val, " ", type(val))
                   val = rbtree:get{{4, 6}, "k3", cmp}
                   ngx.say(val, " ", type(val))

                   val = rbtree:get{7, 3, cmp}
                   ngx.say(val[2], " ", type(val[2]))
                   val = rbtree:get{19, 1, cmp}
                   ngx.say(val[1], " ", type(val[1]))

               ';
           }
       }
   }
```

# API

## ngx.shared.RBTREE.insert

**syntax:** *success, message = ngx.shared.RBTREE:insert{key, value, compare_function}*

* `key`: key of insert node.
* `value`: value of insert node.
* `compare_function`: a function to compare two keys.

* `success`: boolean value to indicate whether the node is stored or not.
* `message`: textual error message, can be `"no memory"`.

## ngx.shared.RBTREE.get

**syntax:** *value, message = ngx.shrbtree.RBTREE:get{key, [field, compare_function}*

* `key`: key of want to get node.
* `field`: Optional, key of table that if the value is table type.
* `compare_function`: a function to compare two keys.

* `value`: value of get by *key and optional field*.
* `message`: textual error message, can be `"no memory"`.

## ngx.shared.RBTREE.delete

**syntax:** *success, message = ngx.shrbtree.RBTREE:delete{key, compare_function}*

* `key`: key of delete node.
* `compare_function`: a function to compare two keys.

* `success`: boolean value to indicate whether the node is delete or not.
* `message`: textual error message, can be `"no memory"`.

# implementation

## Why to use table as argument of API

The implementation of ngx.shared.RBTREE requires a *compare function*,
that is a base function of a workable *shrbtree*.
In fact, this is that to support key of table type.
Oh! it is a really wonderful demand.

There is usual way that calling lua from C:

1. Define a Global Variable as function in lua.
2. get the function from Global table.
3. calling it in lua stack.

But I don't like Global Variable, then I pass *compare function*
to the API.

1. Define a function in lua.
2. Pass the function to the API.
3. calling it in lua stack.

But after the call, the function pop from lua stack.
I need to call the function many times, so this way was rejected.

Almighty table! like the Titan, powerful but heavy.

1. Define a function in lua.
2. Pass table that contains the function to the API.
3. get the function from table of argument.
4. calling it in lua stack.

## handle error 

- Throw Exception, when some errors caused by users (example bad argument).
- Return Message, when some problems caused by API (example no shared memory).
- Crash, when some errors occurred that the program not be processed. (example call lua function with lua_call).

## How to store *lua tables*

I use red-black tree to store *lua tables*.

- The table context stored in a *rbtree node*.
- The table elements stored in a *rbtree*.

```
                                                                    +---------+ 
                    .                                               |         | ...
                    .                                               | element |
                    .                                               +---------+
               +---------------+                                     ^
               |               |                                    /
               | table context |----------------------->+---------+/
               +---------------+                        |         |
               /             \                          | element |
              /               \                         +---------+\
             V                 V                                    \ 
       +---------+       +---------+                                 V
       |         |       |         |                                +---------+
       |         |       |         |                                |         |
       +---------+       +---------+                                | element | ...
    ...                       ...                                   +---------+

```

## about ngx.shared.DICT and ngx.shared.RBTREE

Now, I inject shdict and shrbtree API to ngx.shared table,
for example:

```nginx
  ngx_shared_rbtree a 1m;
  ngx_shared_dict   b 1m;
  ngx_shared_rbtree c 1m;
  ngx_shared_dict   d 1m;
```

After conf parser, the ngx.shdict is:

```lua
   ngx.shdict = {a={zone1}, b={zone2}, c={zone3}, d={zone4}}
```

I think it's a good way:

- Semantically, the *shrbtree* is belong to *ngx.shared*.

- Semantically, the shared memory of *shrbtree* is belong to
*shm_zones* of *ngx_http_lua_main_conf_t*.

- In the implementation, this way doesn't need extra info,
and use of existing information efficiently.

Concrete:

```c
    /* ngx_http_lua_util.c */
    static void
    ngx_http_lua_inject_shared_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
    {
        /* ngx.shared table*/
        if (lmcf->shm_zones != NULL) {
            /* ngx.shared table exactly  */
            lua_createtable(L, 0, lmcf->shm_zones->nelts/* nrec */);
            ngx_http_lua_inject_shdict_api(lmcf, L);
            ngx_http_lua_inject_shrbtree_api(lmcf, L);
        } else {
            lua_newtable(L);
        }
        lua_setfield(L, -2, "shared");
    }
```

But, this changed ngx_lua source really.

Thanks!

### DIFF
--- a/config
+++ b/config
@@ -393,3 +393,7 @@ ngx_feature_test='setsockopt(1, SOL_SOCKET, SO_PASSCRED, NULL, 0);'
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_PATH='\"/usr/local/openresty/lualib/?.lua\"'"
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_CPATH='\"/usr/local/openresty/lualib/?.so\"'"
 
+
+ngx_lua_dquote='"'
+CFLAGS="$CFLAGS -DLUA_DEFAULT_PATH='$ngx_lua_dquote/usr/local/openresty/lualib/?.lua;/usr/local/openresty/lualib/?/init.lua$ngx_lua_dquote'"
+CFLAGS="$CFLAGS -DLUA_DEFAULT_CPATH='$ngx_lua_dquote/usr/local/openresty/lualib/?.so$ngx_lua_dquote'"
--- a/config
+++ b/config
@@ -268,6 +268,7 @@ NGX_ADDON_SRCS="$NGX_ADDON_SRCS \
                 $ngx_addon_dir/src/ngx_http_lua_pcrefix.c \
                 $ngx_addon_dir/src/ngx_http_lua_headerfilterby.c \
                 $ngx_addon_dir/src/ngx_http_lua_shdict.c \
+                $ngx_addon_dir/src/ngx_http_lua_shrbtree.c \
                 $ngx_addon_dir/src/ngx_http_lua_socket_tcp.c \
                 $ngx_addon_dir/src/ngx_http_lua_api.c \
                 $ngx_addon_dir/src/ngx_http_lua_logby.c \
@@ -320,6 +321,7 @@ NGX_ADDON_DEPS="$NGX_ADDON_DEPS \
                 $ngx_addon_dir/src/ngx_http_lua_pcrefix.h \
                 $ngx_addon_dir/src/ngx_http_lua_headerfilterby.h \
                 $ngx_addon_dir/src/ngx_http_lua_shdict.h \
+                $ngx_addon_dir/src/ngx_http_lua_shrbtree.h \
                 $ngx_addon_dir/src/ngx_http_lua_socket_tcp.h \
                 $ngx_addon_dir/src/api/ngx_http_lua_api.h \
                 $ngx_addon_dir/src/ngx_http_lua_logby.h \
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -1473,7 +1473,7 @@ Uses Lua code specified in <code><lua-script-str></code> to define an output hea
 Note that the following API functions are currently disabled within this context:
 
 * Output API functions (e.g., [[#ngx.say|ngx.say]] and [[#ngx.send_headers|ngx.send_headers]])
-* Control API functions (e.g., [[#ngx.exit|ngx.exit]] and [[#ngx.exec|ngx.exec]])
+* Control API functions (e.g., [[#ngx.redirect|ngx.redirect]] and [[#ngx.exec|ngx.exec]])
 * Subrequest API functions (e.g., [[#ngx.location.capture|ngx.location.capture]] and [[#ngx.location.capture_multi|ngx.location.capture_multi]])
 * Cosocket API functions (e.g., [[#ngx.socket.tcp|ngx.socket.tcp]] and [[#ngx.req.socket|ngx.req.socket]]).
 
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -24,6 +24,7 @@
 #include "ngx_http_lua_initby.h"
 #include "ngx_http_lua_initworkerby.h"
 #include "ngx_http_lua_shdict.h"
+#include "ngx_http_lua_shrbtree.h"
 
 
 #if defined(NDK) && NDK
@@ -92,7 +93,7 @@ ngx_http_lua_shared_dict(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
     ctx->log = &cf->cycle->new_log;
 
     zone = ngx_shared_memory_add(cf, &name, (size_t) size,
-                                 &ngx_http_lua_module);
+                                 &ngx_http_lua_module_shdict);
     if (zone == NULL) {
         return NGX_CONF_ERROR;
     }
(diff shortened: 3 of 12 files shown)


agentzh:
@helloyi As mentioned in the other ticket, I'd like to introduce a C API in ngx_lua to allow other nginx C modules to provide additional shm-based storage APIs without changing ngx_lua.

Basically I'd like to keep ngx_lua's core small and encourage the community to provide as many alternatives (as external plugins) as possible :) What do you think?

helloyi:
@agentzh 
## What I understand
Yeah, I agree with you! Based on my understanding, the advantage is:
1. Avoid introducing bugs.
2. Easy to manage.
3. More motivating.
Keep small keep pretty, this is really very nice!

## What I think
I have read `ngx_http_lua_api.h`.
if make ngx.shared.RBTREE as a third-party module (the following call it `ngx-lua-shared-rbtree`).
I think I should do this:
1. Define `lua_shared_rbtree` directive in `ngx-lua-shared-rbtree` module.
2. Management of shared memory is in `ngx-lua-shared-rbtree` module.
3. Add a lua module (like `os`) that is `shared_rbtree` via `ngx_http_lua_add_package_preload()`.

Here is a fake example:
```lua
 http {
     lua_shared_rbtree dogs 10m;
     lua_shared_rbtree pigs 10m;
     server {
         location /test {
             content_by_lua '
                 local shrbtrees = require("shared_rbtree")

                 local cmp = function(s1, s2)
                      if s1 > s2 then
                          return 1
                      elseif s1 < s2 then
                          return -1
                      else
                          return 0
                      end 
                  end

                  local dogs = shrbtrees.dogs
                  dogs:insert{"Jim", 8, cmp}
                  ngx.say(dogs:get{"Jim", cmp})

                  local pigs = shrbtrees.pigs
                  pigs:insert{"Tim", 1, cmp}
                  ngx.say(pigs:get{"Tim", cmp})
             ';
         }

     }
 }
```

Thanks!

agentzh:
@helloyi I think it's better to let ngx_lua allocate and initialize your shm zones by registering your zone via some exposed C API to the ngx_lua core. We need special shm zone initialization logic to ensure all the shm zones are properly initialized before the Lua code in init_by_lua is run properly.

Will you consider submitting a patch to ngx_lua for such a shm zone registration C API?

helloyi:
@agentzh I love to do this! Wait for my good news!

agentzh:
@helloyi Great. Thanks!

helloyi:
nginx core allocate and initialize shm zone:
1. call `ngx_shared_memory_add`.
2. set initialize function of shm zone.

ngx_lua core allocate and initialize shm zone:
1. call `ngx_http_lua_shared_memory_add`.
2. set initialize function of shm_zone.

A prime difference is that, if we use ngx core to allocate and initialize shm zone,
the shm zone will be managed of ngx_lua. And this is our purpose.

Here is `ngx_http_lua_shared_memory_add`:

```c++
    ngx_shm_zone_t *
    ngx_http_lua_shared_memory_add(ngx_conf_t *cf,
                                   ngx_str_t *name,
                                   size_t size,
                                   void *tag)
    {
        ngx_http_lua_main_conf_t *lmcf;
        ngx_shm_zone_t           **zp;

        lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module);
        if (lmcf == NULL) {
            return NULL;
        }

        if (lmcf->shm_zones == NULL) {
            lmcf->shm_zones = ngx_palloc(cf->pool, sizeof(ngx_array_t));
            if (lmcf->shm_zones == NULL) {
                return NULL;
            }

            if (ngx_array_init(lmcf->shm_zones, cf->pool, 2,
                               sizeof(ngx_shm_zone_t *))
                != NGX_OK)
                {
                    return NULL;
                }
        }

        zone = ngx_http_lua_find_zone(name->data, name->len);
        if (zone != NULL) {
            return zone;
        }

        zone = ngx_shared_memory_add(cf, name, (size_t) size, tag);
        if (zone == NULL) {
            return NULL;
        }

        zp = ngx_array_push(lmcf->shm_zones);
        if (zp == NULL) {
            return NULL;
        }

        *zp = zone;

        lmcf->requires_shm = 1;

        return zone;
    }
```
Here are some changes, or an example.
```c++
    char *
    ngx_http_lua_shared_dict(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
    {
        /* parse directive and other stuff */
        ...

        /* call ngx_http_lua_shared_memory_add */
        zone = ngx_http_lua_shared_memory_add(cf, &name, (size_t)size,

        ...

        /* set initialize function */
        zone->init = ngx_http_lua_shdict_init_zone;
        zone->data = ctx;

        return NGX_CONF_OK;
    }
```

Do you have anything else consideration that ngx_lua core allocate and initialize shm zone?

agentzh:
@helloyi Looking good to me despite the lack of error messages in case of zones a duplicate name.

helloyi:
without `ngx_http_lua_find_zone`:
```c++
    ngx_shm_zone_t *
    ngx_http_lua_shared_memory_add(ngx_conf_t *cf,
                                   ngx_str_t *name,
                                   size_t size,
                                   void *tag)
    {
        ngx_http_lua_main_conf_t *lmcf;
        ngx_shm_zone_t           **zp;

        lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module);
        if (lmcf == NULL) {
            return NULL;
        }

        if (lmcf->shm_zones == NULL) {
            lmcf->shm_zones = ngx_palloc(cf->pool, sizeof(ngx_array_t));
            if (lmcf->shm_zones == NULL) {
                return NULL;
            }

            if (ngx_array_init(lmcf->shm_zones, cf->pool, 2,
                               sizeof(ngx_shm_zone_t *))
                != NGX_OK)
                {
                    return NULL;
                }
        }

        /* zone = ngx_http_lua_find_zone(name->data, name->len); */
        /* if (zone != NULL) { */
        /*    return zone; */
        /* } */

        zone = ngx_shared_memory_add(cf, name, (size_t) size, tag);
        if (zone == NULL) {
            return NULL;
        }

        zp = ngx_array_push(lmcf->shm_zones);
        if (zp == NULL) {
            return NULL;
        }

        *zp = zone;

        lmcf->requires_shm = 1;

        return zone;
    }
```

I think this is adequate. Because it ensures that share memory is added to `lmcf->shm_zones` and is also added to `cf->shm_zone`, uless `ngx_array_push(lmcf->shm_zones)` reture `NULL`.

Thanks!

agentzh:
@helloyi Agreed. Now we rely on `ngx_shared_memory_add` to log error messages for conflicting zone names. 

Will you create a separate pull request for this new C API of ngx_lua for plugging new shm-based Lua storage components?

splitice:
This is awesome.

Personally I think it should be in its own namespace, i.e ngx.rbtree as there are less API methods implemented. And just for clarity.

helloyi:
@splitice  Oh, It has been a long time since I was coding ngx_lua, haha. And you could look into  [Shm zones registration capi #559 ](https://github.com/openresty/lua-nginx-module/pull/559) PR. :)

PR #557, opened 2015-08-31
Issue548

helloyi:
Belated pull request.

A potential problem is that the metatable of lightuserdata would changed in "shdict", if it set metatable for lightuserdata that anywhere in ngx_lua's VM. The reason is lightuserdata share the metatable globally.

test:
```shell
    $ PATH=${PATH}:/usr/local/openresty/nginx/sbin prove  -I./t/lib  t/126-shdict- frag.t t/043-shdict.t t/080-hup-shdict.t 
    t/126-shdict-frag.t .. ok     
    t/043-shdict.t ....... ok       
    t/080-hup-shdict.t ... ok     
    All tests successful.
    Files=3, Tests=320, 40 wallclock secs ( 0.14 usr  0.01 sys +  4.49 cusr  1.49  csys =  6.13 CPU)
    Result: PASS
```

I find some test-suits rely on external configuration of nginx.
Could you please provide your nginx configuration in `t/servroot`.

Thanks!


.gitignore
```diff
@@ -162,3 +162,7 @@ tthread
 addr2line
 hup
 theaders
+
```
agentzh: Hmm, this change is unrelated and should be excluded.

src/ngx_http_lua_shdict.c
```diff
@@ -347,10 +347,13 @@ ngx_http_lua_inject_shdict_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
             lua_pushlstring(L, (char *) ctx->name.data, ctx->name.len);
                 /* shared mt key */
 
+            lua_createtable(L, 1 /* narr */, 0 /* nrec */); /* table of zone[i] */
             lua_pushlightuserdata(L, zone[i]); /* shared mt key ud */
+            lua_rawseti(L, -2, 1); /* {zone[i]} */
```
agentzh: I think it's better to replace the magic number index `1` with a named enum name. For example,

```C
enum {
    SHDICT_USERDATA_INDEX = 1
};
```

near the beginning of this `.c` file and then write

```C
            lua_rawseti(L, -2, SHDICT_USERDATA_INDEX); /* {zone[i]} */
```

What do you think?

src/ngx_http_lua_shdict.c
```diff
@@ -347,10 +347,13 @@ ngx_http_lua_inject_shdict_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
...
             lua_pushvalue(L, -3); /* shared mt key ud mt */
             lua_setmetatable(L, -2); /* shared mt key ud */
             lua_rawset(L, -4); /* shared mt */
+                /* name = {zone[i]} */
```
agentzh: Strange indentation here.

src/ngx_http_lua_shdict.c
```diff
@@ -401,7 +404,13 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
                           "but only seen %d", n);
     }
 
-    zone = lua_touserdata(L, 1);
+    if (LUA_TTABLE != lua_type(L, 1)) {
```
agentzh: Style: nginx always place constants to be compared on the right hand side of the comparison operator. I do understand the rationale behind this style but modern C compilers already emit this kind of typos (using `=` for `==`) with `-Wall`, like gcc.

src/ngx_http_lua_shdict.c
```diff
@@ -401,7 +404,13 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
...
+        return luaL_error(L, "bad \"zone\" argument");
+    }
+
+    lua_rawgeti(L, 1, 1);
```
agentzh: We should avoid the magic number index here as well (see above).

src/ngx_http_lua_shdict.c
```diff
@@ -401,7 +404,13 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
...
+    zone = lua_touserdata(L, -1);
```
agentzh: Better check if the returned `zone` is NULL.

src/ngx_http_lua_shdict.c
```diff
@@ -619,9 +630,11 @@ ngx_http_lua_shdict_flush_expired(lua_State *L)
         return luaL_error(L, "expecting 1 or 2 argument(s), but saw %d", n);
     }
 
-    luaL_checktype(L, 1, LUA_TLIGHTUSERDATA);
+    luaL_checktype(L, 1, LUA_TTABLE);
 
-    zone = lua_touserdata(L, 1);
+    lua_rawgeti(L, 1, 1);
+    zone = lua_touserdata(L, -1);
+    lua_pop(L, 1);
```
agentzh: Because this code sequence (the 3 lines above) appears frequently, it makes sense to abstract it away with a static (inline) C function like `ngx_http_lua_shdict_get_zone`. What do you think?

agentzh:
@helloyi Thank you for the patch! Will you address the issues in my comments? Thanks!

.gitignore
helloyi: Haha, yeah, it's personal. This should not happen.

src/ngx_http_lua_shdict.c
helloyi: This is good style. abstract, when necessary.

src/ngx_http_lua_shdict.c
helloyi: Perhaps the code more direct.

src/ngx_http_lua_shdict.c
helloyi: It's really a stale style.

src/ngx_http_lua_shdict.c
helloyi: It's great that taking into account both of readability and efficiency.

helloyi:
Thank you very much for your advice and patient!

src/ngx_http_lua_shdict.c
```diff
@@ -34,12 +34,17 @@ static int ngx_http_lua_shdict_delete(lua_State *L);
 static int ngx_http_lua_shdict_flush_all(lua_State *L);
 static int ngx_http_lua_shdict_flush_expired(lua_State *L);
 static int ngx_http_lua_shdict_get_keys(lua_State *L);
+static inline ngx_shm_zone_t* ngx_http_lua_shdict_get_zone(lua_State *L,
+                                                           int index);
 
 
 #define NGX_HTTP_LUA_SHDICT_ADD         0x0001
 #define NGX_HTTP_LUA_SHDICT_REPLACE     0x0002
 #define NGX_HTTP_LUA_SHDICT_SAFE_STORE  0x0004
```
agentzh: Style: two blank lines needed here.

src/ngx_http_lua_shdict.c
```diff
@@ -34,12 +34,17 @@ static int ngx_http_lua_shdict_delete(lua_State *L);
...
 
+enum {
+    SHDICT_USERDATA_INDEX = 1,
+};
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
```diff
@@ -376,6 +383,17 @@ ngx_http_lua_shdict_get_stale(lua_State *L)
   return ngx_http_lua_shdict_get_helper(L, 1 /* stale */);
 }
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
```diff
@@ -376,6 +383,17 @@ ngx_http_lua_shdict_get_stale(lua_State *L)
...
+ngx_http_lua_shdict_get_zone(lua_State *L, int index)
+{
+    ngx_shm_zone_t *zone;
+
+    lua_rawgeti(L, index, SHDICT_USERDATA_INDEX);
+    zone = lua_touserdata(L, -1);
+    lua_pop(L, 1);
+
+    return zone;
+}
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
```diff
@@ -376,6 +383,17 @@ ngx_http_lua_shdict_get_stale(lua_State *L)
   return ngx_http_lua_shdict_get_helper(L, 1 /* stale */);
 }
 
+static inline ngx_shm_zone_t *
```
agentzh: Style: we should use `ngx_inline` instead of `inline` here.

agentzh:
@helloyi Thank you for the quick fixes! Please address the remaining minor style issues :) Thanks!

src/ngx_http_lua_shdict.c
helloyi: Ok! Thanks! :)

agentzh:
@helloyi Looks good to me now. One last thing, sorry, will you please add a couple of test cases for the issue? Maybe we could add the following 2 test cases to `t/043-shdict.t`:

1. check if the `ngx.shared.DICT` value type is a table.
2. check if the lightuserdata `ngx.null` has methods of shared dicts.

We need to ensure these tests indeed fail with the older version though :)

Thanks a lot!

helloyi:
@agentzh I feel that I leant a lot from there, thanks a lot!

agentzh:
@helloyi Great! Looks perfect now :) Thanks for your patience and hard work! I'll merge this thing in the coming days.

src/ngx_http_lua_shdict.c
```diff
@@ -347,7 +356,9 @@ ngx_http_lua_inject_shdict_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
             lua_pushlstring(L, (char *) ctx->name.data, ctx->name.len);
                 /* shared mt key */
 
+            lua_createtable(L, 1 /* narr */, 0 /* nrec */); /* table of zone[i] */
```
agentzh: Style: line exceeding 80 columns. I'll fix it locally during the merge :)

src/ngx_http_lua_shdict.c
```diff
@@ -401,7 +425,11 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
                           "but only seen %d", n);
     }
 
-    zone = lua_touserdata(L, 1);
+    if (lua_type(L, 1) != LUA_TTABLE ) {
```
agentzh: Style: extra space before the `)` character. I'll fix it locally during the merge anyway.

src/ngx_http_lua_shdict.c
```diff
@@ -36,11 +36,20 @@ static int ngx_http_lua_shdict_flush_expired(lua_State *L);
 static int ngx_http_lua_shdict_get_keys(lua_State *L);
 
 
+static ngx_inline ngx_shm_zone_t* ngx_http_lua_shdict_get_zone(lua_State *L,
```
agentzh: Style: Line exceeding 80 columns and also we need to swap the `*` and space character right before the function name.

agentzh:
@helloyi It's worth mentioning that we usually use the [ngx-releng](https://github.com/openresty/nginx-devel-utils/blob/master/ngx-releng) tool to catch a few categories of common coding style mistakes automatically for us. Well, just a good trick to save the time ;)

src/ngx_http_lua_shdict.c
```diff
@@ -36,11 +36,20 @@ static int ngx_http_lua_shdict_flush_expired(lua_State *L);
...
+                                                           int index);
```
agentzh: Style: the parameter declarations need to align with the first argument in the previous line.

t/043-shdict.t
```diff
@@ -2298,3 +2298,100 @@ foo = nil
 --- no_error_log
 [error]
 
+
+
+=== TEST 86: the lightuserdata ngx.null hasn't methods of shared dicts.
```
agentzh: This test does not exercise the bug and also passes successfully with the current master. This is because you did not define any `lua_shared_dict` in this test case, thus no metatables registered on any light user data.

This is why we need to run the tests again the old code version to ensure that they actually fail :) Tests that always pass are meaningless :)

t/043-shdict.t
```diff
@@ -2298,3 +2298,100 @@ foo = nil
...
+    }
+--- request
+GET /test
+--- response_body_like: 500 Internal Server Error
+--- error_code: 500
+--- error_log
+
+
+
+=== TEST 90: value is a table
```
agentzh: This test case is unrelated to the current PR and should be excluded from this patch.

agentzh:
@helloyi A slightly modified version of your latest patch has just been applied to master. Thanks!

helloyi:
@agentzh Thank you very much for the review and merge.

PR #558, opened 2015-09-02
Shm zones registration capi

helloyi:
@agentzh Oh, Sorry, there is a issue that `shm-zones-registration-capi` branch is base on `issue548`.
After I will fix. And the next capi of shm_zone registration is focus.

### need:
ngx_lua allocate and initialize shared memory that is need for users.

### needed:
User need to ensure all the shm zones are properly initialized before the Lua code in init_by_lua runs
that is needed by users.

like this:

```c++
    lmcf->shm_zones_inited++;

    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts
        && lmcf->init_handler)
    {
        if (lmcf->init_handler(ctx->log, lmcf, lmcf->lua) != NGX_OK) {
            /* an error happened */
            return NGX_ERROR;
        }
    }

```

### So
1. define `ngx_http_lua_shared_memory_add` as need for user.
2. define `ngx_http_lua_shared_memory_init_` to avoid needed by user.
3. define `ngx_http_lua_shm_zone_ctx_t` as data for shm_zone,
and store `shm_zone` `data` `init function` of user.

### processes

add:
1. user call `ngx_http_lua_shared_memory_add`.
2. `ngx_http_lua_shared_memory_add` store `shm_zone` of user in `ngx_http_lua_shm_zone_ctx_t`.
3. `ngx_http_lua_shared_memory_add` set `ngx_http_lua_shared_memory_init` to `shm_zone` of ngx_lua.
4. `ngx_http_lua_shared_memory_add` set `ngx_http_lua_shm_zone_ctx_t` to `shm_zone` of ngx_lua.
5. user set `data` and `init` for `shm_zone` of self.

init:
1. ngx_core call `ngx_http_lua_shared_memory_init` to init shm_zone of ngx_lua.
2. `ngx_http_lua_shared_memory_init` call `user's init`.
3. `ngx_http_lua_shared_memory_init` ensure all the shm zones are properly initialized before the Lua code in init_by_lua runs.

Thinks!

### DIFF
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -124,9 +124,8 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf,
     }
 
     if (zone->data) {
-        ngx_conf_log_error(NGX_LOG_EMERG, cf, 0,
-                           "share memory \"%V\" of ngx_lua is already defined" , name);
-        return NULL;
+        ctx = (ngx_http_lua_shm_zone_ctx_t *)zone->data;
+        return (ngx_shm_zone_t *)&ctx->data;
     }
 
     n = offsetof(ngx_http_lua_shm_zone_ctx_t, data)
@@ -156,31 +155,6 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf,
 
     return *zp;
 }
-
-
-/* ngx_int_t */
-/* ngx_http_lua_shared_memory_inited(ngx_log_t *log, ngx_conf_t *cf) */
-/* { */
-/*     ngx_http_lua_main_conf_t *lmcf; */
-
-    /* lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module); */
-    /* if (lmcf == NULL) { */
-    /*     return NGX_ERROR; */
-    /* } */
-
-    /* lmcf->shm_zones_inited++; */
-
-    /* if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts */
-    /*     && lmcf->init_handler) */
-    /* { */
-    /*     if (lmcf->init_handler(cf->log, lmcf, lmcf->lua) != NGX_OK) { */
-    /*         /\* an error happened *\/ */
-    /*         return NGX_ERROR; */
-    /*     } */
-    /* } */
-
-/*     return NGX_OK; */
-/* } */
 
 
 static ngx_int_t
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -11,6 +11,16 @@
 #include "ngx_http_lua_util.h"
 
 
+typedef struct ngx_http_lua_shm_zone_ctx_s {
+    ngx_log_t                *log;
+    ngx_http_lua_main_conf_t *lmcf;
+
+    u_char                    data; /* ngx_shm_zone_t */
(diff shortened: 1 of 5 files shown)


src/ngx_http_lua_api.c
```diff
@@ -74,4 +86,123 @@ ngx_http_lua_add_package_preload(ngx_conf_t *cf, const char *package,
     return NGX_OK;
 }
 
+
+ngx_shm_zone_t *
+ngx_http_lua_shared_memory_add(ngx_conf_t *cf,
+                               ngx_str_t *name,
+                               size_t size,
+                               void *tag)
```
agentzh: Style: better avoid unnecessary line breaks above unless 78 columns are to be exceeded otherwise.

agentzh:
@helloyi Thanks for the patch! Will you also include a *minimal* and standalone fake shm-storage module in the test suite? The module source itself can be put under `t/data/` I suppose. I hope that everything is covered by the tests :)

src/ngx_http_lua_shdict.c
```diff
@@ -376,6 +363,17 @@ ngx_http_lua_shdict_get_stale(lua_State *L)
   return ngx_http_lua_shdict_get_helper(L, 1 /* stale */);
 }
```
agentzh: style: two blank lines needed here.

src/ngx_http_lua_shdict.c
```diff
@@ -376,6 +363,17 @@ ngx_http_lua_shdict_get_stale(lua_State *L)
...
 
+static inline ngx_shm_zone_t *
```
agentzh: Style: we need `ngx_inline` instead of `inline`.

agentzh: Ah, okay, it seems like an old base for the other PR.

src/ngx_http_lua_api.c
```diff
@@ -11,6 +11,18 @@
...
+typedef struct ngx_http_lua_shm_zone_ctx_s {
+    ngx_log_t                *log;
+    ngx_http_lua_main_conf_t *lmcf;
+
+    u_char                    data; /* ngx_shm_zone_t */
+} ngx_http_lua_shm_zone_ctx_t;
+
+
+static ngx_int_t
+ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data);
```
agentzh: Style: we don't need a line break here for C function declarations (we only need the break after the return value type for C function *definitions*).

src/ngx_http_lua_api.c
```diff
@@ -11,6 +11,18 @@
...
 
+typedef struct ngx_http_lua_shm_zone_ctx_s {
+    ngx_log_t                *log;
+    ngx_http_lua_main_conf_t *lmcf;
+
+    u_char                    data; /* ngx_shm_zone_t */
+} ngx_http_lua_shm_zone_ctx_t;
+
+
+static ngx_int_t
```
agentzh: Style: we don't need a line break here for C function declarations (we only need the break after the return value type for C function definitions).

agentzh:
@helloyi Please rebase onto the latest git master branch to help review. Thanks!

helloyi:
@agentzh I'm very sorry that I made inconvenience for you. I will fix these issues, quickly.

helloyi:
@agentzh Sorry, I made a stupid mistake again!:)

PR #560, opened 2015-09-07
fix nrec for lua_createtable, maybe __index ignored

ops-dev-cn:
adjust the "nrec" size for the lua_createtable.

### DIFF
--- a/t/062-count.t
+++ b/t/062-count.t
@@ -240,7 +240,7 @@ n = 10
 
 
 
-=== TEST 11: entries under ngx._reqsock_meta
+=== TEST 11: entries under the metatable of req sockets
 --- config
         location = /test {
             content_by_lua '
@@ -445,7 +445,7 @@ worker: 2
 
 
 
-=== TEST 20: entries under ngx._udp_meta
+=== TEST 20: entries under the metatable of udp sockets
 --- config
         location = /test {
             content_by_lua '
@@ -466,7 +466,7 @@ n = 6
 
 
 
-=== TEST 21: entries under ngx._rawreqsock_meta
+=== TEST 21: entries under the metatable of req raw sockets
 --- config
         location = /test {
             content_by_lua '
--- a/t/062-count.t
+++ b/t/062-count.t
@@ -241,19 +241,24 @@ n = 10
 
 
 === TEST 11: entries under ngx._reqsock_meta
---- SKIP
 --- config
         location = /test {
             content_by_lua '
                 local n = 0
-                for k, v in pairs(ngx._reqsock_meta) do
+                local sock, err = ngx.req.socket()
+                if not sock then
+                    ngx.say("failed to get the request socket: ", err)
+                end
+
+                for k, v in pairs(getmetatable(sock)) do
                     n = n + 1
                 end
                 ngx.say("n = ", n)
             ';
         }
 --- request
-GET /test
+POST /test
+hello world
 --- response_body
 n = 4
 --- no_error_log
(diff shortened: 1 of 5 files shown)


agentzh:
Will you please add some corresponding tests in the test file `t/062-count.t`? Please check out the existing tests for some examples. Thanks!

ops-dev-cn:
My pleasure. But i have to figure out  how tests are used.

agentzh:
@ops-dev-cn Let me know if you have any specific questions on the testing toolchain. Usually you just need a recent version of OpenResty to run the test suite (though some of the tests may fail due to the lack of special dependencies).

ops-dev-cn:
OK. I have got some questions need your advice. 
1.  Why ngx table is tested by set_by_lua, content_by_lua and so on  while ngx table is created at on place. is it necessary ?
2.  How can i decide which table should tested(count) on which phase ?

agentzh:
Hello!

On Fri, Sep 11, 2015 at 12:41 AM, ops-dev-cn wrote:
> 1. Why ngx table is tested by set_by_lua, content_by_lua and so on while ngx table is created at on place. is it necessary ?

Well, the tests just ensure that.

> 2. How can i decide which table should tested(count) on which phase ?
>

For your tests, pick just one phase is sufficient I think.

ops-dev-cn:
I'm not what to do with ngx req raw socket ·tcpsock, err = ngx.req.socket(raw)·. Maybe like this. But when i use curl to test,it just pending there. 
tcpsock, err = ngx.req.socket(raw)

```lua
        location = /test {
            content_by_lua '
                local n = 0
                local sock, err = ngx.req.socket(true)
                local output = "HTTP/1.1 200 OK\\r\\nConnection: keep-alive\\r\\nContent-Length: 3\\r\\n\\r\\n"
                sock:receive()
                for k, v in pairs(getmetatable(sock)) do
                    n = n + 1
                end
                sock:send(output.."n = "..n)
            ';
        }
```

agentzh:
Hello!

On Sun, Sep 13, 2015 at 1:00 AM, ops-dev-cn <notifications@github.com> wrote:
>
> I'm not what to do with ngx req raw socket ·tcpsock, err = ngx.req.socket(raw)·. Maybe like this.
>But when i use curl to test,it just pending there.
>

You can check out existing tests in the test suite for the raw req socket.

>         location = /test {
>             content_by_lua '
>                 local n = 0
>                 local sock, err = ngx.req.socket(true)

No error checking here? This call may fail.

>                 local output = "HTTP/1.1 200 OK\\r\\nConnection: keep-alive\\r\\nContent-Length: 3\\r\\n\\r\\n"

Because you're using content_by_lua to inline Lua here. You need to
double escape the back-slash char in Lua strings. See

    https://github.com/openresty/lua-nginx-module#special-escaping-sequences

(Oh, I'm working on content_by_lua_block {...} atm so that we no
longer need to worry about double-escaping when inlining Lua code
directly in nginx.conf.)

>                 sock:receive()

Are you sure you want to read a request body line here? This is
probably where the request hangs since there's really nothing to read.

Regards,
-agentzh

ops-dev-cn:
Tell me if there is anything i missed.

agentzh:
@ops-dev-cn Thank you for your hard work! I'll look into this shortly.

t/062-count.t
```diff
@@ -241,19 +241,24 @@ n = 10
 
 
 === TEST 11: entries under ngx._reqsock_meta
```
agentzh: Will fix the test title accordingly? We're no longer referencing `ngx._reqsock_meta` in the current test block any more. The same applies to the subsequent tests you edited. Thanks!

agentzh: Maybe we can just say "the metatable of req sockets" here?

agentzh:
@ops-dev-cn Will you please address the issues in my latest comments? Also, will you please rebase your branch upon the latest git master of the official repos? Other than those, this PR looks good to merge :) Thanks a lot!

ops-dev-cn:
OK.

PR #564, opened 2015-09-10
doc: typo fix at ngx.req.get_post_args

lancelijade:


agentzh:
@lancelijade Nice catch! Merged.

PR #565, opened 2015-09-11
doc:typo fix in ngx.req.get_headers

lancelijade:


agentzh:
@lancelijade Patch applied to master. Thanks!

PR #570, opened 2015-09-18
Update ngx_http_lua_subrequest.c

itpp16:
http2 condition.

### DIFF
--- a/src/ngx_http_lua_subrequest.c
+++ b/src/ngx_http_lua_subrequest.c
@@ -1489,6 +1489,10 @@ ngx_http_lua_subrequest(ngx_http_request_t *r,
     sr->spdy_stream = r->spdy_stream;
 #endif
 
+#if (NGX_HTTP_V2)
+    sr->stream = r->stream;
+#endif
+
 #ifdef HAVE_ALLOW_REQUEST_BODY_UPDATING_PATCH
     sr->content_length_n = -1;
 #endif


agentzh:
@itpp16 Will you combine this PR with #569? I don't see why they should be separated in the first place :) Thanks!

itpp16:
I wish I could but I haven't figured out how to combine patches or how to make a multi-file patchset, each patch I make to a file always ends up with one choice which is 'make a pull request' (github online) so if you can tell me how (or where github has hidden this option) I'll gladly combine them :)

agentzh:
@itpp16 You can just use `git cherry-pick` to apply the patch in branch A to branch B, and then only keeps branch B.

agentzh:
@itpp16 Oh, and then in branch B, you can use the `git rebase -i` command to squash the last two commits into a single one.

itpp16:
Yes if you use the 'offline' git commands, I can only submit code via the webinterface which does not have a multi-file commit mode/pull, each pull has maximum 1 file in it, I can't add other modifications in other files to the same pull unfortunately.

agentzh:
@itpp16 You're on Windows? Maybe you should try installing a Windows port of git there?

itpp16:
We import code from 5 different repositories including git, company policy does not allow any type of export via those tools, any code we want to share has to be submitted via webinterface (if available).

agentzh:
@itpp16 That's weird. You don't have Internet access at your own places? And you never do nontrivial code edits and commits on GitHub?

Anyway, I'll take care of #570 and #569 for you.

PR #572, opened 2015-09-27
missing return 

membphis:


agentzh:
@membphis Nice catch. Patch applied. Thanks!

PR #576, opened 2015-10-01
ngx.shared.DICT add two feature: incr add an option argument initial value, replace add an option argument check_flags

doujiang24:
1. incr add an option argument initial value
When the key is not exist in the dictionary and if init argument,
1) is not specified, it will return nil and "not found",
2) is specified, it will create a new key with the new value: value + init.

This is usefull when we used it as an counter.

2. replace add an option argument check_flags
Seems like check-and-set, replace will act only check_flags matched the current user_flags in the dictionary, and user_flags is used as an version.
This is usefull, cause we have a race-condition window between dict:get() and dict:set() across multiple nginx worker processes, and we can do like this example

I'm not sure if this api is good,  maybe a new api named cas_by_flags is better?

```lua
local value, flags = dict:get(key)
local success, err, forcible
while true do
    local newvalue = value + 1 -- something like this
    success, err, forcible, flags = dict:replace(key, newvalue, 0, new_flags, flags)
    if sucess then
        break
    elseif err == "not found" then
        -- turn to add
    elseif err == "flags not matched" then
         -- we just try again
    else
         -- something like no memory, just break
        break
    end
end

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -5626,17 +5626,22 @@ See also [ngx.shared.DICT](#ngxshareddict).
 
 ngx.shared.DICT.incr
 --------------------
-**syntax:** *newval, err = ngx.shared.DICT:incr(key, value)*
+**syntax:** *newval, err = ngx.shared.DICT:incr(key, value, init?)*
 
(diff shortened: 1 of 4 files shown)


agentzh:
Hmm, I prefer separate PRs for seperate new features in lua_shared_dict :)

doujiang24:
@agentzh sure, just created an new PR for incr, and sorry for the late:)

PR #579, opened 2015-10-09
feature: ngx.shared.dict add an option arg init

doujiang24:
incr add an option argument initial value When the key is not exist in the dictionary and if init argument, 1) is not specified, it will return nil and "not found", 2) is specified, it will create a new key with the new value: value + init.
This is usefull when we used it as an counter.


doujiang24:
And, now the incr with init just like `safe_add`, I'm not sure it's suitable enough
This is not good enough in lua-resty-traffic-limmit, maybe, we need an new api `safe_incr` or just force delete in `incr` with init?

doujiang24:
@agentzh I just fixed a stupid bug.
And now it will try to force remove unexpired nodes like DICT.add.
The option return value forcible won't be returned when init is not specified.

README.markdown
```diff
@@ -5844,17 +5844,24 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
 --------------------
-**syntax:** *newval, err = ngx.shared.DICT:incr(key, value)*
+**syntax:** *newval, err, forcible? = ngx.shared.DICT:incr(key, value, init?)*
 
 **context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
 
 Increments the (numerical) value for `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) by the step value `value`. Returns the new resulting number if the operation is successfully completed or `nil` and an error message otherwise.
 
-The key must already exist in the dictionary, otherwise it will return `nil` and `"not found"`.
+When the key is not exist in the dictionary and if `init` argument,
```
agentzh: grammar: "is not exist" => "does not exist"

agentzh: grammar: the trailing comma looks wrong to me.

agentzh: grammar; "if `init` argument" => "if the `init` argument".

doujiang24:
@agentzh Thanks for your review, fixed and rebased :)

src/ngx_http_lua_shdict.c
```diff
@@ -1238,11 +1242,38 @@ ngx_http_lua_shdict_incr(lua_State *L)
...
+        if (n == 3) {
+            ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+            lua_pushnil(L);
+            lua_pushliteral(L, "not found");
+            return 2;
+        }
+
+        /* add value */
+        num = value + luaL_checknumber(L, 4);
```
agentzh: The `luaL_checknumber` function may throw an exception, leaving the shm mutex locked.

src/ngx_http_lua_shdict.c
```diff
@@ -1238,11 +1242,38 @@ ngx_http_lua_shdict_incr(lua_State *L)
...
+
+        if (rc == NGX_DONE) {
+
+            if ((size_t) sd->value_len == sizeof(double)) {
+                ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ctx->log, 0,
+                            "lua shared dict incr: found old entry and value "
```
agentzh: Style: need to be aligned up.

src/ngx_http_lua_shdict.c
agentzh: We should check for the init argument earlier, before acquiring the lock.

doc/HttpLuaModule.wiki
```diff
@@ -5140,17 +5140,24 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
...
 
-The key must already exist in the dictionary, otherwise it will return <code>nil</code> and <code>"not found"</code>.
+When the key does not exist in the dictionary and if the <code>init</code> argument
+
+# is not specified, it will return <code>nil</code> and <code>"not found"</code>,
+# is specified, it will create a new <code>key</code> with the new value: <code>value + init</code>.
+
+Like the [[#ngx.shared.DICT.add|add]] method, it also overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone.
+
+The <code>forcible</code> return value will not be returned when <code>init</code> argument is not specified.
```
agentzh: I think the `forcible` return value needs some documentation about its semantics.

agentzh:
@doujiang24 Please check out my most recent comments :)

BTW, will you please update `ngx_http_lua_ffi_shdict_incr` too. The FFI implementation needs to be in sync. Thanks!

doujiang24:
@agentzh Fixed & rebased :)

src/ngx_http_lua_shdict.c
```diff
@@ -1238,11 +1247,38 @@ ngx_http_lua_shdict_incr(lua_State *L)
...
+        }
+
+        /* add value */
+        num = value + init;
+
+        if (rc == NGX_DONE) {
+
+            if ((size_t) sd->value_len == sizeof(double)) {
+                ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ctx->log, 0,
+                               "lua shared dict incr: found old entry and value "
```
agentzh: Style: exceeded 80 columns (found by `ngx-releng`).

src/ngx_http_lua_shdict.c
```diff
@@ -1823,9 +1957,35 @@ ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
...
+        }
+
+        /* add value */
+        num = *value + init;
+
+        if (rc == NGX_DONE) {
+
+            if ((size_t) sd->value_len == sizeof(double)) {
+                ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ctx->log, 0,
+                               "lua shared dict incr: found old entry and value "
```
agentzh: Style: exceeded 80 columns.

src/ngx_http_lua_shdict.c
```diff
@@ -1823,9 +1957,35 @@ ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
...
+            *err = "not found";
+            return NGX_ERROR;
+        }
+
+        /* add value */
+        num = *value + init;
+
+        if (rc == NGX_DONE) {
+
+            if ((size_t) sd->value_len == sizeof(double)) {
```
agentzh: I think we should check the value type instead of simply checking value length. This is because different value types MAY have identical value lengths.

src/ngx_http_lua_shdict.c
```diff
@@ -1823,9 +1957,35 @@ ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
...
+
+                ngx_queue_remove(&sd->queue);
+                ngx_queue_insert_head(&ctx->sh->queue, &sd->queue);
+
+                dd("go to setvalue");
+                goto setvalue;
+            }
+
+            dd("go to remove");
+            goto remove;
```
agentzh: I think if the key already takes a non-numerical value, then we should really return an error instead of initializing with `init`. I think `init` should only be used when the key does not exist.

agentzh:
BTW, please rebase to the latest git master branch. Thank you!

src/ngx_http_lua_shdict.c
doujiang24: Yes, the way your suggested is more safer :)
But I think may be we should keep the same semantics as `shdict.set` when the key is expired or not exists?
And sometime we need to set the key even key is expired.

agentzh: @doujiang24 Yes, we use the user `init` argument value when the key is nonexistent or expired.

src/ngx_http_lua_shdict.c
doujiang24: I think it's safe to replace different value type when key is expired according to the last comment :)

agentzh:
Already merged into master.

PR #583, opened 2015-10-16
feature: add a new api ngx.shared.DICT.cas

doujiang24:
Just like the replace method, but only stores the key-value pair into the dictionary ngx.shared.DICT if the old_value argument and old_flags argument does match the value and flags in the dictionary ngx.shared.DICT (or one of them).


dualface:
+1

doc/HttpLuaModule.wiki
```diff
@@ -4698,6 +4699,56 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
 
 See also [[#ngx.shared.DICT|ngx.shared.DICT]].
 
+== ngx.shared.DICT.cas ==
+'''syntax:''' ''success, err, forcible, current_value?, current_flags? = ngx.shared.DICT:cas(key, value, exptime, flags, old_value, old_flags?)''
+
+'''context:''' ''init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.*''
```
agentzh: Need to add the newly added contexts recently, like `balancer_by_lua*` and `ssl_certificate_by_lua*`.

doc/HttpLuaModule.wiki
```diff
@@ -4698,6 +4699,56 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
...
+
+Just like the [[#ngx.shared.DICT.replace|replace]] method, but only stores the key-value pair into the dictionary [[#ngx.shared.DICT|ngx.shared.DICT]] if the <code>old_value</code> argument and <code>old_flags</code> argument ''does'' match the value and flags in the dictionary [[#ngx.shared.DICT|ngx.shared.DICT]] (or one of them).
+
+The <code>old_value</code> argument can be <code>nil</code> only when <code>old_flags</code> argument is specified, only <code>flags</code> will be checked.
```
agentzh: This sentence is confusing. I think you should add "in which case" before the last sub-sentence.

doc/HttpLuaModule.wiki
```diff
@@ -4698,6 +4699,56 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
 
 See also [[#ngx.shared.DICT|ngx.shared.DICT]].
 
+== ngx.shared.DICT.cas ==
+'''syntax:''' ''success, err, forcible, current_value?, current_flags? = ngx.shared.DICT:cas(key, value, exptime, flags, old_value, old_flags?)''
+
+'''context:''' ''init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.*''
+
+Just like the [[#ngx.shared.DICT.replace|replace]] method, but only stores the key-value pair into the dictionary [[#ngx.shared.DICT|ngx.shared.DICT]] if the <code>old_value</code> argument and <code>old_flags</code> argument ''does'' match the value and flags in the dictionary [[#ngx.shared.DICT|ngx.shared.DICT]] (or one of them).
```
agentzh: Grammar: "do match" instead of "does match" since you have two things in the subject.

agentzh: I think it's better to replace "if" with "if and only if".

The "or one of them" is confusing. In fact, when both arguments are specified (non-nil), both of the values must match instead of just one of them matches. I think you really mean that the arguments are optional and can take `nil` values, in which case the checks are omitted. Let's be explicit rather than confusing :)

doc/HttpLuaModule.wiki
```diff
@@ -4698,6 +4699,56 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
...
+
+The <code>old_value</code> argument can be <code>nil</code> only when <code>old_flags</code> argument is specified, only <code>flags</code> will be checked.
+
+If <code>old_flags</code> argument is not specified, only <code>value</code> will be checked.
+
+The optional <code>old_flags</code> argument can be <code>nil</code>, and it means <code>0</code>.
+
+If they do ''not'' match, the <code>success</code> return value will be <code>false</code> and the <code>err</code> return value will be <code>"not matched"</code>. The <code>current_value</code> return value and <code>current_flags</code> return value will be the current <code>value</code> and current <code>flags</code> in the dictionary [[#ngx.shared.DICT|ngx.shared.DICT]], just like [[#ngx.shared.DICT.get|get]] does.
+
```
agentzh: I think we need explanation on the use cases here instead of just a code example.

doc/HttpLuaModule.wiki
```diff
@@ -4698,6 +4699,56 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
...
+    local cats = ngx.shared.cats
+    cats:set("foo", 1, 1)
+
+    local old_value, old_flags = cats:get("foo")
+
+    while true do
+        local newvalue = calculate(old_value) -- some logic
+        local newflags = (old_flags or 0) + 1
+
+        local success, err, forcibly, current_value, current_flags = cats:cas("foo", newvalue, 0, newflags, old_value, old_flags)
```
agentzh: This line is too long. Better split it.

agentzh:
Please rebase to the latest git master. Thank you :)

doujiang24:
@agentzh Thanks, I have rebased to the master and fixed the confusing descriptions.

splitice:
+1

I'll be merging this into my fork (with incr ttl). It would be good if there was a FFI (resty core) PR for this too

agentzh:
@dolik-rce Just the lack of time and higher priority items. I think @thibaultcha will take care of this :)

PR #584, opened 2015-10-21
a simple implementation of semaphore in ngx_lua from kugou inc

cuiweixie:
implement basical sempahore operation wait and post,using luajit ffi,and c.


doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5592,71 @@ According to the current implementation, only the parent coroutine (or "light th
...
+<geshi lang="lua">
+    require "resty.core.semaphore"
+    local print = ngx.print
+    if not ngx.semaphore.test then
+        local sem,err = ngx.semaphore.new(0)
+        if not sem then
+            print("create semaphore failed: "..err)
+            ngx.exit(500)
+        end
+        ngx.semaphore.test = sem
```
agentzh: Hmm, writing to the `ngx.semaphore` table from the user side is a bad practice.

src/ngx_http_lua_common.h
```diff
@@ -153,6 +153,8 @@ struct ngx_http_lua_main_conf_s {
 
     ngx_uint_t                      shm_zones_inited;
 
+    ngx_int_t                  semaphore_threshold;
```
agentzh: Style: please align up the identifier with adjacent lines.

doc/HttpLuaModule.wiki
```diff
@@ -2011,6 +2011,17 @@ When exceeding this limit, Nginx will stop running the callbacks of newly expire
...
 
+== lua_semaphore_threshold ==
+
+'''syntax:''' ''lua_semaphore_threshold <threshold>''
+
+'''default:''' ''lua_semaphore_threshold 100000''
+
+'''context:''' ''http''
+
+The threshold of lua semaphore.
```
agentzh: I'm afraid the "threshold" word is quite confusing. Do you mean the maximum count of thresholds allowed in any individual nginx worker? Or something else? Please be explicit.

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5592,71 @@ According to the current implementation, only the parent coroutine (or "light th
...
+</geshi>
+
+== ngx.semaphore.wait ==
+'''syntax:''' ''ok, err = ngx.semaphore.wait(sem,timeout?)''
+
+'''context:''' ''rewrite_by_lua*, access_by_lua*, content_by_lua*, ngx.timer.*''
+The param sem is create by [[#ngx.semaphore.new|ngx.semaphore.new]].wait on one semapohre.if there a resouce then it return immediately,else the light thread or main thread or coroutine will sleep,then it will be wake up when some one else call the post method[#ngx.semaphore.post|ngx.semaphore.post]] or timeout event occur.timeout default is 0,which means it will return nil,"busy" if there is no resource to use.
+
+<geshi lang="lua">
+    require "resty.core.semaphore"
```
agentzh: I guess using a Lua module here in this example is a better choice.

src/ngx_http_lua_module.c
```diff
@@ -384,6 +384,13 @@ static ngx_command_t ngx_http_lua_cmds[] = {
       offsetof(ngx_http_lua_loc_conf_t, use_default_type),
       NULL },
 
+      { ngx_string("lua_semaphore_threshold"),
+      NGX_HTTP_MAIN_CONF|NGX_CONF_TAKE1,
```
agentzh: Style: please be careful with indentation. See other entries for examples. Thank you :)

src/ngx_http_lua_module.c
```diff
@@ -545,6 +553,16 @@ ngx_http_lua_init(ngx_conf_t *cf)
         }
     }
 
+    /*add the cleanup of semaphores after the lua_close */
```
agentzh: Style: need a space after `/*`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+        sem->in_post_event = 1;
+        ngx_post_event((&sem->sem_event), &ngx_posted_events);
+    }
+
+    return NGX_OK;
+}
+
+
+int
```
agentzh: We use `ngx_int_t` for values like `NGX_ERROR` and `NGX_OK`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+ngx_http_lua_ffi_sem_wait(ngx_http_request_t *r, ngx_http_lua_semaphore_t *sem,
+                          int time,u_char *errstr, size_t *errlen)
+{
+    ngx_http_lua_ctx_t           *ctx;
+    ngx_http_lua_co_ctx_t        *wait_co_ctx;
+    ngx_int_t                    rc;
```
agentzh: Style: `rc` should be aligned up with `wait_co_ctx` in the previous line (instead of with `*wait_co_ctx`).

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+    if (r == NULL) {
+
+        *errlen = ngx_snprintf(errstr, *errlen,
+                               "request is null") - errstr;
+         return NGX_ERROR;
```
agentzh: Indentation issue above?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
+
```
agentzh: We need to support the `NGX_LUA_NO_FFI_API` macro here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+int
+ngx_http_lua_ffi_get_semaphore_threshold()
+{
+    return (int)ngx_http_lua_semaphore_threshold;
+}
+
+
+static ngx_http_lua_semaphore_t*
```
agentzh: Style: need a space before `*`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+int ngx_http_lua_ffi_sem_wait(ngx_http_request_t *r, ngx_http_lua_semaphore_t *sem,
+                              int time, u_char *errstr, size_t *errlen);
+static void ngx_http_lua_ffi_semaphore_cleanup(void *data);
+void ngx_http_lua_ffi_semaphore_handler(ngx_event_t *ev);
+static void ngx_http_lua_sem_timeout_handler(ngx_event_t *ev);
+void ngx_http_lua_ffi_sem_gc(ngx_http_lua_semaphore_t *sem);
+
+static ngx_http_lua_semaphore_t *ngx_http_lua_sem_free_list = NULL;
+static ngx_int_t ngx_http_lua_semaphore_threshold = 0;
+static ngx_int_t ngx_http_lua_sem_free_count = 0;
```
agentzh: Can we avoid these global C variables above? They *may* cause problems during HUP reload. Better hook them up in the `ngx_http_lua_main_conf_t` construct.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    ngx_http_lua_semaphore_t *sem;
+
+    if (ngx_http_lua_sem_free_list) {
+         sem = ngx_http_lua_sem_free_list;
+         ngx_http_lua_sem_free_list = ngx_http_lua_sem_free_list->next;
+         ngx_http_lua_sem_free_count--;
+
+    } else {
+
+        sem = ngx_alloc(sizeof(ngx_http_lua_semaphore_t), log);
```
agentzh: Instead of creating an individual memory block here, maybe we could preallocate more blocks here to reduce the "warmup" cost?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+
+void
+ngx_http_lua_cleanup_sem(void *data)
+{
+    dd("ngx_http_lua_cleanup_sem");
+
+    ngx_http_lua_semaphore_t * sem;
+
+    while ((sem=ngx_http_lua_sem_free_list) != NULL)
```
agentzh: Style: we avoid assignments in loop condition expressions in the NGINX world. Besides, you miss spaces around the `=` operator.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    sem->in_post_event = 0;
+    sem->log = ngx_cycle->log;
+    *psem = sem;
+
+    dd("ngx_http_lua_ffi_sem_new semaphore %p value %d", sem, sem->count);
+
+    return NGX_OK;
+}
+
+#if 0
```
agentzh: Let's just remove this dead code block?

src/ngx_http_lua_semaphore.c
agentzh: Okay, never mind, my bad. This is part of the ABI, so use of `int` here is appropriate.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+    if (rc == NGX_DECLINED) {
+        return NGX_ERROR;
+    }
+
+    if (!sem->in_post_event && sem->count > 0) {
+         sem->count--;
+         return NGX_OK;
+
+    } else if (time == 0) {
```
agentzh: `else` here is not needed since we return in the `if` branch.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+        return NGX_DECLINED;
+
+    } else {
```
agentzh: Ditto.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+         } else {
+             ctx->resume_handler = ngx_http_lua_semaphore_resume;
+             ngx_http_core_run_phases(r);
+         }
+
+         ngx_http_run_posted_requests(c);
+
+    }
+
+    /*after dealing semaphore post event */
```
agentzh: Style: need a space after `/*`.

Grammar: `dealing with` is the correct form.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+         if (ctx->entered_content_phase) {
+             (void) ngx_http_lua_semaphore_resume(r);
+
+         } else {
+             ctx->resume_handler = ngx_http_lua_semaphore_resume;
+             ngx_http_core_run_phases(r);
+         }
+
+         ngx_http_run_posted_requests(c);
+
```
agentzh: Style: a redundant blank link.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    ngx_http_lua_semaphore_t    *sem;
+
+    wait_co_ctx = ev->data;
+
+    r = ngx_http_lua_get_req(wait_co_ctx->co);
+
+    dd("ngx_http_lua_sem_timeout_handler timeout coctx:%p", wait_co_ctx);
+
+    sem = wait_co_ctx->data;
+    sem->wait_count --;
```
agentzh: Style: a space is not allowed before postfix operators.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+
+#ifndef _NGX_HTTP_LUA_SEMAPHORE_H_INCLUDED_
+#define _NGX_HTTP_LUA_SEMAPHORE_H_INCLUDED_
+
+
+#include "ngx_http_lua_common.h"
+
+#define ngx_http_lua_sem_dis_api_err(c) \
```
agentzh: Do we have to re-invent this? I remember we already have something similar.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+
+typedef struct ngx_http_lua_semaphore_s {
+    ngx_queue_t wait_queue;
+    ngx_event_t sem_event;
+    struct ngx_http_lua_semaphore_s *next;
+    ngx_log_t *log;
+    int count;
+    unsigned in_post_event:1;
+    unsigned wait_count:31;
+ }ngx_http_lua_semaphore_t;
```
agentzh: Style: a space is required after `}`.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+     "api is disable in header_filter_by_lua*"   \
+    : (c) == NGX_HTTP_LUA_CONTEXT_BODY_FILTER ? \
+    "api is disable in body_filter_by_lua*"    \
+    : (c) == NGX_HTTP_LUA_CONTEXT_TIMER ? \
+    "api is disable in ngx.timer"                       \
+    : (c) == NGX_HTTP_LUA_CONTEXT_INIT_WORKER ? \
+    "api is disable in init_worker_by_lua*"       \
+    : "(unknown)")
+
+typedef struct ngx_http_lua_semaphore_s {
```
agentzh: The struct fields' identifiers need to be aligned.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+    ngx_queue_t wait_queue;
+    ngx_event_t sem_event;
+    struct ngx_http_lua_semaphore_s *next;
+    ngx_log_t *log;
+    int count;
+    unsigned in_post_event:1;
+    unsigned wait_count:31;
+ }ngx_http_lua_semaphore_t;
+
```
agentzh: Style: 2 blank lines required between such definitions.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+#define NGX_HTTP_LUA_SEM_WAIT_TIMEOUT 1
+#define NGX_HTTP_LUA_SEM_WAIT_SUCC    0
+int ngx_http_lua_ffi_set_semaphore_threshold(ngx_int_t threshold);
+void ngx_http_lua_cleanup_sem(void *data);
+#endif /* _NGX_HTTP_LUA_SEMAPHORE_H_INCLUDED_ */
```
agentzh: Style: 2 blank lines needed before this `#endif` line.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+    struct ngx_http_lua_semaphore_s *next;
+    ngx_log_t *log;
+    int count;
+    unsigned in_post_event:1;
+    unsigned wait_count:31;
+ }ngx_http_lua_semaphore_t;
+
+#define NGX_HTTP_LUA_SEM_WAIT_TIMEOUT 1
+#define NGX_HTTP_LUA_SEM_WAIT_SUCC    0
+int ngx_http_lua_ffi_set_semaphore_threshold(ngx_int_t threshold);
```
agentzh: Ditto.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    sem = ev->data;
+
+    while (!ngx_queue_empty(&sem->wait_queue) && sem->count > 0) {
+
+         q = ngx_queue_head(&sem->wait_queue);
+         ngx_queue_remove(q);
+         wait_co_ctx = ngx_queue_data(q, ngx_http_lua_co_ctx_t, sem_wait_queue);
+
+         sem->count--;
+         sem->wait_count --;
```
agentzh: Style: spaces are not allowed before postfix operators.

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,49 @@
...
+    ngx_event_t sem_event;
+    struct ngx_http_lua_semaphore_s *next;
+    ngx_log_t *log;
+    int count;
+    unsigned in_post_event:1;
+    unsigned wait_count:31;
+ }ngx_http_lua_semaphore_t;
+
+#define NGX_HTTP_LUA_SEM_WAIT_TIMEOUT 1
+#define NGX_HTTP_LUA_SEM_WAIT_SUCC    0
```
agentzh: Do these constants really need to be exported outside the semaphore compilation unit?

src/ngx_http_lua_semaphore.c
cuiweixie: do you mean creating a list of  individual memory block when workers start,or just malloc a page one time,and then split it into small  blocks(sizeof(ngx_http_lua_semaphore_t)).it think malloc a page one time is not the best way,becasue glibc already do something like that to some extend,and  malloc a page one time means we have to maintain the page set by using extra mem.

src/ngx_http_lua_semaphore.c
cuiweixie: it seems that the pointer ngx_cycle which we need to get ngx_http_lua_main_conf_t from is point to the ngx_exit_cycle after the luavm is destoryed.I just don't know how to get lmcf when ngx_http_lua_free_sem and ngx_http_lua_cleanup_sem is call.ngx_http_cycle_get_module_main_conf(ngx_cycle,ngx_http_lua_module) doesn't work for this situation.

agentzh: @cuiweixie It seems that you're not making use of the `data` parameter of the cleanup handler; that's where you think should hook up `lmcf` or alike to.

src/ngx_http_lua_semaphore.c
agentzh: @cuiweixie I'm not suggesting creating page-size blocks in particular. Instead, I'm suggesting doing some heuristics to reduce `malloc` and `free` calls.

BTW, I don't think we should assume the underlying libc is always glibc (it could be something much worse) and even for glibc's allocator implementation it cannot defeat careful userland memory management at any rate since it lacks higher level information to be really smart. Creating so many small memory blocks in the system allocator has inevitable bookkeeping overhead. Since we are doing fixed-size (small) blocks here exclusively, my hunch is that it could be easier to do better with our own policies.

I suggest we should think more about it.

src/ngx_http_lua_module.c
```diff
@@ -545,6 +553,16 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
+    cln = ngx_pool_cleanup_add(cf->pool, 0);
+    if (cln == NULL) {
+        return NGX_ERROR;
+    }
+    cln->data = NULL;
```
agentzh: It's a shame that we do not utilize this `data` pointer but rely on global C variables.

src/ngx_http_lua_module.c
```diff
@@ -638,6 +656,7 @@ ngx_http_lua_create_main_conf(ngx_conf_t *cf)
     lmcf->pool = cf->pool;
     lmcf->max_pending_timers = NGX_CONF_UNSET;
     lmcf->max_running_timers = NGX_CONF_UNSET;
+    lmcf->semaphore_threshold = NGX_CONF_UNSET;
```
agentzh: A blank line needed after this line.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+
+#include "ngx_http_lua_util.h"
+#include "ngx_http_lua_semaphore.h"
+#include "ngx_http_lua_contentby.h"
+
```
agentzh: 2 blank lines needed here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+int ngx_http_lua_ffi_set_semaphore_threshold(ngx_int_t threshold);
+static ngx_http_lua_semaphore_t *ngx_http_lua_alloc_sem(ngx_log_t *log);
+void ngx_http_lua_cleanup_sem(void *data);
+static void ngx_http_lua_free_sem(ngx_http_lua_semaphore_t *sem, ngx_log_t *log);
+static ngx_int_t ngx_http_lua_semaphore_resume(ngx_http_request_t *r);
+int ngx_http_lua_ffi_sem_new(ngx_http_lua_semaphore_t **psem, int n, char **errstr);
+int ngx_http_lua_ffi_sem_post(ngx_http_lua_semaphore_t *sem, int n, char **errstr);
+int ngx_http_lua_ffi_sem_wait(ngx_http_request_t *r, ngx_http_lua_semaphore_t *sem,
+                              int time, u_char *errstr, size_t *errlen);
```
agentzh: Style: for C function declarations (instead of calls), the indentation for the continued parameter list should be 4 spaces.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+static void ngx_http_lua_ffi_semaphore_cleanup(void *data);
+void ngx_http_lua_ffi_semaphore_handler(ngx_event_t *ev);
+static void ngx_http_lua_sem_timeout_handler(ngx_event_t *ev);
+void ngx_http_lua_ffi_sem_gc(ngx_http_lua_semaphore_t *sem);
+
```
agentzh: 2 blank lines needed here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+static ngx_http_lua_semaphore_t *ngx_http_lua_sem_free_list = NULL;
+static ngx_int_t ngx_http_lua_semaphore_threshold = 0;
+static ngx_int_t ngx_http_lua_sem_free_count = 0;
+
+
+int
+ngx_http_lua_ffi_set_semaphore_threshold(ngx_int_t threshold)
+{
+    ngx_http_lua_semaphore_threshold = threshold;
+    return (int)ngx_http_lua_semaphore_threshold;
```
agentzh: Style: a space required after `(int)`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+int
+ngx_http_lua_ffi_get_semaphore_threshold()
+{
+    return (int)ngx_http_lua_semaphore_threshold;
```
agentzh: Ditto.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    return sem;
+}
+
+
+void
+ngx_http_lua_cleanup_sem(void *data)
+{
+    dd("ngx_http_lua_cleanup_sem");
+
+    ngx_http_lua_semaphore_t * sem;
```
agentzh: Style: no space is allowed after `*` in the local variable definitions.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+         ngx_http_lua_sem_free_count -= n;
+
+         while (n--) {
+             sem = ngx_http_lua_sem_free_list;
+             ngx_http_lua_sem_free_list = ngx_http_lua_sem_free_list->next;
+             ngx_free(sem);
+         }
+
+         dd("ngx_http_lua_free_sem sem_count:%lld",
+                 (long long int) ngx_http_lua_sem_free_count);
```
agentzh: Style: continued argument list should be aligned right after the `(` char in the previous line.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    }
+}
+
+
+static ngx_int_t
+ngx_http_lua_semaphore_resume(ngx_http_request_t *r)
+{
+    lua_State                   *vm;
+    ngx_connection_t            *c;
+    ngx_int_t                   rc;
```
agentzh: Style: identifiers should be aligned up *excluding* the `*` part.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+        return ngx_http_lua_run_posted_threads(c, vm, r, ctx);
+    }
+
+    /* rc == NGX_ERROR || rc >= NGX_OK */
+    if (ctx->entered_content_phase) {
+        ngx_http_lua_finalize_request(r, rc);
+        return NGX_DONE;
+    }
+
+     return rc;
```
agentzh: Style: 5 space indentation here?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    if (!ngx_queue_empty(&sem->wait_queue)) {
+        ngx_log_debug1(NGX_LOG_ERR, ngx_cycle->log, 0,
+                       "ngx_http_lua_ffi_sem_gc wait queue is"
+                       " not empty while the semaphore is "
+                       "going to be destroyed", sem);
+    }
+
+    ngx_http_lua_free_sem(sem,sem->log);
```
agentzh: Style: a space is required after `,`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+
+    ngx_http_run_posted_requests(c);
+}
+
+
+void
+ngx_http_lua_ffi_sem_gc(ngx_http_lua_semaphore_t *sem)
+{
+    ngx_log_debug1(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+     "ngx_http_lua_ffi_sem_gc %p", sem);
```
agentzh: Style: continued argument list should be aligned up with the position right after the `(` char in the previous line.

src/ngx_http_lua_semaphore.h
agentzh: Okay, I've checked the code base. Yes, we already have the macro `ngx_http_lua_check_context` and I suggest you make use of it instead of reinventing your own :)

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+         wait_co_ctx->data = sem;
+         wait_co_ctx->cleanup = ngx_http_lua_ffi_semaphore_cleanup;
+
+         return NGX_AGAIN;
+    }
+}
+
+
+static void
+ngx_http_lua_ffi_semaphore_cleanup(void *data)
```
agentzh: Hmm, some comments:

1. I suggest using the `ngx_http_lua_ffi_` prefix for public C functions that are part of the ABI.
2. I see you are using both "sem" and "semaphore" in your C function names, which is inconsistent. Please be consistent :)

Thanks!

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    ngx_http_lua_co_ctx_t          *coctx = data;
+    ngx_queue_t                    *q;
+    ngx_http_lua_semaphore_t       *sem;
+
+    sem = coctx->data;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, sem->log, 0,
+                   "ngx_http_lua_semaphore_cleanup");
+
+    ngx_del_timer(&coctx->sleep);
```
agentzh: Better check `coctx->sleep.timer_set` before deleting it.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+    ngx_http_lua_semaphore_t          *sem;
+    ngx_http_request_t                *r;
+    ngx_http_lua_ctx_t                *ctx;
+    ngx_http_lua_co_ctx_t             *wait_co_ctx;
+    ngx_connection_t                  *c;
+    ngx_queue_t                       *q;
+
+    sem = ev->data;
+
+    while (!ngx_queue_empty(&sem->wait_queue) && sem->count > 0) {
```
agentzh: Do we really need both of these conditions? Is it possible for the `count` field to get out of sync here?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,433 @@
...
+{
+    ngx_log_debug1(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+     "ngx_http_lua_ffi_sem_gc %p", sem);
+
+    if (sem == NULL){
+        return;
+    }
+
+    if (!ngx_queue_empty(&sem->wait_queue)) {
+        ngx_log_debug1(NGX_LOG_ERR, ngx_cycle->log, 0,
```
agentzh: I guess you really want to use `ngx_log_error` instead of `ngx_log_debug1` here. Also, such errors should be of the `crit` logging level I suppose.

README.markdown
```diff
@@ -6592,6 +6610,83 @@ This API was first enabled in the `v0.9.9` release.
...
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.semaphore.post
+------------------
+**syntax:** *ok, err = ngx.semaphore.post(sem)*
+
+**context:** *init_by_lua*, init_worker_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua, log_by_lua*, ngx.timer.**
+
+The param sem is create by [ngx.semaphore.new](#ngxsemaphorenew).release one resource to a semaphore.if one light wethread or coroutine is waiting on this semaphore,then it will be wake up.
```
agentzh: I think you need to capitalize the first word of an English sentence. Also, a space is usually needed after a punctuation mark, like `.` and `,`.

The docs contain typos like `wethread` and grammar errors like `be wake up`. Please fix them too. Thanks!

README.markdown
```diff
@@ -6592,6 +6610,83 @@ This API was first enabled in the `v0.9.9` release.
...
+
+```lua
+
+ require "resty.core.semaphore"
```
agentzh: I think the correct way of loading a Lua module is like this:

```lua
local semaphore = require "ngx.semaphore"
```

And then, we do

```
local sem = semaphore.new(0)
```

to instantiate a new object without using any global variables (but the `require` itself).

README.markdown
```diff
@@ -6592,6 +6610,83 @@ This API was first enabled in the `v0.9.9` release.
...
+ if not ok then
+     ngx.print(err)
+ end
+```
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.semaphore.wait
+------------------
+**syntax:** *ok, err = ngx.semaphore.wait(sem,timeout?)*
```
agentzh: Style: a space is required after `,` (oh, please also fix all other occurrences of this issue since there're too many of them).

agentzh: Also, maybe we should encourage the more compact form? That is, `sem:wait(timeout)`.

src/ngx_http_lua_module.c
```diff
@@ -23,7 +23,7 @@
 #include "ngx_http_lua_initby.h"
 #include "ngx_http_lua_initworkerby.h"
 #include "ngx_http_lua_probe.h"
-
+#include "ngx_http_lua_semaphore.h"
```
agentzh: Style: 2 blank lines needed here. Yes, we use blank lines and spaces very carefully in the NGINX coding style.

agentzh:
@cuiweixie I appreciate your hard work! Please address my comments above. I suggest the first pass be focused on coding style issues since that would make my eyes much easier when reviewing this patch :)

Thanks!

src/ngx_http_lua_semaphore.c
cuiweixie: good idea.but when luavm is destoryed,in the gc the lmcf is need to,and no param here.

src/ngx_http_lua_semaphore.h
cuiweixie: you area right

src/ngx_http_lua_semaphore.c
cuiweixie: yes,if delete the first,we don't know whether if some one is waitting for it,if delete the second we donnot know whether we can offer the resource.

agentzh: @cuiweixie Fair enough. I confused the "count" field's meaning. Maybe it deserves a better name? Like `resource_count`?

cuiweixie: resource count is better.

src/ngx_http_lua_semaphore.c
cuiweixie: style fixed

src/ngx_http_lua_semaphore.c
cuiweixie: support

cuiweixie: support

cuiweixie:
Hello, agentzh I have fix the style that memtioned in your comment one by one. Please review the code  
at your convenience.

agentzh:
@cuiweixie Awesome! I'll do another pass soon :)

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
 
 This API was first enabled in the <code>v0.9.9</code> release.
 
+== ngx.semaphore.new ==
```
agentzh: I think we'd better put the documentation of the `ngx.semaphore` module to the lua-resty-core library instead since the Lua API is defined there. In addition, it deserves a separate documentation file as `lib/ngx/semaphore.md`, just like the module documentation files in the `lua-resty-limit-traffic` library.

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
...
+        print("create semaphore failed: "..err)
+    end
+</geshi>
+
+== ngx.semaphore.post ==
+'''syntax:''' ''ok, err = sem:post()''
+
+'''context:''' ''init_worker_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua, log_by_lua*, ngx.timer.*''
+
+The param sem is create by [[#ngx.semaphore.new|ngx.semaphore.new]]. Release one resource to a semaphore. If one light thread or coroutine is waiting on this semaphore, then it will be waked up. LuaJIT's ffi is need by this api.
```
agentzh: Grammar: "is create by" -> "is created by".

doc/HttpLuaModule.wiki
agentzh: We just need a simple reference here for the `ngx.semaphore` module in `lua-resty-core`.

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
...
+
+<geshi lang="lua">
+    local semaphore = require "ngx.semaphore"
+    local shared =  getmetatable(_G).__index
```
agentzh: Oh, man, this is bloodily hacky. We are against use of Lua globals especially such "true globals". Please see https://github.com/openresty/lua-nginx-module#lua-variable-scope

Instead, we should use proper Lua modules for worker-level data sharing in our own documenmtation, as documented here: https://github.com/openresty/lua-nginx-module#data-sharing-within-an-nginx-worker

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
 
 This API was first enabled in the <code>v0.9.9</code> release.
 
+== ngx.semaphore.new ==
+'''syntax:''' ''sem, err = ngx.semaphore.new(n)''
+
+'''context:''' ''init_worker_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua, log_by_lua*, ngx.timer.*''
+
+Create a semaphore that has n resource. LuaJIT's ffi is need by this api.
```
agentzh: Grammar: "Create" -> "Creates". We need `<code>` around `n`. Also, FFI is preferred over ffi.

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
...
+    if not ok then
+        ngx.print(err)
+    end
+</geshi>
+
+== ngx.semaphore.wait ==
+'''syntax:''' ''ok, err = sem:wait(timeout?)''
+
+'''context:''' ''rewrite_by_lua*, access_by_lua*, content_by_lua*, ngx.timer.*''
+The param sem is create by [[#ngx.semaphore.new|ngx.semaphore.new]]. Wait on one semapohre. If there a resouce then it return immediately, else the light thread or main thread or coroutine will sleep, then it will be waked up when some one else call the post method[#ngx.semaphore.post|ngx.semaphore.post]] or timeout event occur. The timeout default is 0,which means it will return nil,"busy" if there is no resource to use. LuaJIT's ffi is need by this api.
```
agentzh: English requires the past particle form of verbs in passive voice. Please fix other places of mistakes like "is create by", "is need by", and etc.

agentzh: Grammar: "If there a resource" -> "If there is a resource", "it return" -> "it returns". Be careful :)

doc/HttpLuaModule.wiki
```diff
@@ -5581,6 +5581,73 @@ According to the current implementation, only the parent coroutine (or "light th
...
+
+<geshi lang="lua">
+    local semaphore = require "ngx.semaphore"
+    local shared =  getmetatable(_G).__index
+    local print = ngx.print
+    if not shared.test then
+        local sem,err = semaphore.new(0)
```
agentzh: Style: please ensure there is a space after a comma (`,`). I see many places having this problem.

config
```diff
@@ -391,3 +393,4 @@ ngx_feature_test='setsockopt(1, SOL_SOCKET, SO_PASSCRED, NULL, 0);'
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_PATH='\"/usr/local/openresty/lualib/?.lua\"'"
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_CPATH='\"/usr/local/openresty/lualib/?.so\"'"
 
+
```
agentzh: Style: a redundant newline here?

doc/HttpLuaModule.wiki
```diff
@@ -1,4 +1,4 @@
-= Name =
+﻿= Name =
```
agentzh: You have U+FEFF at the beginning of this line. Please remove it.

src/ngx_http_lua_module.c
```diff
@@ -638,12 +650,28 @@ ngx_http_lua_create_main_conf(ngx_conf_t *cf)
...
     lmcf->max_pending_timers = NGX_CONF_UNSET;
     lmcf->max_running_timers = NGX_CONF_UNSET;
+
 #if (NGX_PCRE)
     lmcf->regex_cache_max_entries = NGX_CONF_UNSET;
     lmcf->regex_match_limit = NGX_CONF_UNSET;
 #endif
     lmcf->postponed_to_rewrite_phase_end = NGX_CONF_UNSET;
 
+
```
agentzh: Style: 2 successive blank lines?

src/ngx_http_lua_module.c
```diff
@@ -673,6 +701,10 @@ ngx_http_lua_init_main_conf(ngx_conf_t *cf, void *conf)
         lmcf->max_running_timers = 256;
     }
 
+    if (lmcf->semaphore_mm->num_per_block == NGX_CONF_UNSET_UINT) {
+        lmcf->semaphore_mm->num_per_block = 4094;
```
agentzh: Maybe add a comment explaining the magical number 4094 a bit? It's not a power of 2 and deserves a little explanation I suppose :)

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+{
+    ngx_http_lua_semaphore_t          *sem;
+    ngx_http_lua_semaphore_t          *iter;
+    ngx_http_lua_main_conf_t          *lmcf;
+    ngx_queue_t                       *q;
+    ngx_uint_t                         i, n;
+    ngx_http_lua_semaphore_mm_block_t *block;
+    ngx_http_lua_semaphore_mm_t       *mm;
+
+    if (!ngx_cycle || !ngx_cycle->conf_ctx) {
```
agentzh: Are these checks protecting misuses as in contexts like `init_by_lua*`?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+        dd("in ngx_http_lua_alloc_semaphore"
+           "ngx_cycle or ngx_cycle->conf_ctx is null");
+
+        return NULL;
+    }
+
+    lmcf = ngx_http_cycle_get_module_main_conf(ngx_cycle,
+                                               ngx_http_lua_module);
+
+    if (!lmcf) {
```
agentzh: Style: we usually test against `NULL` for pointers, as in

```C
if (lmcf == NULL) {
    ...
}
```

But I wonder what contexts you're protecting against here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+            }
+
+            dd("free semaphore block: %p at final", block);
+
+            ngx_free(block);
+
+        } else {
+            ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
+               "ngx_http_lua_cleanup_semaphore_mm when cleanup"
+               " block: %p is still using by someone", block);
```
agentzh: Grammar: wrong passive voice again: "is still using by someone" -> "is still used by someone". Also, I'm afraid the colon (`:`) after "block" is not needed at all.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+                ngx_queue_remove(&iter->chain);
+            }
+
+            dd("free semaphore block: %p at final", block);
+
+            ngx_free(block);
+
+        } else {
+            ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
+               "ngx_http_lua_cleanup_semaphore_mm when cleanup"
```
agentzh: Style: arguments on the continued lines should align up with the first argument of the function call.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+               " block: %p is still using by someone", block);
+            /* some thing go wrong just return directly */
```
agentzh: Grammar: "some thing go wrong" -> "something goes wrong"

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+    ngx_http_lua_semaphore_mm_block_t *block;
+    ngx_http_lua_semaphore_mm_t       *mm;
+
+    block = sem->block;
+    block->used--;
+
+    mm = block->mm;
+    mm->used--;
+
+    start_epoch = mm->cur_epoch - mm->total/mm->num_per_block;
```
agentzh: Style: needs spaces around binary operators (`/` in this case). Please fix similar issues in other places too :)

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+        && block->epoch <= start_epoch + ((mm->cur_epoch - start_epoch) >> 1))
+    {
+        ngx_queue_insert_tail(&mm->free_queue, &sem->chain);
+
+    } else {
+        ngx_queue_insert_head(&mm->free_queue, &sem->chain);
+    }
+
+    if (block->used == 0 && mm->used < (mm->total >> 1)
+        && block->epoch <= (start_epoch + ((mm->cur_epoch-start_epoch) >> 1)))
```
agentzh: Style: spaces needed around the binary operator `-`.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+    {
+        /* load <= 50% and it's the older */
+        iter = (ngx_http_lua_semaphore_t*) (block+1);
```
agentzh: Style: spaces needed around the binary operator `+`. Also need a space before the `*` char.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+    if (rc == NGX_AGAIN) {
+        return ngx_http_lua_run_posted_threads(c, vm, r, ctx);
+    }
+
+    if (rc == NGX_DONE) {
+        ngx_http_lua_finalize_request(r, NGX_DONE);
+        return ngx_http_lua_run_posted_threads(c, vm, r, ctx);
+    }
+
+    /* rc == NGX_ERROR || rc >= NGX_OK */
```
agentzh: Style: needs a blank line after this comment line.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+        *errstr = "semaphore is null";
+        return NGX_ERROR;
+    }
+
+    sem->resource_count += n;
+
+    dd("ngx_http_lua_ffi_semaphore_post semaphore: %p, resource_count: %d",
+       sem, sem->resource_count);
+
+    if (!sem->in_post_event && !ngx_queue_empty(&sem->wait_queue)) {
```
agentzh: I'm afraid `in_post_event` is not a good name here. Grammatically, it should be "in_posted_event". But semantically this flag is to avoid posting duplicate events, I think `event_posted` is better. Thoughts?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
+    ngx_http_lua_semaphore_t *sem, int wait_ms, u_char *errstr, size_t *errlen)
+{
+    ngx_http_lua_ctx_t           *ctx;
+    ngx_http_lua_co_ctx_t        *wait_co_ctx;
+    ngx_int_t                     rc;
+
+    if (r == NULL) {
+        *errlen = ngx_snprintf(errstr, *errlen, "request is null") - errstr;
+         return NGX_ERROR;
```
agentzh: Style: 5-space indentation?

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+
+int
+ngx_http_lua_ffi_semaphore_wait(ngx_http_request_t *r,
+    ngx_http_lua_semaphore_t *sem, int wait_ms, u_char *errstr, size_t *errlen)
+{
+    ngx_http_lua_ctx_t           *ctx;
+    ngx_http_lua_co_ctx_t        *wait_co_ctx;
+    ngx_int_t                     rc;
+
+    if (r == NULL) {
```
agentzh: We usually do not check NULL request in FFI C API functions since it's the caller's responsibility to check that so that we can save one branch here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+ngx_http_lua_semaphore_cleanup(void *data)
+{
+    ngx_http_lua_co_ctx_t          *coctx = data;
+    ngx_queue_t                    *q;
+    ngx_http_lua_semaphore_t       *sem;
+
+    sem = coctx->data;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, sem->log, 0,
+                   "ngx_http_lua_semaphore_cleanup");
```
agentzh: Usually we do not use raw C function names in the debug log message. Traditionally, we use something like "http lua semaphore cleanup" as the message body here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+
+    return NGX_OK;
+}
+
+
+int
+ngx_http_lua_ffi_semaphore_post(ngx_http_lua_semaphore_t *sem,
+    int n, char **errstr)
+{
+    if (sem == NULL) {
```
agentzh: Better require the caller check `sem` so that we can save the check here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+
+    ngx_http_run_posted_requests(c);
+}
+
+
+void
+ngx_http_lua_ffi_semaphore_gc(ngx_http_lua_semaphore_t *sem)
+{
+    ngx_log_debug1(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                   "ngx_http_lua_ffi_semaphore_gc %p", sem);
```
agentzh: Better use a more descriptive debug log message here.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+
+    if (sem == NULL) {
+        return;
+    }
+
+    if (!ngx_queue_empty(&sem->wait_queue)) {
+        ngx_log_error(NGX_LOG_CRIT, ngx_cycle->log, 0,
+                       "ngx_http_lua_ffi_semaphore_gc wait queue is"
```
agentzh: Style: arguments on the continued lines should align up with the first argument of the function call.

agentzh: Also, we need a more descriptive error message here, as in "lua semaphore gc: wait queue blah blah blah".

src/ngx_http_lua_semaphore.h
```diff
@@ -0,0 +1,48 @@
...
+
+typedef struct ngx_http_lua_semaphore_s {
+    ngx_queue_t                          wait_queue;
+    ngx_queue_t                          chain;
+    ngx_event_t                          sem_event;
+    ngx_log_t                           *log;
+    ngx_http_lua_semaphore_mm_block_t   *block;
+    int                                  resource_count;
+    unsigned                             in_post_event:1;
+    unsigned                             wait_count:31;
```
agentzh: Better always put larger fields before smaller fields.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+                       " not empty while the semaphore: %p is "
```
agentzh: Grammar: "while the semaphore: %p is going to be destroyed" -> "while the semaphore %p is being destroyed"

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+    ngx_uint_t                           i;
+    ngx_http_lua_main_conf_t            *lmcf;
+    ngx_queue_t                         *q;
+    ngx_http_lua_semaphore_mm_block_t   *block;
+    ngx_http_lua_semaphore_mm_t         *mm;
+
+    lmcf = (ngx_http_lua_main_conf_t*) data;
+    mm = lmcf->semaphore_mm;
+
+    while (!ngx_queue_empty(&mm->free_queue)) {
```
agentzh: It seems to me that this loop never modifies the `mm->free_queue` queue and thus never terminates. Has this been tested?

BTW, the clang static analyzer reports the following warning:

```
src/ngx_http_lua_semaphore.c:138:13: warning: Use of memory after it is freed
        if (block->used == 0) {
            ^~~~~~~~~~~
```

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+
+        if (wait_co_ctx->sleep.timer_set) {
+            ngx_del_timer(&wait_co_ctx->sleep);
+        }
+
+        r = ngx_http_lua_get_req(wait_co_ctx->co);
+        c = r->connection;
+
+        ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
+        if (ctx == NULL) {
```
agentzh: Is this even possible? I think we should add an assertion here (with `ngx_http_lua_assert`).

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+    ngx_http_lua_co_ctx_t             *wait_co_ctx;
+    ngx_connection_t                  *c;
+    ngx_queue_t                       *q;
+
+    sem = ev->data;
+
+    while (!ngx_queue_empty(&sem->wait_queue) && sem->resource_count > 0) {
+
+        q = ngx_queue_head(&sem->wait_queue);
+        ngx_queue_remove(q);
```
agentzh: Let's fix the warning from the clang static analyzer:

```
src/ngx_http_lua_semaphore.c:416:9: warning: Access to field 'prev' results in a dereference of a null pointer (loaded from field 'next')
        ngx_queue_remove(q);
        ^~~~~~~~~~~~~~~~~~~
nginx/src/core/ngx_queue.h:73:21: note: expanded from macro 'ngx_queue_remove'
    (x)->next->prev = (x)->prev;                                              \
         ~~~~       ^
```

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+       sem, sem->resource_count);
+
+    if (!sem->in_post_event && !ngx_queue_empty(&sem->wait_queue)) {
+
+        sem->sem_event.handler = ngx_http_lua_semaphore_handler;
+        sem->sem_event.data = sem;
+        sem->sem_event.log = sem->log;
+
+        sem->in_post_event = 1;
+        ngx_post_event((&sem->sem_event), &ngx_posted_events);
```
agentzh: Style: we don't need the parentheses around `&sem->sem_event` since no possible ambiguity here.

agentzh: Valgrind memcheck reports the following error on this line:

```
==119538== Conditional jump or move depends on uninitialised value(s)
==119538==    at 0x4D58D7: ngx_http_lua_ffi_semaphore_post (ngx_http_lua_semaphore.c:308)
==119538==    by 0x505D245: lj_vm_ffi_call (in /opt/luajit21sysm/lib/libluajit-5.1.so.2.1.0)
==119538==    by 0x50A9FE7: lj_ccall_func (lj_ccall.c:962)
==119538==    by 0x50BEF95: lj_cf_ffi_meta___call (lib_ffi.c:230)
==119538==    by 0x505B177: lj_BC_FUNCC (in /opt/luajit21sysm/lib/libluajit-5.1.so.2.1.0)
==119538==    by 0x4C3D59: ngx_http_lua_run_thread (ngx_http_lua_util.c:1015)
==119538==    by 0x4D4C44: ngx_http_lua_sleep_resume (ngx_http_lua_sleep.c:169)
==119538==    by 0x4D4D7C: ngx_http_lua_sleep_handler (ngx_http_lua_sleep.c:118)
==119538==    by 0x434D7D: ngx_event_expire_timers (ngx_event_timer.c:96)
==119538==    by 0x4349B3: ngx_process_events_and_timers (ngx_event.c:256)
==119538==    by 0x43C812: ngx_single_process_cycle (ngx_process_cycle.c:309)
==119538==    by 0x41B14C: main (nginx.c:356)
```

Seems like you forgot to initialize the `posted` field of `sem->sem_event`.

agentzh:
The title of this PR has a grammatical mistake: "a simple implement of ..." -> "a simple implementation of ...".

doc/HttpLuaModule.wiki
cuiweixie: I wonder whether can we use a new in example. It seems a new file make things simple.

agentzh: @cuiweixie We can and should use a separate document file for this `ngx.semaphore` module, similar to this:

https://github.com/openresty/lua-resty-limit-traffic/blob/master/lib/resty/limit/traffic.md

Then we can just reference it elsewhere as `[resty.limit.traffic](lib/resty/limit/traffic.md)`. Easy and simple :)

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,523 @@
...
+        ngx_queue_insert_tail(&mm->free_queue, &sem->chain);
+
+    } else {
+        ngx_queue_insert_head(&mm->free_queue, &sem->chain);
+    }
+
+    if (block->used == 0 && mm->used < (mm->total >> 1)
+        && block->epoch <= (start_epoch + ((mm->cur_epoch-start_epoch) >> 1)))
+    {
+        /* load <= 50% and it's the older */
```
cuiweixie: this is a comment. So do we need a blank line after it?

src/ngx_http_lua_semaphore.c
cuiweixie: Is ngx_http_lua_assert is better here?

src/ngx_http_lua_semaphore.c
agentzh: @cuiweixie Not really in this context.

src/ngx_http_lua_semaphore.c
agentzh: Yeah.

agentzh:
Just a quick note: this PR requires the following PR for lua-resty-core to be fully functional:

https://github.com/openresty/lua-resty-core/pull/5

cuiweixie:
@agentzh  Please check this PR again at your convenience.

agentzh:
@cuiweixie Cool, thanks. I'll reiterate through it. Thanks!

doc/HttpLuaModule.wiki
```diff
@@ -5806,7 +5806,6 @@ Kills a running "light thread" created by [[#ngx.thread.spawn|ngx.thread.spawn]]
 According to the current implementation, only the parent coroutine (or "light thread") can kill a thread. Also, a running "light thread" with pending NGINX subrequests (initiated by [[#ngx.location.capture|ngx.location.capture]] for example) cannot be killed due to a limitation in the NGINX core.
 
 This API was first enabled in the <code>v0.9.9</code> release.
-
```
agentzh: Why this?

doc/HttpLuaModule.wiki
agentzh: Still no fix here?

doc/HttpLuaModule.wiki
cuiweixie: the blank may be added by accident, when @doujiang24 merge the latest version of ngx_lua which we need the *_by_lua_block in.

src/ngx_http_lua_semaphore.c
```diff
@@ -0,0 +1,519 @@
+
+/*
+ * Copyright (C) cuiweixie
```
agentzh: @cuiweixie Will you please state that "I hereby assign copyright in this code to the lua-nginx-module project, to be licensed under the same terms as the rest of the code"? This is just for legal resolution. Thanks for your understanding.

doujiang24:
@agentzh I have fixed a bug and added some test cases. Please check this PR again at your convenience.

agentzh:
@doujiang24 Awesome. I'll re-iterate through the PRs today :)

agentzh:
I've made some changes locally. Please have a look:

```diff
diff --git a/config b/config
index 101fd19..a0ed75d 100644
--- a/config
+++ b/config
@@ -462,4 +462,3 @@ ngx_feature_test='setsockopt(1, SOL_SOCKET, SO_PASSCRED, NULL, 0);'
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_PATH='\"/usr/local/openresty/lualib/?.lua\"'"
 #CFLAGS=$"$CFLAGS -DLUA_DEFAULT_CPATH='\"/usr/local/openresty/lualib/?.so\"'"
 
-
diff --git a/src/ngx_http_lua_common.h b/src/ngx_http_lua_common.h
index 5d9cd73..f03e4d8 100644
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -105,7 +105,7 @@ typedef struct {
 typedef struct ngx_http_lua_main_conf_s  ngx_http_lua_main_conf_t;
 
 
-typedef struct ngx_http_lua_semaphore_mm_s ngx_http_lua_semaphore_mm_t;
+typedef struct ngx_http_lua_semaphore_mm_s  ngx_http_lua_semaphore_mm_t;
 
 
 typedef ngx_int_t (*ngx_http_lua_conf_handler_pt)(ngx_log_t *log,
diff --git a/src/ngx_http_lua_module.c b/src/ngx_http_lua_module.c
index 29f7113..f37d684 100644
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -553,6 +553,7 @@ ngx_http_lua_init(ngx_conf_t *cf)
     ngx_int_t                   rc;
     ngx_array_t                *arr;
     ngx_http_handler_pt        *h;
+    volatile ngx_cycle_t       *saved_cycle;
     ngx_http_core_main_conf_t  *cmcf;
     ngx_http_lua_main_conf_t   *lmcf;
     ngx_pool_cleanup_t         *cln;
@@ -659,7 +660,14 @@ ngx_http_lua_init(ngx_conf_t *cf)
         }
 
         if (!lmcf->requires_shm && lmcf->init_handler) {
-            if (lmcf->init_handler(cf->log, lmcf, lmcf->lua) != NGX_OK) {
+            saved_cycle = ngx_cycle;
+            ngx_cycle = cf->cycle;
+
+            rc = lmcf->init_handler(cf->log, lmcf, lmcf->lua);
+
+            ngx_cycle = saved_cycle;
+
+            if (rc != NGX_OK) {
                 /* an error happened */
                 return NGX_ERROR;
             }
diff --git a/src/ngx_http_lua_semaphore.c b/src/ngx_http_lua_semaphore.c
index a923c1d..b211d26 100644
--- a/src/ngx_http_lua_semaphore.c
+++ b/src/ngx_http_lua_semaphore.c
@@ -3,9 +3,10 @@
  * Copyright (C) Yichun Zhang (agentzh)
  * Copyright (C) cuiweixie
  * I hereby assign copyright in this code to the lua-nginx-module project,
- * to be licensed under the same terms as the rest of the code
+ * to be licensed under the same terms as the rest of the code.
  */
 
+
 #ifndef NGX_LUA_NO_FFI_API
 
 
diff --git a/src/ngx_http_lua_semaphore.h b/src/ngx_http_lua_semaphore.h
index 6dcf6e3..b53bff7 100644
--- a/src/ngx_http_lua_semaphore.h
+++ b/src/ngx_http_lua_semaphore.h
@@ -3,7 +3,7 @@
  * Copyright (C) Yichun Zhang (agentzh)
  * Copyright (C) cuiweixie
  * I hereby assign copyright in this code to the lua-nginx-module project,
- * to be licensed under the same terms as the rest of the code
+ * to be licensed under the same terms as the rest of the code.
  */
 
 
```

If it's okay, then I'll merge this PR with my changes to master. Thanks!

agentzh:
@cuiweixie @doujiang24 An updated version of my local changes is here:

https://gist.github.com/agentzh/f3b8bca600a676f0f4db

Thanks!

cuiweixie:
@agentzh, thank you! I think it is ok!

doujiang24:
@agentzh Thanks a lot! It's ok for me too.

agentzh:
Okay, a slightly modified version of the patch has just been applied to git master. Thank you for your contribution!

Ahsialh:
wow,so cooooooooooool

Issue #552, opened 2015-08-24
shared.DICT and expiration time


bungle:
I would like to see a way to retrieve and set shared dictionary expiration time.

E.g.

``` lua
value, flags, exptime = ngx.shared.DICT:get(key)
-- it would be nicer to have same order as in set (but that would break bc):
value, exptime, flags = ngx.shared.DICT:get(key)
```

Or

``` lua
exptime = ngx.shared.DICT:expires(key)
```

And

``` lua
success, err = ngx.shared.DICT:expire(key, exptime)
```

Does it make sense or is it unneccessarely difficult and bloat?

agentzh:
@bungle Maybe we can just borrow the redis command names here? i.e.,

```lua
local exptime, err = dict:ttl(key)
local ok, err = dict:expire(key, seconds)
```

The direction is to support more and more redis commands in lua_shared_dict in the future :)

bungle:
@agentzh, oh, I had forgotten that for lua_shared_dict you had such a broad goals. That would work just as well. I kinda had an use case for this, and that's why I asked. You see I have been adding pluggable storage adapters to lua-resty-session. One of the adapters is shm that uses lua_shared_dict, as you can see here:

https://github.com/bungle/lua-resty-session/blob/storages/lib/resty/session/storage/shm.lua#L31

The line 31 is really just updating the expiration time, and it is unneccessary to update the value at that point.

agentzh:
@bungle Fair enough :) Will you contribute a patch for this feature?

bungle:
I will look at it.

agentzh:
Cool, thanks!

navneet-flipkart:
Even I felt the need of such a feature. Any update on whether this is available in any release?

agentzh:
@navneet-flipkart Not yet. The PR is still being reviewed.

guanhui07:
queue

Gitzhaoyang:
queue... I need it

junos:
Need this as well

agentzh:
@guanhui07 @Gitzhaoyang the shdict list data type support has already been released in the latest OpenResty version 1.11.2.1.

@junos I'm not sure what you are referring to. This ticket is more about expiration time, isn't it?

thibaultcha:
Done in #140 - we can close this

PR #586, opened 2015-10-28
feature: add api in shdict: lpush, lpop, rpush, rpop, llen

doujiang24:
I'm not sure if it is an good idea to get the list queue head in the way:
```
#define ngx_http_lua_shdict_get_list_head(sd, key_len)                      \
    (ngx_queue_t *) ngx_align_ptr(((u_char *) &sd->data + key_len),         \
                                  NGX_ALIGNMENT)
```

Anything not suitable, let me know, thanks :)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6316,7 +6316,7 @@ ngx.shared.DICT.lpush
 
 **context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
 
-Insert the specified (numerical or string) `value` at the head of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict). Returns the length of the list after the push operations.
+Inserts the specified (numerical or string) `value` at the head of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict). Returns the length of the list after the push operations.
 
 If `key` does not exist, it is created as empty list before performing the push operations. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
 
@@ -6334,7 +6334,7 @@ ngx.shared.DICT.rpush
 
 **context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
 
-Similar to the [lpush](#ngxshareddictlpush) method, but insert the specified (numerical or string) `value` at the tail of the list named `key`.
+Similar to the [lpush](#ngxshareddictlpush) method, but inserts the specified (numerical or string) `value` at the tail of the list named `key`.
 
 This feature was first introduced in the `v0.*.*` release.
 
--- a/README.markdown
+++ b/README.markdown
@@ -3041,6 +3041,12 @@ Nginx API for Lua
 * [ngx.shared.DICT.replace](#ngxshareddictreplace)
 * [ngx.shared.DICT.delete](#ngxshareddictdelete)
 * [ngx.shared.DICT.incr](#ngxshareddictincr)
+* [ngx.shared.DICT.lpush](#ngxshareddictlpush)
+* [ngx.shared.DICT.rpush](#ngxshareddictrpush)
+* [ngx.shared.DICT.lpop](#ngxshareddictlpop)
+* [ngx.shared.DICT.rpop](#ngxshareddictrpop)
+* [ngx.shared.DICT.llen](#ngxshareddictllen)
+* [ngx.shared.DICT.expire](#ngxshareddictexpire)
 * [ngx.shared.DICT.flush_all](#ngxshareddictflush_all)
 * [ngx.shared.DICT.flush_expired](#ngxshareddictflush_expired)
 * [ngx.shared.DICT.get_keys](#ngxshareddictget_keys)
@@ -6035,6 +6041,12 @@ The resulting object `dict` has the following methods:
 * [replace](#ngxshareddictreplace)
 * [delete](#ngxshareddictdelete)
 * [incr](#ngxshareddictincr)
+* [lpush](#ngxshareddictlpush)
+* [rpush](#ngxshareddictrpush)
+* [lpop](#ngxshareddictlpop)
+* [rpop](#ngxshareddictrpop)
+* [llen](#ngxshareddictllen)
+* [expire](#ngxshareddictexpire)
 * [flush_all](#ngxshareddictflush_all)
 * [flush_expired](#ngxshareddictflush_expired)
 * [get_keys](#ngxshareddictget_keys)
@@ -6293,6 +6305,102 @@ The `value` argument and `init` argument can be any valid Lua numbers, like nega
 This method was first introduced in the `v0.3.1rc22` release.
 
 The optional `init` parameter was first added in the `v0.10.6` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.lpush
+---------------------
+**syntax:** *length, err = ngx.shared.DICT:lpush(key, value)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Insert the specified (numerical or string) `value` at the head of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict). Returns the length of the list after the push operations.
+
+If `key` does not exist, it is created as empty list before performing the push operations. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
+
+It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return `nil` and the string "no memory".
+
+This feature was first introduced in the `v0.*.*` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.rpush
+---------------------
+**syntax:** *length, err = ngx.shared.DICT:rpush(key, value)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Similar to the [lpush](#ngxshareddictlpush) method, but insert the specified (numerical or string) `value` at the tail of the list named `key`.
+
+This feature was first introduced in the `v0.*.*` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.lpop
+--------------------
+**syntax:** *val, err = ngx.shared.DICT:lpop(key)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Removes and returns the first element of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+If `key` does not exist, it will return `nil`. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
+
+This feature was first introduced in the `v0.*.*` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.rpop
+--------------------
+**syntax:** *val, err = ngx.shared.DICT:rpop(key)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Removes and returns the last element of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+If `key` does not exist, it will return `nil`. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
+
+This feature was first introduced in the `v0.*.*` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.llen
+--------------------
+**syntax:** *len, err = ngx.shared.DICT:llen(key)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Returns the length of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+If key does not exist, it is interpreted as an empty list and 0 is returned. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
+
+This feature was first introduced in the `v0.*.*` release.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.expire
+----------------------
+**syntax:** *ok, err = ngx.shared.DICT:expire(key, exptime)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+The `exptime` argument specifies expiration time (in seconds) for the inserted key-value pair. The time resolution is 0.001 seconds. If the exptime takes the value 0, then the item will never be expired.
+
+If key does not exist or has been expired, it will return `false` and `"not found"`. Otherwise it will return `true`.
+
+This feature was first introduced in the `v0.*.*` release.
 
 See also [ngx.shared.DICT](#ngxshareddict).
 
--- a/README.markdown
+++ b/README.markdown
@@ -3046,7 +3046,6 @@ Nginx API for Lua
 * [ngx.shared.DICT.lpop](#ngxshareddictlpop)
 * [ngx.shared.DICT.rpop](#ngxshareddictrpop)
 * [ngx.shared.DICT.llen](#ngxshareddictllen)
-* [ngx.shared.DICT.expire](#ngxshareddictexpire)
 * [ngx.shared.DICT.flush_all](#ngxshareddictflush_all)
 * [ngx.shared.DICT.flush_expired](#ngxshareddictflush_expired)
 * [ngx.shared.DICT.get_keys](#ngxshareddictget_keys)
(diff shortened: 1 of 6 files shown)


src/ngx_http_lua_shdict.c
```diff
@@ -1473,92 +1556,775 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+
+
+static int
+ngx_http_lua_shdict_rpop(lua_State *L)
+{
+    return ngx_http_lua_shdict_pop_helper(L, NGX_HTTP_LUA_SHDICT_RIGHT);
+}
+
+
+
```
agentzh: Style: 2 blank lines needed instead of 3.

src/ngx_http_lua_shdict.c
```diff
@@ -1051,6 +1128,21 @@ ngx_http_lua_shdict_set_helper(lua_State *L, int flags)
 
 remove:
 
+        if (sd->value_type == LUA_TTABLE) {
+            queue = ngx_http_lua_shdict_get_list_head(sd, key.len);
+
+            for (q = ngx_queue_head(queue);
+                q != ngx_queue_sentinel(queue);
```
agentzh: Style: need to get aligned with the `q` in the previous line.

src/ngx_http_lua_shdict.c
```diff
@@ -1473,92 +1556,775 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+                       "NOT matched, removing it first");
+
+remove:
+
+        if (sd->value_type == LUA_TTABLE) {
+
+            queue = ngx_http_lua_shdict_get_list_head(sd, key_len);
+
+            for (q = ngx_queue_head(queue);
+                q != ngx_queue_sentinel(queue);
```
agentzh: Style: alignment needed.

src/ngx_http_lua_shdict.c
```diff
@@ -1473,92 +1556,775 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+
+    if (rc == NGX_OK || rc == NGX_DONE) {
+
+        if (value_type == LUA_TNIL) {
+            goto remove;
+        }
+
+replace:
+
+        if (str_value_buf && str_value_len == (size_t) sd->value_len
```
agentzh: Style: better break the `&&` to align with the `&&` in the following line.

src/ngx_http_lua_shdict.c
```diff
@@ -1473,92 +1556,775 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+    dd("lookup returns %d", (int) rc);
+
+    if (op & NGX_HTTP_LUA_SHDICT_REPLACE) {
+
+        if (rc == NGX_DECLINED || rc == NGX_DONE) {
+            ngx_shmtx_unlock(&ctx->shpool->mutex);
+            *errmsg = "not found";
+            return NGX_DECLINED;
+        }
+        /* rc == NGX_OK */
```
agentzh: Style: a blank line needed before this line.

src/ngx_http_lua_shdict.c
```diff
@@ -1473,92 +1556,775 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+    dd("exptime: %d", exptime);
+
+    ctx = zone->data;
+
+    *forcible = 0;
+
+    hash = ngx_crc32_short(key, key_len);
+
+    switch (value_type) {
+    case LUA_TSTRING:
```
agentzh: Style: a blank line needed before this line (I understand existing code might miss this blank line after the `switch` line, but they are to be fixed).

README.markdown
```diff
@@ -5860,6 +5872,102 @@ See also [ngx.shared.DICT](#ngxshareddict).
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.shared.DICT.lpush
+---------------------
+**syntax:** *length, err = ngx.shared.DICT:lpush(key, value)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Insert the specified (numerical or string) `value` at the head of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict). Returns the length of the list after the push operations.
```
agentzh: Should be "Inserts".

src/ngx_http_lua_shdict.h
```diff
@@ -25,6 +25,14 @@ typedef struct {
 
 
 typedef struct {
+    uint8_t                      value_type;
+    uint32_t                     value_len;
+    ngx_queue_t                  queue;
+    u_char                       data[1];
+} ngx_http_lua_shdict_list_node_t;
```
agentzh: I suggest reordering the struct fields above to minimize padding overhead.

agentzh:
@doujiang24 The feature implemented in this PR is awesome. I'll have a closer look as soon as I can manage (need to cut a new OpenResty release ASAP).

bungle:
Yes, this is great, thanks @doujiang24. I was going to look at implementing just one thing, as discussed in #552, but you went ahead and implemented a lot more of these Maybe when you are up to the speed with this, you can also take a look at implementing ttl ( expires cousin, :-) )?

src/ngx_http_lua_shdict.h
doujiang24: Nice, three bytes saved :+1:

doujiang24:
@agentzh Thanks for your review :)
I just fixed some code style and a stupid bug.

doujiang24:
@bungle Thanks, I tried to implement `ttl` and `type` api, but it brought some duplicated codes. I haven't come up with an idea to fix it. I'll try again later after this PR  is merged :)

thibaultCha:
Wow, that is really nice! Good job.

dualface:
+1

guanglinlv:
++1

src/ngx_http_lua_shdict.c
```diff
@@ -43,13 +51,20 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
...
+#define NGX_HTTP_LUA_SHDICT_LEFT        0x0001
+#define NGX_HTTP_LUA_SHDICT_RIGHT       0x0002
 
 
 enum {
     SHDICT_USERDATA_INDEX = 1,
 };
 
 
+#define ngx_http_lua_shdict_get_list_head(sd, key_len)                      \
```
agentzh: Shall we use an inline function here instead of a macro? Thanks!

src/ngx_http_lua_shdict.c
```diff
@@ -43,13 +51,20 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
 #define NGX_HTTP_LUA_SHDICT_ADD         0x0001
 #define NGX_HTTP_LUA_SHDICT_REPLACE     0x0002
 #define NGX_HTTP_LUA_SHDICT_SAFE_STORE  0x0004
+#define NGX_HTTP_LUA_SHDICT_LEFT        0x0001
```
agentzh: Better put a blank line above this line since these 2 constants do not belong to the previous ones.

src/ngx_http_lua_shdict.c
```diff
@@ -531,6 +581,14 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
         lua_pushboolean(L, c ? 1 : 0);
         break;
 
+    case LUA_TTABLE:
+
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+        lua_pushnil(L);
+        lua_pushliteral(L, "wrongtype operation");
```
agentzh: The error message needs to be more descriptive.

src/ngx_http_lua_shdict.c
```diff
@@ -531,6 +581,14 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
         lua_pushboolean(L, c ? 1 : 0);
         break;
 
+    case LUA_TTABLE:
```
agentzh: I think using `LUA_TTABLE` here is inappropriate since a Lua table can be either an array or a hash table or both at the same time. We'll eventually have hash-table value type in shdict. So better invent our own type constants instead of abusing Lua's types?

src/ngx_http_lua_shdict.c
doujiang24: I found another question now:
Can we use `shdict.set` on list type key even when key is existed and not expired?
It's not allow now in this PR implementing, but I tested redis, it's ok to set key value on existed list type key.

agentzh: @doujiang24 Yes, `set` should always be unconditional.

agentzh:
@doujiang24 Will you please rebase it to the latest master branch? Thanks!

src/ngx_http_lua_shdict.c
```diff
@@ -1482,107 +1582,794 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+            lua_pushliteral(L, "wrongtype operation");
+            return 2;
+        }
+
+        ngx_queue_remove(&sd->queue);
+        ngx_queue_insert_head(&ctx->sh->queue, &sd->queue);
+
+        lua_pushnumber(L, (lua_Number) sd->value_len);
+
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
```
agentzh: Better put `ngx_shmtx_unlock()` before `lua_pushnumber()` because the latter might throw. Thanks!

doc/HttpLuaModule.wiki
```diff
@@ -5163,6 +5169,84 @@ This feature was first introduced in the <code>v0.3.1rc22</code> release.
...
 See also [[#ngx.shared.DICT|ngx.shared.DICT]].
 
+== ngx.shared.DICT.lpush ==
+'''syntax:''' ''length, err = ngx.shared.DICT:lpush(key, value)''
+
+'''context:''' ''init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.*''
+
+Inserts the specified (numerical or string) <code>value</code> at the head of the list named <code>key</code> in the shm-based dictionary [[#ngx.shared.DICT|ngx.shared.DICT]]. Returns the length of the list after the push operations.
+
+If <code>key</code> does not exist, it is created as empty list before performing the push operations. When key holds a value that is not a list, it will return <code>nil</code> and <code>"wrongtype operation"</code>.
```
agentzh: I think the error string `value not a list`  is better than `wrongtype operation`.

src/ngx_http_lua_shdict.c
```diff
@@ -1482,107 +1582,794 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+            lnode = ngx_queue_data(q, ngx_http_lua_shdict_list_node_t, queue);
+            ngx_slab_free_locked(ctx->shpool, lnode);
+        }
 
-            sd->value_len = (uint32_t) str_value_len;
+        ngx_queue_init(queue);
 
-            dd("setting value type to %d", value_type);
+        ngx_queue_remove(&sd->queue);
+        ngx_queue_insert_head(&ctx->sh->queue, &sd->queue);
```
agentzh: Because we have two different queues here, list-typed values and the global LRU queue, maybe it's the time to rename `ctx->sh->queue` to `ctx->sh->lru_queue` to be clearer? Frankly I stumbled on these lines and got confused with these two different queues mixed together in the code base ;)

src/ngx_http_lua_shdict.c
```diff
@@ -1482,107 +1582,794 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
 
-            sd->value_type = (uint8_t) value_type;
+        dd("go to push_node");
+        goto push_node;
 
-            p = ngx_copy(sd->data, key, key_len);
-            ngx_memcpy(p, str_value_buf, str_value_len);
+    } else if (rc == NGX_OK) {
```
agentzh: I don't think we need an `else` here since the previous branch always jumps away via `goto push_node`, right?

src/ngx_http_lua_shdict.c
```diff
@@ -1482,107 +1582,794 @@ ngx_http_lua_ffi_shdict_store(ngx_shm_zone_t *zone, int op, u_char *key,
...
+
+        node = (ngx_rbtree_node_t *)
+                    ((u_char *) sd - offsetof(ngx_rbtree_node_t, color));
+
+        ngx_rbtree_delete(&ctx->sh->rbtree, node);
+
+        ngx_slab_free_locked(ctx->shpool, node);
+
+    } else {
+
```
agentzh: Style: extra blank line here.

src/ngx_http_lua_shdict.c
```diff
@@ -44,11 +52,30 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
...
+enum {
+    SHDICT_TNIL = 0,        /* same as LUA_TNIL */
+    SHDICT_TBOOLEAN = 1,    /* same as LUA_TBOOLEAN */
+    SHDICT_TNUMBER = 3,     /* same as LUA_TNUMBER */
+    SHDICT_TSTRING = 4,     /* same as LUA_TSTRING */
+    SHDICT_TLIST = 5,
+};
+
+
+static ngx_queue_t *
```
agentzh: Better add the `ngx_inline` word after `static` here.

src/ngx_http_lua_shdict.c
```diff
@@ -44,11 +52,30 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
 #define NGX_HTTP_LUA_SHDICT_REPLACE     0x0002
 #define NGX_HTTP_LUA_SHDICT_SAFE_STORE  0x0004
 
+#define NGX_HTTP_LUA_SHDICT_LEFT        0x0001
+#define NGX_HTTP_LUA_SHDICT_RIGHT       0x0002
+
 
 enum {
     SHDICT_USERDATA_INDEX = 1,
 };
```
agentzh: Style: need another blank line here (we need for total 2 blank lines).

src/ngx_http_lua_shdict.c
```diff
@@ -44,11 +52,30 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
 #define NGX_HTTP_LUA_SHDICT_REPLACE     0x0002
 #define NGX_HTTP_LUA_SHDICT_SAFE_STORE  0x0004
```
agentzh: Style: need another blank line here

README.markdown
```diff
@@ -6164,6 +6176,102 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.expire
+----------------------
+**syntax:** *ok, err = ngx.shared.DICT:expire(key, exptime)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+The `exptime` argument specifies expiration time (in seconds) for the inserted key-value pair. The time resolution is 0.001 seconds. If the exptime takes the value 0, then the item will never be expired.
```
agentzh: It's better to be "... specifies a new expiration time ...".

Also "for the inserted key-value pair" is better to be "for an existing key-value pair in the shm store specified by the `key` argument".

agentzh: Grammar: "the item will never be expired" => "the item will never expire". We shouldn't use the passive voice here. Please fix other places as well.

README.markdown
```diff
@@ -6164,6 +6176,102 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+
+ngx.shared.DICT.rpop
+--------------------
+**syntax:** *val, err = ngx.shared.DICT:rpop(key)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Removes and returns the last element of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+If `key` does not exist, it will return `nil`. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
```
agentzh: "When key holds a value" => "When `key` holds a value".

README.markdown
```diff
@@ -6164,6 +6176,102 @@ See also [ngx.shared.DICT](#ngxshareddict).
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.shared.DICT.lpush
+---------------------
+**syntax:** *length, err = ngx.shared.DICT:lpush(key, value)*
+
+**context:** *init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua*, log_by_lua*, ngx.timer.**
+
+Inserts the specified (numerical or string) `value` at the head of the list named `key` in the shm-based dictionary [ngx.shared.DICT](#ngxshareddict). Returns the length of the list after the push operations.
```
agentzh: The "length of the list" is ambiguous. Is it a length in bytes or the number of elements? We should be clearer here. Better be "Returns the number of elements in the list after the push operation". Oh yeah, "operations" should be "operation" since it is a single operation, not multiple.

README.markdown
```diff
@@ -6164,6 +6176,102 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+
+If `key` does not exist, it is created as empty list before performing the push operations. When key holds a value that is not a list, it will return `nil` and `"wrongtype operation"`.
```
agentzh: "When key holds a value that is not a list" => "When the key already takes a value that is not a list".

agentzh: "it is created as empty list" => "it is created as an empty list".

agentzh:
It seems to me that the `expire` method is irrelevant to the rest of the changes in this PR. I think it's better to move it to a separate PR to simplify the review and merge work. What do you think? I prefer small and self-contained PRs over mixed ones.

src/ngx_http_lua_shdict.c
doujiang24: I have rename `queue` to `lru_queue` in `ngx_http_lua_shdict_shctx_t` (`ctx->sh->queue`).
Do we also need to rename `queue` to `lru_queue`  in `ngx_http_lua_shdict_node_t` (`sd->queue`) ?

doujiang24:
@agentzh Thanks for your review :)
All fixed and the `expire` method may be another PR after this PR merged :)

src/ngx_http_lua_shdict.c
```diff
@@ -1534,7 +2206,7 @@ ngx_http_lua_find_zone(u_char *name_data, size_t name_len)
         if (name->len == name_len
             && ngx_strncmp(name->data, name_data, name_len) == 0)
         {
-            return &zone[i];
+            return zone;
```
agentzh: I'm afraid this change is wrong and bad.

doujiang24: Sorry, this may be something wrong from rebase :(
I have rechecked the left change :)

agentzh:
Merged into master already.

README.markdown
yangshuxin: In "push operations",  the operation does not need to take plural form. Or maybe we can just say, "Returns the length of the resulting list".

PR #590, opened 2015-10-29
optimize: reduce memory alloc

doujiang24:
1) not need to alloc for host when get peer from cosocket_pool
2) reuse cosocket cleanup

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -372,7 +372,7 @@ typedef struct ngx_http_lua_ctx_s {
 
     ngx_http_cleanup_pt     *cleanup;
 
-    ngx_http_cleanup_t      *free_cleanup;
+    ngx_http_cleanup_t      *free_cleanup; /* free list of cleanup records */
 
     ngx_chain_t             *body; /* buffered subrequest response body
                                       chains */
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -372,6 +372,8 @@ typedef struct ngx_http_lua_ctx_s {
 
     ngx_http_cleanup_pt     *cleanup;
 
+    ngx_http_cleanup_t      *free_cleanup;
+
     ngx_chain_t             *body; /* buffered subrequest response body
                                       chains */
 
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -600,6 +600,8 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
 
     /* rc == NGX_DECLINED */
 
+    /* TODO: we should avoid this in-pool allocation */
+
     host.data = ngx_palloc(r->pool, len + 1);
     if (host.data == NULL) {
         return luaL_error(L, "no memory");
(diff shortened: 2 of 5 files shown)


src/ngx_http_lua_common.h
```diff
@@ -372,6 +372,8 @@ typedef struct ngx_http_lua_ctx_s {
 
     ngx_http_cleanup_pt     *cleanup;
 
+    ngx_http_cleanup_t      *free_cleanup;
```
agentzh: Better add a comment here: "free list of cleanup records".

src/ngx_http_lua_socket_tcp.c
```diff
@@ -610,6 +600,16 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
 
     /* rc == NGX_DECLINED */
 
+    host.data = ngx_palloc(r->pool, len + 1);
```
agentzh: Better add a comment here: "TODO: we should avoid this in-pool allocation".

src/ngx_http_lua_util.c
```diff
@@ -3980,4 +3985,84 @@ ngx_http_lua_get_raw_phase_context(lua_State *L)
...
 }
 
+
+ngx_http_cleanup_t *
+ngx_http_lua_cleanup_add(ngx_http_request_t *r, size_t size)
+{
+    ngx_http_cleanup_t  *cln;
+    ngx_http_lua_ctx_t  *ctx;
+
+    r = r->main;
```
agentzh: I don't think this line is needed here and actually this may cause problems since the main request may not use any Lua at all, in which case the ngx_lua ctx for the main request is always NULL and we lose cleanup record recycling in the suberquest using Lua. The same applies to the `ngx_http_lua_cleanup_free` function below.

src/ngx_http_lua_util.c
```diff
@@ -3980,4 +3985,84 @@ ngx_http_lua_get_raw_phase_context(lua_State *L)
...
+        return;
+    }
+
+    cln = (ngx_http_cleanup_t *)
+              ((u_char *) cleanup - offsetof(ngx_http_cleanup_t, handler));
+
+    dd("cln: %p, cln->handler: %p, &cln->handler: %p",
+        cln, cln->handler, &cln->handler);
+
+    if (r->cleanup == cln) {
```
agentzh: We can simplify the code from this line to line 4067 as follows:

```C
ngx_http_cleanup_t    **last;
...
last = &r->cleanup;
while (*last) {
    if (*last == cln) {
        *last = cln->next;
        cln->next = ctx->free_cleanup;
        ctx->free_cleanup = cln;
        ngx_log_debug1(...);
        return;
    }
    last = &(*last)->next;
}
```

src/ngx_http_lua_util.c
agentzh: Oh, we should add a test case to exercise this special subrequest case stated above as well :)

doujiang24:
@agentzh I committed a patch, thanks for your review :)

agentzh:
@doujiang24 Thanks for your quick fix! I'll try merging this later today :)

agentzh:
Just a side note: the cleanup recycling optimization should apply to other parts of the ngx_lua core like udp cosockets and timers. Such changes can go with another PR.

agentzh:
@doujiang24 Patch applied to git master. Thanks for your contribution!

PR #596, opened 2015-11-03
bugfix: missing right long brackets when "--" before it

doujiang24:
I got error: "nginx: [emerg] Lua code block missing the closing long bracket"

```
content_by_lua_block {
     ngx.say([[ok--]])
}
```

### DIFF
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -52,6 +52,17 @@ struct ngx_http_lua_block_parser_ctx_s {
     int         expect_right_lbracket;
 };
 
+enum {
+    FOUND_LEFT_CURLY = 0,
+    FOUND_RIGHT_CURLY,
+    FOUND_LEFT_LBRACKET_STR,
+    FOUND_LEFT_LBRACKET_CMT,
+    FOUND_RIGHT_LBRACKET,
+    FOUND_COMMENT_LINE,
+    FOUND_DOUBLE_QUOTED,
+    FOUND_SINGLE_QUOTED
+};
+
 
 char *
 ngx_http_lua_shared_dict(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_directive.c
```diff
@@ -52,6 +52,17 @@ struct ngx_http_lua_block_parser_ctx_s {
     int         expect_right_lbracket;
 };
 
+enum {
```
agentzh: Style: need a blank line above this.

src/ngx_http_lua_directive.c
```diff
@@ -1602,7 +1603,12 @@ ngx_http_lua_conf_read_lua_token(ngx_conf_t *cf,
             }
         }
 
-        b->pos += ovec[1];
+        if (ctx->expect_right_lbracket && rc == FOUND_COMMENT_LINE) {
+            b->pos += (ovec[0] + 2);
```
agentzh: We can omit the parentheses here? I think there's no ambiguity here :)

agentzh: Hmm, I think it's better to avoid returning `FOUND_COMMENT_LINE` in this case since we are not really seeing a comment line. Maybe directly skip the current token and try reading the next one? If that's too complicated, maybe we can just introduce a `FOUND_DONTCARE` psudo token and return that instead? Thoughts?

agentzh:
@doujiang24 Thanks for catching this corner case. Glad it won't creep into the upcoming OpenResty 1.9.3.2 formal release ;)

src/ngx_http_lua_directive.c
doujiang24: I choose the easy way:)
Another one is a bit complicated for me now.

doujiang24:
@agentzh I commited a patch, and I hope haven't misunderstand what you mean:)

agentzh:
@doujiang24 Looking perfect to me now. Thanks for the quick fixes!

agentzh:
@doujiang24 Applied a slightly modified version of your patch to master. Thanks again!

PR #605, opened 2015-11-23
fixed a bug that nginx crashed when processing requests with no args

cfsego:
If the length of r->variables is greater than 250, it is alloced by
ngx_palloc_large(), in specific, by malloc. And if r->uri.len equals to 0,
ngx_http_lua_ngx_req_get_uri_args() allocates a zero-length buffer in a
pool. We found that when we use jemalloc or tcmalloc, these two pointers
are probably the same. Therefore, free this buffer will destroy r->variables
by chance, and nginx crashes.

### DIFF
--- a/src/ngx_http_lua_args.c
+++ b/src/ngx_http_lua_args.c
@@ -113,6 +113,11 @@ ngx_http_lua_ngx_req_get_uri_args(lua_State *L)
 
     lua_createtable(L, 0, 4);
 
+    if (r->args.len == 0) {
+        lua_createtable(L, 0, 0);
+        return 1;
+    }
+
     /* we copy r->args over to buf to simplify
      * unescaping query arg keys and values */
 
@@ -196,6 +201,11 @@ ngx_http_lua_ngx_req_get_post_args(lua_State *L)
     }
 
     dd("post body length: %d", (int) len);
+
+    if (len == 0) {
+        lua_createtable(L, 0, 0);
+        return 1;
+    }
 
     buf = ngx_palloc(r->pool, len);
     if (buf == NULL) {


agentzh:
@cfsego Hmm, it seems like a bug in jemalloc or tcmalloc? Freeing a pointer returned by malloc() should never crash I think.

agentzh:
@cfsego But I agree, allocating a zero-length buffer is wasting. Will you change your patch to avoid the allocation in the first place?

cfsego:
>> @cfsego Hmm, it seems like a bug in jemalloc or tcmalloc? Freeing a pointer returned by malloc() should never crash I think.

It is not a bug in jemalloc or tcmalloc, but I think it can be a bug in nginx because nginx allows to allocate zero-length buffer and gives out a malformed pointer, by chance equaling to another pointer. 

>> @cfsego But I agree, allocating a zero-length buffer is wasting. Will you change your patch to avoid the allocation in the first place?

I'm sorry I can not now because I do not know how Lua works.

agentzh:
@cfsego You don't need to know how Lua works for that. Just avoid calling `ngx_palloc` when `len == 0`. Right now you simply avoid calling `ngx_pfree` when `len == 0` which looks hacky.

cfsego:
done

src/ngx_http_lua_args.c
```diff
@@ -197,23 +201,27 @@ ngx_http_lua_ngx_req_get_post_args(lua_State *L)
...
-    if (buf == NULL) {
-        return luaL_error(L, "no memory");
-    }
+    retval = 0;
 
-    p = buf;
-    for (cl = r->request_body->bufs; cl; cl = cl->next) {
-        p = ngx_copy(p, cl->buf->pos, cl->buf->last - cl->buf->pos);
-    }
+    if (len) {
```
agentzh: Hmm, maybe checking the condition `len == 0` first is cleaner?

src/ngx_http_lua_args.c
```diff
@@ -197,23 +201,27 @@ ngx_http_lua_ngx_req_get_post_args(lua_State *L)
 
     dd("post body length: %d", (int) len);
 
-    buf = ngx_palloc(r->pool, len);
-    if (buf == NULL) {
-        return luaL_error(L, "no memory");
-    }
+    retval = 0;
```
agentzh: It looks wrong to return no values when `len == 0` since we used to return an empty Lua table in this case.

cfsego:
then still return 1?

cfsego:
I think you can give me a perfered lua document, so I will understand how it works

agentzh:
@cfsego I think the following snippet should work:

```
if (len == 0) {
    lua_createtable(L, 0, 0);
    return 1;
}
```

See the following document for the Lua C API usage:

http://www.lua.org/manual/5.1/manual.html

agentzh:
@cfsego BTW, will you please add a test case to the existing test suite that can reproduce the problem?

cfsego:
done.
and I looked through the test cases, and it covered the case with no args or posts. so I didn't add any test case.
I post two images up to describe the problem occurred in our system .
malloc
palloc
the former image shows nginx allocated a block by malloc(), and return a pointer. the pointer was used by r->variables.
and we added some debug info in ngx_palloc(), so it showed that nginx allocated a block by ngx_palloc(), in ngx.req.get_uri_args(), and return the same pointer. the latter image shows this.

agentzh:
@cfsego Is it possible to construct a test case that can reproduce the crash as you demonstrated by the debugging logs? The existing no-arg test cases do not demonstrate any memory issues nor crashes after all. I hope your changes can be really covered by the test suite. Is that possible?

agentzh:
Patch just applied to master. Thanks.

Issue #603, opened 2015-11-16
ngx.location.capture and sendfile


cmaion:
Hi,

Got bitten by workers crashes in LUA due to ngx.location.capture requesting a location that has the ability to sendfile large files.

Found out that if subrequests makes use of the X-Accel-Redirect, the actual file content gets fully loaded within the subrequest, and then sent back to LUA at once, as a single large chunk of memory... and potentially blowing stuff in the LUA VM.

This is actually understandable, but a quick note in the doc might help as it's easy to overlook this!

My fix in my case is to:
1. push a different sendfile header from the backend (ie, something different than X-Accel-Redirect) in order to prevent the ngx.location.capture subrequest from loading the file,
2. test this header when returning from ngx.location.capture() and,
3. if found, use return ngx.exec(header value) to exit from LUA and stream the file back to the client

Not sure if a better/nicer way exists though!

Cedric

agentzh:
@cmaion Crashing is not acceptable and must be fixed. Will you please provide a minimal and self-contained test case that can reproduce it easily on our side?

agentzh:
@cmaion And BTW, it's Lua, not LUA. Please see http://www.lua.org/about.html#name

cmaion:
OK ;)

The crashes left the following traces in the logs:

```
nginx: lua atpanic: Lua VM crashed, reason: string length overflow
*** stack smashing detected ***: nginx: worker process terminated
2015/11/16 10:53:17 [alert] 9556#0: worker process 10195 exited on signal 6 (core dumped)
```

or

```
nginx: lua atpanic: Lua VM crashed, reason: string length overflow
*** longjmp causes uninitialized stack frame ***: nginx: worker process terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7338f)[0x7f5b56e7538f]
[...]
```

I ran gdb on one of the core file I got: it confirmed that it was related to a huge body being sent back from the subrequest (due to a subrequest serving a file with sendfile which got fully buffered within the subrequest):

```
Core was generated by `nginx: worker process'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f6c3eab8cc9 in raise () from ./lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007f6c3eab8cc9 in raise () from ./lib/x86_64-linux-gnu/libc.so.6
[...]

PR #607, opened 2015-11-23
Note concerning ngx.location.capture buffering all subrequest c…

cmaion:
https://github.com/openresty/lua-nginx-module/issues/603

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3120,6 +3120,7 @@ Nginx's subrequests provide a powerful way to make non-blocking internal request
 `ngx_postgres`, `ngx_drizzle`, and even ngx_lua itself and etc etc etc.
 
 Also note that subrequests just mimic the HTTP interface but there is *no* extra HTTP/TCP traffic *nor* IPC involved. Everything works internally, efficiently, on the C level.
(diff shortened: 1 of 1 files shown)


README.markdown
```diff
@@ -3120,6 +3120,7 @@ Nginx's subrequests provide a powerful way to make non-blocking internal request
 `ngx_postgres`, `ngx_drizzle`, and even ngx_lua itself and etc etc etc.
 
 Also note that subrequests just mimic the HTTP interface but there is *no* extra HTTP/TCP traffic *nor* IPC involved. Everything works internally, efficiently, on the C level.
+This also mean that subrequests are fully performed and buffered before being returned to Lua: making a subrequest call to a location that can return or stream a large amount of data (including by using sendfile) is probably a bad idea.
```
agentzh: Hmm, I'm afraid there is no causal relationship here. The real reason behind the always-fully-buffering behaviour is that this API just does not support streaming.

agentzh:
@cmaion Just committed a different patch to git master to address this concern. Thanks for the patch all the same.

cmaion:
Thanks @agentzh!
At first I wasn't sure about how to rewrite this in a way that is both correct and would suit you :) and then I simply forgot about it.
Thanks again!

PR #618, opened 2015-12-08
cache: added functions to load code without having an ngx_http_request_t

majek:
This patch adds the functions ngx_http_lua_cache_loadbuffer2() and
ngx_http_lua_cache_loadfile2() that take an ngx_log_t argument directly
instead of an ngx_http_request_t. These can be used in contexts where an
HTTP request is not available.

The existing ngx_http_lua_cache_loadbuffer() and ngx_http_lua_cache_loadfile()
functions have been updated to wrap the new functions (in theory the new
functions could simply replace the old ones, but all the callers would
need to be updated as well, and that's quite a big diff).

### DIFF
--- a/src/ngx_http_lua_cache.c
+++ b/src/ngx_http_lua_cache.c
@@ -32,7 +32,7 @@
  *
  * */
 static ngx_int_t
-ngx_http_lua_cache_load_code(ngx_http_request_t *r, lua_State *L,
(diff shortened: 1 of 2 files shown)


agentzh:
I don't think it's worth introducing the xxx2 version of these functions (and yeah, the names are ugly). We could just make the existing version accept `ngx_log_t` directly instead of `ngx_http_request_t` (it seems that these C functions do not really need a full request object anyway).

PR #623, opened 2015-12-13
add body_filter_by_lua* in ngx_http_lua_context_name

doujiang24:
I'm adding test case in semaphore PR. If this is ok without test case?


agentzh:
@doujiang24 Nice catch. Will you add a test case too? Thanks!

doujiang24:
@agentzh I have added a test case :)

agentzh:
@doujiang24 Patch applied. Many thanks!

PR #624, opened 2015-12-14
cache: make load functions take a log directly

ghedo:
This patch makes the functions ngx_http_lua_cache_loadbuffer() and
ngx_http_lua_cache_loadfile() take an ngx_log_t argument directly
instead of an ngx_http_request_t so that they can be used in contexts
where an HTTP request is not available.

---

This supersedes #618 which can probably be closed now (/cc @majek).


agentzh:
@ghedo Thanks for the patch. Just merged.

PR #625, opened 2015-12-16
feature: allow tcpsock:setkeepalive() to receive nil args

thibaultCha:
`tcpsock:setkeepalive()` arguments are documented as being optional, yet they cannot be `nil`. This patch proposes to check if those arguments are `nil` before checking their type. If they are nil, then the configuration values are used as one would expect.

It comes handy when doing things like `sock:setkeepalive(var1, var2)` and if `var1` and `var2` are actually `nil`. It also allows one to do `sock:setkeepalive(nil, 25)`, though I hope this particular usage does not break the API.

It was motivated by an edge-case I encountered [here](https://github.com/thibaultCha/lua-cassandra/commit/759289d7404ff47a24779ee1c554861648e05adb#diff-ec126ebe12c61abf0bec8b3f0d0a3180R365) with this [travis failure](https://travis-ci.org/thibaultCha/lua-cassandra/jobs/96895510), in a scenario where a user of a lua-resty module can configure those options for the sockets used by that module.

---

Side note:

I would like to say I ran the tests but sadly I did not.
- I first followed the ngx_lua development instructions (I wanted to run all the tests with the full mysql/redis/memcached environment) but it seems [util/build2.sh](https://github.com/openresty/lua-nginx-module/blob/master/util/build2.sh) is outdated (some modules apparently got renamed, some paths are hard-coded, the Nginx version looks somewhat old).
- Fair enough, I decided to update it and slightly tweak it for my system, but so far drizzle has prevented me from finishing this. I have it compiled but `drizzle-nginx-module` does not seem to like any `LIBDRIZZLE_LIB` I am giving to it (granted this is totally outside the scope of this PR).
- I decided to simply download the latest OpenResty bundle and only run the `068-socket-keepalive.t` tests against it, without much success since the tests are failing, mainly due to the response size being checked:
```
# -received response of 156 bytes
# +received response of 160 bytes
```
I decided to ignore them (assumed it was a platform issue) and did not investigate. 

- I thought I'd make my changes to ngx_lua and compiled the bundle directly with it:
```
./configure --without-http_lua_module --add-module=../lua-nginx-module
```
And I added my test with an `--- ONLY` flag. However, the Nginx process cannot start when I compile OpenResty this way (with or without my changes):
```
t/068-socket-keepalive.t .. # I found ONLY: maybe you're debugging?
nginx: [error] failed to initialize Lua VM in /path/to/openresty-dev/lua-nginx-module/t/servroot/conf/nginx.conf:103
Bailout called.  Further testing stopped:  TEST 10b: sock:keepalive_timeout(nil) - Cannot start nginx using command "nginx -p /path/to/openresty-dev/lua-nginx-module/t/servroot/ -c /path/to/openresty-dev/lua-nginx-module/t/servroot/conf/nginx.conf > /dev/null".
```

Is there any recommended development installation (maybe more recent than `build2.sh`?) or have I done something wrong along the way? (I am on OS X btw). I am fine with using a VM if you think it would be easier. **Sorry for the long note!**

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -4464,7 +4464,7 @@ ngx_http_lua_socket_tcp_setkeepalive(lua_State *L)
     if (spool == NULL) {
         /* create a new socket pool for the current peer key */
 
-        if (n == 3) {
+        if (n == 3 && !lua_isnil(L, 3)) {
             pool_size = luaL_checkinteger(L, 3);
 
         } else {
@@ -4567,7 +4567,7 @@ ngx_http_lua_socket_tcp_setkeepalive(lua_State *L)
         ngx_del_timer(c->write);
     }
 
-    if (n >= 2) {
+    if (n >= 2 && !lua_isnil(L, 2) ) {
         timeout = (ngx_msec_t) luaL_checkinteger(L, 2);
 
     } else {
--- a/t/068-socket-keepalive.t
+++ b/t/068-socket-keepalive.t
@@ -776,7 +776,84 @@ qr/lua tcp socket connection pool size: 30\b/]
 
 
 
-=== TEST 11: sanity (uds)
+=== TEST 11: lua_socket_keepalive_timeout nil args
+--- config
+   server_tokens off;
+   location /t {
+       keepalive_timeout 60s;
+       lua_socket_keepalive_timeout 100ms;
+
+        set $port $TEST_NGINX_SERVER_PORT;
+        content_by_lua '
+            local port = ngx.var.port
+
+            local sock = ngx.socket.tcp()
+
+            local ok, err = sock:connect("127.0.0.1", port)
+            if not ok then
+                ngx.say("failed to connect: ", err)
+                return
+            end
+
+            ngx.say("connected: ", ok)
+
+            local req = "GET /foo HTTP/1.1\\r\\nHost: localhost\\r\\nConnection: keepalive\\r\\n\\r\\n"
+
+            local bytes, err = sock:send(req)
+            if not bytes then
+                ngx.say("failed to send request: ", err)
+                return
+            end
+
+            ngx.say("request sent: ", bytes)
+
+            local reader = sock:receiveuntil("\\r\\n0\\r\\n\\r\\n")
+            local data, res = reader()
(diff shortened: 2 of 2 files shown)


agentzh:
@thibaultCha Just a quick note: I'm still using `util/build.sh` for everyday development myself :)

agentzh:
@thibaultCha I'm going to change the `setkeepalive` method myself actually. The connection pool size will be specified in the options table of the `connect()` method, thus obsoleting the current 2nd parameter of the `setkeepalive` method. This change will allow upstream concurrency control (queueing) based on the connection pool size.

Having said that, allowing nil values for the arguments do make sense.

agentzh:
@thibaultCha The "failed to initialize Lua VM" error might be that your nginx is just linking against the wrong lua.

agentzh:
@thibaultCha Though the development and testing toolchain should mostly work on Mac OS X (I'm also working on a macbook everyday), it's recommended to use a Linux VM for the OpenResty development (and I'm running a Linux VM in VMWare atop Mac OS X).

thibaultCha:
> The connection pool size will be specified in the options table of the connect() method, thus obsoleting the current 2nd parameter of the setkeepalive method

Alright! This can be closed then I guess, in the meantime I'll just use a workaround :)

---

> I'm running a Linux VM in VMWare atop Mac OS X

Good to know! I will try to setup one too.

> I just prefer a different directory name for some of the dependency modules.

Fair enough! Though the build2.sh script also uses a path to your home directory apparently ([here](https://github.com/openresty/lua-nginx-module/blob/master/util/build2.sh#L42), and the default Nginx version seemed much older than the `1.7.10/1.9.3` range currently on qa.openresty,org, hence why I thought it was outdated. My understanding is that this is more of an example to follow rather than a "drop-in" build script, so no big deal.

agentzh:
@thibaultCha Oh, I haven't updated the default nginx version there for years since I always explicitly specify a version when invoking it, as in `./util/build2.sh 1.9.7`.

agentzh:
@thibaultCha And yes, the test cluster behind qa.openresty.org invokes the same `./util/build2.sh` script in the same way :)

agentzh:
Patch applied to git master. Thanks for your contribution!

PR #632, opened 2015-12-24
shdict check exptime param

membphis:
throw a lua error when the `exptime` is less than 0 .


agentzh:
Looking good to me though we also need a PR for lua-resty-core otherwise your newly-added test case will fail with lua-resty-core :)

membphis:
submit an PR to lua-resty-core right now, pls check

agentzh:
@membphis I've just merged a modified version of your patch to git master. Also updated lua-resty-core accordingly. Thanks!

PR #634, opened 2015-12-25
description about ngx.worker.id

RocFang:
Hello, I guess in ngx.worker.id(), it shoud be "init_worker_by" rather than "init_by"?

Thanks.


agentzh:
@RocFang Nice catch. Thanks!

PR #635, opened 2015-12-27
feature: implemented the balancer_by_lua_block and balancer_by_lua_file directives to allow NGINX load balancers written in Lua

agentzh:
This PR is coupled by the following PR for lua-resty-core:

https://github.com/openresty/lua-resty-core/pull/8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,6 +154,7 @@ src/timer.[ch]
 src/config.[ch]
 src/worker.[ch]
 src/lex.[ch]
+src/balancer.[ch]
 *.plist
 lua
 ttimer
--- a/README.markdown
+++ b/README.markdown
@@ -1028,6 +1028,8 @@ Directives
 * [log_by_lua](#log_by_lua)
 * [log_by_lua_block](#log_by_lua_block)
 * [log_by_lua_file](#log_by_lua_file)
+* [balancer_by_lua_block](#balancer_by_lua_block)
+* [balancer_by_lua_file](#balancer_by_lua_file)
 * [lua_need_request_body](#lua_need_request_body)
 * [lua_shared_dict](#lua_shared_dict)
 * [lua_socket_connect_timeout](#lua_socket_connect_timeout)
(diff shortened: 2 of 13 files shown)


agentzh:
@doujiang24 @lordnynex @membphis @bungle Please have a look if you are interested :) Thanks!

lordnynex:
:+1: 

Just a few questions 
- Is the cosocket API available in this context? The docs and tests don't show this. If so, is it worth adding a test demonstrating this? 
- Is there any way to return an upstream address for the proxy module to continue proxying as normal? A use case for this is a zookeeper lua module (I think @bungle is interested in this feature as well). I am currently working with Apache Mesos and Marathon. Their HA setup uses zookeeper, so if a stand-by master is promoted, the upstream block is broken because the old master issues a 302 redirect. Many projects supporting HA features use zookeeper like this. It would be great to be able to return() an upstream IP and continue processing as normal. 
- Are there any lua API's that are unavailable in this context? The documentation in this PR doesn't mention it. 
- Can this functionality co-exist with lua-upstream-nginx-module? A possible use case for this is maintaining the upstream pool based on a lua healthcheck by adding/removing servers periodically.
- Does lua-upstream-nginx-module get_servers() see this lua block?

lordnynex:
I also notice that this PR doesn't support the old style `balancer_by_lua '';` syntax. I think the block syntax is much better but it is inconsistent with the other directives. Is there a plan to eventually drop support for the old style?

agentzh:
Hi Brandon

On Sun, Dec 27, 2015 at 12:23 PM, Brandon Github
<notifications@github.com> wrote:
> Is the cosocket API available in this context?

No.

> The docs and tests don't show this.

I added some docs on this earlier today:

+This Lua code execution context does not support yielding, so Lua
APIs that may yield
+(like cosockets and "light threads") are disabled in this context.
One can usually work
+around this limitation by doing such operations in an earlier phase
handler (like
+[access_by_lua*](#access_by_lua)) and passing along the result into
this context
+via the [ngx.ctx](#ngxctx) table.

> If so, is it worth adding a test demonstrating this?

Just added two tests for it and it indeed found an issue that I have
quickly fixed. Thank you :)

> Is there any way to return an upstream address for the proxy module to continue proxying as normal?

Yes, via the set_current_peer() function of the ngx.upstream module in
the lua-resty-core library. See the following PR:

https://github.com/openresty/lua-resty-core/pull/8/files

> A use case for this is a zookeeper lua module (I think @bungle is interested in this feature as well).
> I am currently working with Apache Mesos and Marathon. Their HA setup uses zookeeper, so if a
> stand-by master is promoted, the upstream block is broken because the old master issues a 302 redirect.
> Many projects supporting HA features use zookeeper like this. It would be great to be able to return() an
> upstream IP and continue processing as normal.

Please let me know if the Lua API provided by the ngx.balancer module
in lua-resty-core (mentioned above) works for you.

> Are there any lua API's that are unavailable in this context?

All of the Lua APIs that may yield are disabled in this context. The
docs updated the "context" list for those tested working Lua APIs.

> Can this functionality co-exist with lua-upstream-nginx-module? A possible use case for this is maintaining the upstream pool based
> on a lua healthcheck by adding/removing servers periodically.

Nope. lua-upstream-nginx-module only works with static server
configurations in the `upstream {}` blocks. We need a different
healthchecker
to work with highly dynamic Lua-based balancers used in balancer_by_lua*.

> Does lua-upstream-nginx-module get_servers() see this lua block?
>

It can't because balancer_by_lua* can define dynamic peers to use
which can vary on the request level.

> I also notice that this PR doesn't support the old style balancer_by_lua ''; syntax. I think the block syntax is
> much better but it is inconsistent with the other directives. Is there a plan to eventually drop support for the
> old style?

This is intentional. I want to discourage the use of the old *_by_lua
"..." form so newly added *_by_lua* directives only support the block
and file forms. But I don't want to remove existing *_by_lua "..."
directives either, for backward compatibility. Alas. We always have to
compromise things.

Thanks for your review!

Best regards,
-agentzh

bungle:
@agentzh,

I added a few locals to balancer.lua:
https://github.com/openresty/lua-resty-core/pull/9

I tried to run the tests, but I was not able to get them all run on my OS X El Capitan. I will check if I have some problems on my side, and report back.

Here is the summary so far (that could be a problem on my side):

```bash
Test Summary Report
-------------------
t/balancer.t (Wstat: 0 Tests: 126 Failed: 68)
  Failed tests:  1-3, 5, 7-9, 11, 13-15, 17, 19-21, 23, 41-44
                46-49, 51-53, 56, 58-60, 63, 65-78, 80-83
                109-126
  Parse errors: Bad plan.  You planned 112 tests but ran 126.
```

agentzh:
@bungle Thanks! I'll check your PR out.

Regarding the test failures on your Mac OS X system, will you please provide the first few failures' output? The test summary is not quite helpful for diagnostics :) Thanks!

agentzh:
@bungle Merged your patch. Thanks!

bungle:
Here is the test run log:
https://gist.github.com/bungle/3fbbd2021877c1350697

The error seems to be same on all of these, and that's why I was thinking that maybe the problem is on my side.

bungle:
If I change that 127.0.0.3 to 127.0.0.1 I get tests passed:

```
t/balancer.t .. 7/112 nginx: [warn] load balancing method redefined in /Users/bungle/Sources/lua-resty-core/t/servroot/conf/nginx.conf:33
t/balancer.t .. ok       
All tests successful.
Files=1, Tests=112,  8 wallclock secs ( 0.06 usr  0.02 sys +  0.64 cusr  0.46 csys =  1.18 CPU)
Result: PASS
```

bungle:
Here is a pull request to fix the balancer.t tests:
https://github.com/openresty/lua-resty-core/pull/11

agentzh:
@bungle Okay, 127.0.0.3:12345 simple drops your packets on your side instead of sending RST as on my side. It's understandable. Need to figure out a more portable way to test testing a different IP address than 127.0.0.1. I'm not satisfied with your openresty/lua-resty-core#11 pull request since I want to test an IP address other than 127.0.0.1 for better code coverage. I'd better change the tests to point to agentzh.org:12345 instead of 127.0.0.3:12345 since the former always drops any incoming packets and we can test if the proxy times out (with a very short timeout threshold) as expected. Will you contribute the change? Thanks!

BTW, I've already merged this PR's branch into master. So I'm closing this. Further pull requests should target the master branch instead :)

agentzh:
@bungle I tried to run these tests on Mac OS X today and found that we just need to add 127.0.0.3 to the loopback device by the following command:

```
sudo ifconfig lo0 alias 127.0.0.3 up
```

Then all the tests in balancer.t pass for me there :)

PR #636, opened 2015-12-29
conversion from 'int' to 'in_port_t', possible loss of data

itpp16:
src/ngx_http_lua_balancer.c(467) : warning C4244: '=' : conversion from 'int' to 'in_port_t', possible loss of data

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -409,7 +409,7 @@ ngx_http_lua_balancer_free_peer(ngx_peer_connection_t *pc, void *data,
 
 int
 ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
-    const u_char *addr, size_t addr_len, int port, char **err)
+    const u_char *addr, size_t addr_len, in_port_t port, char **err)
 {
     ngx_url_t              url;
     ngx_http_lua_ctx_t    *ctx;


src/ngx_http_lua_balancer.c
```diff
@@ -409,7 +409,7 @@ ngx_http_lua_balancer_free_peer(ngx_peer_connection_t *pc, void *data,
 
 int
 ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
-    const u_char *addr, size_t addr_len, int port, char **err)
+    const u_char *addr, size_t addr_len, in_port_t port, char **err)
```
agentzh: We don't want to use `in_port_t` as the parameter type here since this C function is part of the ABI where we want to stick with very basic C data types.  Let's do proper type coercions in the C function body instead.

agentzh:
@itpp16 Thank you for your report but I don't quite like your change. Will you please try the following patch instead?

```diff
diff --git a/src/ngx_http_lua_balancer.c b/src/ngx_http_lua_balancer.c
index f372a66..34a9849 100644
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -464,7 +464,7 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
     ngx_memcpy(url.url.data, addr, addr_len);

     url.url.len = addr_len;
-    url.default_port = port;
+    url.default_port = (in_port_t) port;
     url.uri_part = 0;
     url.no_resolve = 1;
```

itpp16:
Hello agentzh, this will work as well, I prefer proper type coercion's.

agentzh:
@itpp16 Cool. Just applied my patch to git master. Thanks.

PR #639, opened 2016-01-05
$TEST_NGINX_HTML_DIR is too long name for unix socket

maage:
From Linux unix(7): "On Linux sun_path is 108 bytes in size"

I have longer directory names when running different builds of nginx and
these tests fail because of that.

### DIFF
--- a/t/014-bugs.t
+++ b/t/014-bugs.t
@@ -978,7 +978,7 @@ ok
     proxy_ssl_session_reuse off;
 
     server {
-        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        listen unix:nginx.sock ssl;
         ssl_certificate ../html/test.crt;
         ssl_certificate_key ../html/test.key;
 
@@ -988,7 +988,7 @@ ok
     }
 
     upstream local {
-        server unix:$TEST_NGINX_HTML_DIR/nginx.sock;
+        server unix:nginx.sock;
     }
 
 --- config
--- a/t/023-rewrite/socket-keepalive.t
+++ b/t/023-rewrite/socket-keepalive.t
@@ -817,7 +817,7 @@ lua tcp socket keepalive timeout: unlimited
 "
     lua_package_path '$::HtmlDir/?.lua;./?.lua';
     server {
-        listen unix:$::HtmlDir/nginx.sock;
+        listen unix:nginx.sock;
         default_type 'text/plain';
 
         server_tokens off;
@@ -832,7 +832,7 @@ lua tcp socket keepalive timeout: unlimited
         set $port $TEST_NGINX_MEMCACHED_PORT;
         rewrite_by_lua '
             local test = require "test"
-            local path = "$TEST_NGINX_HTML_DIR/nginx.sock";
+            local path = "nginx.sock";
             local port = ngx.var.port
             test.go(path, port)
             test.go(path, port)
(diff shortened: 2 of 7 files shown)


agentzh:
@maage Unfortunately your patch seems to introduce new test failures on my side:

```
t/023-rewrite/socket-keepalive.t .. 109/138 nginx: [emerg] bind() to unix:../html/nginx.sock failed (2: No such file or directory)
Bailout called.  Further testing stopped:  TEST 11: sanity (uds) - Cannot start nginx using command "nginx -p /home/agentzh/git/lua-nginx-module/t/servroot/ -c /home/agentzh/git/lua-nginx-module/t/servroot/conf/nginx.conf > /dev/null" (status code 256).
FAILED--Further testing stopped: TEST 11: sanity (uds) - Cannot start nginx using command "nginx -p /home/agentzh/git/lua-nginx-module/t/servroot/ -c /home/agentzh/git/lua-nginx-module/t/servroot/conf/nginx.conf > /dev/null" (status code 256).
```

Will you fix that? Thanks!

maage:
Okay, I need to check why it fails, for me it seems to be okay.

t/023-rewrite/socket-keepalive.t .... ok

I think this is because you have set TEST_NGINX_HTML_DIR. And if it is not relative to current directory when running nginx then my patch obviously fails. I'll see if I can fix this.

agentzh:
@maage I've checked and found that I didn't explicitly set the TEST_NGINX_HTML_DIR environment myself.

maage:
Hmm, I wonder why. Anyways I've updated my patch. This time without explicit path at all, so it uses current (config) directory as storage of socket file.

I've applied same change to t/139-ssl-cert-by.t, but could not prove that test passes as my openssl is too low.

agentzh:
@maage It's fine now. Thanks!

agentzh:
Just merged to master.

agentzh:
Hmm, sorry I've made a mistake here. Creating the .sock file in the current working directory is a bad idea since once the .sock file is left there no one will ever clean it up and during development I got errors like this:

```
t/139-ssl-cert-by.t .. # I found ONLY: maybe you're debugging?
nginx: [emerg] bind() to unix:nginx.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:nginx.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:nginx.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:nginx.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:nginx.sock failed (98: Address already in use)
nginx: [emerg] still could not bind()
Bailout called.  Further testing stopped:  TEST 12: connection aborted prematurely - Cannot start nginx using command "nginx -p /home/agentzh/git/lua-nginx-module/t/servroot/ -c /home/agentzh/git/lua-nginx-module/t/servroot/conf/nginx.conf > /dev/null" (status code 256).
FAILED--Further testing stopped: TEST 12: connection aborted prematurely - Cannot start nginx using command "nginx -p /home/agentzh/git/lua-nginx-module/t/servroot/ -c /home/agentzh/git/lua-nginx-module/t/servroot/conf/nginx.conf > /dev/null" (status code 256).
```

This why I used the html dir for the socket file in the first place. 

Sorry, I have to revert your commit for now until we come up with a better solution.

thibaultcha:
Closing this PR for now since I have not heard a similar complaint from anyone else. This only affects maintainers and contributors anyway. I have encountered the issue myself and simply ensured the path were shorter for now, as a workaround. I think we can fix it given we rework the patch, but it doesn't seem like a priority to me. Thanks!

PR #640, opened 2016-01-05
lua-nginx-module\src\ngx_http_lua_ssl_certby.c(274) : warning C4702: …

itpp16:
…unreachable code

lua-nginx-module\src\ngx_http_lua_ssl_certby.c(274) : warning C4702: unreachable code


agentzh:
@itpp16 Nice catch. Just applied to master. Thanks!

PR #644, opened 2016-01-07
skip t/139-ssl-cert-by.t if too old OpenSSL

maage:
At RHEL-7 default openssl version is

OpenSSL 1.0.1e-fips 11 Feb 2013

and certby tests fails because of that. Because test bails out, testing stops there and I can't even get report.

This patch is only tested with RHEL-7 so it might will not work with other openssl installations.


agentzh:
Merged. Thanks!

PR #645, opened 2016-01-07
fix 132-lua-blocks error_log rx

maage:
line is in my env:

nginx: [emerg] Lua code block missing the closing long bracket "]]" in /build/nginx/nginx-1.9.9/ngx_lua/t/servroot/conf/nginx.conf:49

so '\] .*? Lua' does not match single space and line number is off

### DIFF
--- a/t/132-lua-blocks.t
+++ b/t/132-lua-blocks.t
@@ -310,7 +310,7 @@ hello, world
 [error]
 --- must_die
 --- error_log eval
-qr/\[emerg\] .*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:41/
+qr/\[emerg\].*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:41/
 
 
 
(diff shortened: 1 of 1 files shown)


t/132-lua-blocks.t
```diff
@@ -310,7 +310,7 @@ hello, world
 [error]
 --- must_die
 --- error_log eval
-qr/\[emerg\] .*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:41/
+qr/\[emerg\].*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:\d+/
```
agentzh: Hmm, actually I want to test the exact line number in the error message. You change makes the test cases "weaker". Maybe we should work on making the line number always predictable instead? Are you using the latest Test::Nginx from git master?

maage:
I use latest Test::Nginx from git master. I have several changes, but I was able to keep line numbers intact by moving my changes to end of config file. You might want to do that also with later changes. I've updated patch.

agentzh:
@maage Yeah, sure. Thanks!

t/132-lua-blocks.t
```diff
@@ -310,7 +310,7 @@ hello, world
 [error]
 --- must_die
 --- error_log eval
-qr/\[emerg\] .*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:41/
+qr/\[emerg\].*? Lua code block missing the closing long bracket "]]" in .*?\bnginx\.conf:41/
```
agentzh: What's the point of this change?

agentzh:
Consider it resolved.

PR #647, opened 2016-01-15
ngx.re APIs now returns false instead of nil

jdesgats:
Previously, `ngx.re.match` and `ngx.re.find` returned nil when the
subjcet did not match the pattern. This caused issues when results are
stored into a table (holes, wrong length, ...).

This commit fixes this behavior:

* `ngx.re.match` returns `false`
* `ngx.re.find` returns `false, false`

The behavior in case of actual errors is not changed (`nil` is
returned).

About `ngx.re.find`, I noticed that when the subject didn't match the pattern, a single nil was returned and when  it did match but the requested group didn't, two `nil`s were returned. Not sure if it was on purpose or not, but now it returns two `false` in both cases.


agentzh:
@jdesgats Thanks for the patch. But I think we just need to change `ngx.re.match` here since `ngx.re.find` never returns a Lua table anyway :)

jdesgats:
Patch has been heavily reworked, better to open a new PR.

PR #648, opened 2016-01-16
bugfix: support non-RSA private keys as well

ghedo:


agentzh:
@ghedo Thanks for the patch. But will you please add some tests to the existing test suite for this change? Very much appreciated!

ghedo:
So, I added tests for the SSL cert FFI functions (using the test.crt RSA cert) in the first commit, and then added ECDSA tests in the second commit along with the fix.

However, while the ECDSA cert and key are loaded fine (after the fix), I get "self-signed certificate" errors even when setting `lua_ssl_trusted_certificate`. Could you please have a look? Maybe I missed something.

agentzh:
@ghedo Maybe you didn't adjust the `lua_ssl_verity_depth` directive on the client side?

ghedo:
Nope, still getting "self-signed certificate", which isn't wrong since the cert *is* self-signed, but it should also be trusted...

agentzh:
@ghedo Aye. We already have self-signed SSL certificates in the existing test suite.

ghedo:
Maybe there's something wrong with the ECDSA cert, but even `openssl s_client` works fine when setting the option `-CAfile` to the cert itself.

agentzh:
@ghedo Maybe there is a typo or something in the certificate file names referenced in your tests?

ghedo:
Seems fine to me. Otherwise it would say that it couldn't load the files or something like that.

ghedo:
But I bet it's some really dumb mistake like that...

lziest:
Don't know what's wrong, but I would try to use 'ssl_certificate' directly with ecdsa cert directly to see if cert_by_lua is actually related.

ghedo:
Ok, I think I resolved the problem. Basically I needed to call `ngx_http_lua_ffi_ssl_clear_certs()` in the second test (I didn't realize it was mandatory, but only in the second block?).

agentzh:
@ghedo Hmm, interesting. Why is that? That's not reasonable. Is it an OpenSSL issue?

ghedo:
Yeah, I wouldn't be surprised. I haven't checked in detail, but it seems `SSL_use_certificate()` has a different behavior depending on the cert type...

agentzh:
@ghedo Alas.

ghedo:
@agentzh do you need anything else from me to get this merged?

agentzh:
@ghedo Will you please confirm that the necessity of clearing the certs are specific to certificate types?

ghedo:
It seems I can only reproduce this behaviour with nginx+openresty (I tried making a stand-alone OpenSSL application, but it worked fine...), so it may very well be something to do with nginx or ngx-lua.

I need to investigate this more deeply, but feel free to look into this as well if you have some time :)

agentzh:
@ghedo Thanks for looking into this. I would be surprised that it has something to do with nginx or ngx_lua since both of them just call OpenSSL API exclusively to do the certificate and private key handling.

ghedo:
Looked into this a bit more and I think I found what the problem is. Basically OpenSSL picks a certificate to return to the user based on the supported cipher suites. So when you call `SSL_use_certificate()`, if a certificate with the same type was previously added (in this case RSA), that certificate is overwriten, but if the type is different (ECDSA) OpenSSL keeps both.

I confirmed this by changing the cipher suite list used by`ngx_http_lua_socket_tcp_sslhandshake()`: first I tried "RSA,ECDSA" and of course the RSA certificate was returned, then I tried "ECDSA.RSA" and the ECDSA certificate was returned instead (meaning that there was no need to clear certs).

So basically not a bug (either OpenSSL or ngx-lua), just somewhat confusing behaviour. I guess we should leave the clear_certs calls in the test suite though (like it is now).

ghedo:
Oh, btw. I also tried to put another ECDSA cert in the `ssl_certificate` config option and that worked too.

agentzh:
@ghedo Thanks for digging it up! Very much appreciated! I'll try merging this soon.

lziest:
@ghedo thanks, that actually explains something I observed before as well, that it is actually an array of X509 objects inside the ssl ctx.

agentzh:
@ghedo Patch applied to master. Thanks a lot!

PR #650, opened 2016-01-18
Fixes for SSL error reporting

ghedo:


agentzh:
@ghedo I don't think we should pass `""` to `*err` since it's required to only read `*err` when the C functions return an error status code.

ghedo:
Yeah, but avoiding a segfault (even if it's the user's fault) seems like a good thing to do anyway, also considering that the fix is so simple.

agentzh:
@ghedo Nope, unlike Lua, this is the C land so there's no hand-holding. Setting `NULL` is also easier to catch misuses on the caller's side (segfaults) and it's also more consistent with other FFI C APIs defined in ngx_lua.

ghedo:
Fair enough. The other two commits are valid (and separate) fixes Ithink, should I just remove the err initialization commit from this PR?

agentzh:
@ghedo Yes, please. Thank you!

ghedo:
Ok, done.

agentzh:
@ghedo Just merged to master. Thanks!

PR #651, opened 2016-01-18
test: add test for request uri not changing during subrequest

Sirupsen:
@agentzh `ngx.var.request_uri` does not change within the context of a subrequest when accessed from Lua, but `ngx.var.uri` and `ngx.var.args` do, and `ngx.var.request_uri` form the `echo` module—which causes test 77 to fail, but not 78, 79 and 80. This seems like a bug. The subrequest's `unparsed_uri` (which is what `$request_uri` references) is set to the parent's `unparsed_uri` [in ngx-lua here](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_subrequest.c#L1541).

### DIFF
--- a/t/020-subrequest.t
+++ b/t/020-subrequest.t
@@ -2873,3 +2873,94 @@ DELETE /file.txt, response status: 204
 --- no_error_log
 [error]
 --- error_code: 200
+
+
+
+=== TEST 77: ngx.var.request_uri + ngx.location.capture
+--- config
+location /b {
+    content_by_lua '
+        ngx.print(ngx.var.request_uri)
+    ';
+}
+
+location /a {
+    content_by_lua '
+        local res = ngx.location.capture("/b")
+        ngx.say(res.body)
+    ';
+}
+--- request
+GET /a
+--- response_body
+/b
+--- no_error_log
+[error]
+
+
+
+=== TEST 78: ngx.var.request_uri from echo + ngx.location.capture
+--- config
+location /b {
+    echo $echo_request_uri;
+}
+
+location /a {
+    content_by_lua '
+        local res = ngx.location.capture("/b")
+        ngx.print(res.body)
+    ';
+}
+--- request
+GET /a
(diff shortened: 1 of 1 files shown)


Sirupsen:
I can't come up with a workaround until this is in master, since something like:

```lua
local request_uri_workaround = ngx.var.uri
if not ngx.var.args then
  request_uri_workaround = request_uri_workaround .. "?" .. ngx.var.args
end
```

Doesn't work, since you could have a complex URI like: `/foo?a=b#omg`.

agentzh:
@Sirupsen No, it's not a bug because the NGINX builtin variable `$request_uri` is always the raw request URL sent from the HTTP clients. That is, NGINX's internal URL rewrites and subrequests should never affect its value.

Sirupsen:
Yeah, that sounds reasonable. For what I want, i.e. get what the `request_uri` represents in a subrequest, is the above approach reasonable or do you see any problems with it?

agentzh:
@Sirupsen Yes, using the "$uri?$args" is the correct way (I won't say it is a work-around since it is the standard practice, actually) :)

Sirupsen:
:-) How can we document this so the next person doesn't have to spend an hour or two debugging this?

agentzh:
@Sirupsen This is explained in detail in my nginx tutorials:

https://openresty.org/#eBooks

I'll ensure my upcoming book "Programming OpenResty" will have explanation for this as well (if it didn't, just yell at me ;))

agentzh:
@Sirupsen Actually I hope the official nginx documentation for `$request_uri` explicitly says this. Better contact the official NGINX team for this instead?

Sirupsen:
> full original request URI (with arguments)

http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_uri

Subtle.. but kind of there, it's not obvious at least.

Looking forward to your book Yichun! :-)

PR #652, opened 2016-01-19
feature: ngx.log accepts tables with __tostring metafield

thibaultCha:
> Extend `ngx.log()` to allow table arguments when those have a `__tostring`
metafield. If they do not, ngx.log will raise an error (similar to
previous behavior when a table was given to ngx.log).

> When a table with `__tostring` metafield is given, the metamethod will
be called through `luaL_callmeta()`.

#### Use case
While implementing [lua-cassandra](https://github.com/thibaultCha/lua-cassandra), I found myself in the need to return more complex errors than regular strings. Those errors allow one to take particular actions depending on the fields it contains (error code and a given Cassandra peer for example). For those errors to be nicely printed to the user, I implemented a `__tostring` metafield but quickly found out that I needed to do:

```lua
ngx.log(ngx.ERR, tostring(err))
```

This change was primarily motived by this use case. Secondly, it is worth noting that in plain Lua,

```lua
print(t)
```

Will call the `__tostring` metamethod of `t` if any, so this behavior mimics Lua's `print`.

#### Caveats
- Yes, tables are more expensive for Lua to maintain, but in some cases, they are more descriptive than plain strings. Errors are not the only use case in which you would want to log tables either. I think it's a good thing if we get closer to `print`'s behavior.
- What I am more concerned about: currently, the metamethod is called twice (once for the length, once for the actual string). Just like `lua_tolstring()` is called twice for strings too. If the `__tostring` is expensive, we are wasting some time here.

Finally, I did not check if ngx's `print` (overriding ngx.log at NOTICE level) will also benefit from this, nor added a test for it. If this change is considered, I will add such a test.

### DIFF
--- a/src/ngx_http_lua_log.c
+++ b/src/ngx_http_lua_log.c
@@ -160,6 +160,16 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
 
                 break;
 
+            case LUA_TTABLE:
+                if (!luaL_callmeta(L, i, "__tostring")) {
+                    return luaL_argerror(L, i, "expected table to have "
+                                         "__tostring metamethod");
+                }
+
+                lua_tolstring(L, -1, &len);
+                size += len;
+                break;
+
             case LUA_TLIGHTUSERDATA:
                 if (lua_touserdata(L, i) == NULL) {
                     size += sizeof("null") - 1;
@@ -225,6 +235,12 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
                     *p++ = 'e';
                 }
 
+                break;
+
+            case LUA_TTABLE:
+                luaL_callmeta(L, i, "__tostring");
+                q = (u_char *) lua_tolstring(L, -1, &len);
+                p = ngx_copy(p, q, len);
                 break;
 
             case LUA_TLIGHTUSERDATA:
--- a/t/009-log.t
+++ b/t/009-log.t
@@ -265,7 +265,96 @@ qr/\[error\] \S+: \S+ \[lua\] set_by_lua:2: truefalsenil,/
 
 
 
-=== TEST 14: print() in header filter
+=== TEST 14: test table with metamethod
+--- config
+    location /log {
+        content_by_lua_block {
+            ngx.say("before log")
+            local t = setmetatable({v = "value"}, {
+                __tostring = function(self)
+                    return "tostring "..self.v
+                end
+            })
+            ngx.log(ngx.ERR, t)
+            ngx.say("after log")
+        }
+    }
+--- request
+GET /log
+--- response_body
+before log
+after log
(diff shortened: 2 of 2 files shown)


src/ngx_http_lua_log.c
```diff
@@ -227,12 +237,17 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
...
+                q = (u_char *) lua_tolstring(L, -1, &len);
+                p = ngx_copy(p, q, len);
+                break;
+
             case LUA_TLIGHTUSERDATA:
                 *p++ = 'n';
                 *p++ = 'u';
                 *p++ = 'l';
                 *p++ = 'l';
-
```
thibaultCha: I figures this was the proper style, based on what I read elsewhere. I might have totally got that wrong though.

util/run_test.sh
```diff
@@ -7,4 +7,3 @@ $script_dir/reindex $testfile
 export PATH=$root/work/sbin:$PATH
 killall nginx
 prove -I$root/../test-nginx/lib $testfile
-
```
thibaultCha: My bad, will fix this

src/ngx_http_lua_log.c
agentzh: @thibaultCha This change is unrelated to this PR, better exclude it :) But I'm fine with including it anyway.

agentzh:
@thibaultCha Yeah, given that the error logging always uses a fixed size buffer, we don't gain much with pre-calculating the buffer size in the separate scan here anyway. We can address this in a future commit.

t/009-log.t
```diff
@@ -265,7 +265,69 @@ qr/\[error\] \S+: \S+ \[lua\] set_by_lua:2: truefalsenil,/
 
 
 
-=== TEST 14: print() in header filter
+=== TEST 14: test table with metamethod
+--- config
+    location /log {
+        set_by_lua $a '
```
agentzh: Can we just use `content_by_lua_block {}` here? I think it's distraction to use `set_by_lua '... return 32'` and `echo $a` in these test cases :)

thibaultCha:
I updated the patch (not sure if you prefer a new commit or not):

- Revert undesired changes (style issues). I too believe they do not belong to this PR.
- In tests: use `content_by_lua_block` instead of `set_by_lua`. I missed the fact that the initial tests use content_by_lua.
- Add a test for `print()` with such a table.

 A question though: do you which to update the tests to the `*by_block` directives? Because other tests there use `content_by_lua '';`.

agentzh:
@thibaultCha I prefer new test cases use the `_by_lua_block {}` form :)

agentzh:
@thibaultCha Applied your patch to git master. Thanks a lot for your contribution!

thibaultCha:
Great news! Thanks

PR #653, opened 2016-01-20
Unmatched capture are set to false in capture table

jdesgats:
Now unmatched capture groups in `ngx.re.match` and `ngx.re.gmatch` are
set to `false` rather than nil. This prevents to create table with
holes that don't play nice with `ipairs` or `#` operators.

For consistency, unmatched trailing captures (for instance in
`ngx.re.match("hello", "(hello)(.+)?")`) as well as named captures are
also set to `false`.

An interesting side effect of this patch concerns the duplicate pattern option (the `D` flag): the unmatched capture will also receive a `false` value, for instance:
```lua
local regex = '^(?<g>foo)?(?<g>bar)?$'
ngx.re.match('foobar', regex, 'D') -- { [0]='foobar', [1]='foo', [2]='bar', g={ 'foo', 'bar' } }
ngx.re.match('foo', regex, 'D')    -- { [0]='foobar', [1]='foo', [2]=false, g={ 'foo', false } }
ngx.re.match('bar', regex, 'D')    -- { [0]='foobar', [1]=false, [2]='bar', g={ false, 'bar' } }
ngx.re.match('', regex, 'D')       -- { [0]='foobar', [1]=false, [2]=false, g={ false, false } }
```
I have mixed feelings about this behavior: on one hand it is consistent with the other cases, but on the other hand I can't think of any valid uses case and it might create a lot of useless tables. So I am open to any feedback about that.

### DIFF
--- a/src/ngx_http_lua_regex.c
+++ b/src/ngx_http_lua_regex.c
@@ -2071,6 +2071,11 @@ ngx_http_lua_re_collect_named_captures(lua_State *L, int res_tb_idx,
         }
 
         if (flags & NGX_LUA_RE_MODE_DUPNAMES) {
+            /* unmatched groups are not stored in tables in DUPNAMES mode */
+            if (!lua_toboolean(L, -1)) {
+                lua_pop(L, 1);
+                continue;
+            }
 
             lua_getfield(L, -2, name); /* big_tb cap small_tb */
 
--- a/src/ngx_http_lua_regex.c
+++ b/src/ngx_http_lua_regex.c
@@ -592,14 +592,14 @@ exec:
     }
 
     if (res_tb_idx == 0) {
-        lua_createtable(L, rc /* narr */, 0 /* nrec */);
+        lua_createtable(L, re_comp.captures /* narr */, name_count /* nrec */);
         res_tb_idx = lua_gettop(L);
     }
 
-    for (i = 0, n = 0; i < rc; i++, n += 2) {
+    for (i = 0, n = 0; i < re_comp.captures + 1; i++, n += 2) {
         dd("capture %d: %d %d", i, cap[n], cap[n + 1]);
-        if (cap[n] < 0) {
-            lua_pushnil(L);
+        if (i >= rc || cap[n] < 0) {
+            lua_pushboolean(L, 0);
 
         } else {
             lua_pushlstring(L, (char *) &subj.data[cap[n]],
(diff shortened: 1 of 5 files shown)


agentzh:
@jdesgats Hmm, can we leave the duplicate groups alone? I agree it's seldom or never desired.

jdesgats:
Ok, the behavior has been reverted to previous one, only for duplicate groups:

```
local regex = '^(?<g>foo)?(?<g>bar)?$'
ngx.re.match('foobar', regex, 'D') -- { [0]='foobar', [1]='foo', [2]='bar', g={ 'foo', 'bar' } }
ngx.re.match('foo', regex, 'D')    -- { [0]='foo',    [1]='foo', [2]=false, g={ 'foo' } }
ngx.re.match('bar', regex, 'D')    -- { [0]='bar',    [1]=false, [2]='bar', g={ 'bar' } }
ngx.re.match('', regex, 'D')       -- { [0]='',       [1]=false, [2]=false }
```

agentzh:
@jdesgats Thanks for the quick updates! I'll look into it soon.

src/ngx_http_lua_regex.c
```diff
@@ -1116,12 +1116,12 @@ ngx_http_lua_ngx_re_gmatch_iterator(lua_State *L)
 
     dd("rc = %d", (int) rc);
 
-    lua_createtable(L, rc /* narr */, 0 /* nrec */);
+    lua_createtable(L, ctx->ncaptures + 1 /* narr */, name_count /* nrec */);
```
agentzh: Why `+ 1` here?

agentzh:
@jdesgats I think we also need to update the pure C API for lua-resty-core in lua-nginx-module too. You can run ngx_lua's tests using lua-resty-core by setting the following environment, for example,

```bash
export TEST_NGINX_INIT_BY_LUA="package.path = (package.path or '') .. '$PWD/../lua-resty-core/lib/?.lua;$PWD/../lua-resty-lrucache/lib/?.lua;' require 'resty.core'"
```

Then run the `prove` test harness as usual.

src/ngx_http_lua_regex.c
agentzh: Okay, in case that `ctx->ncaptures` is 0, we indeed need to plus one otherwise LuaJIT will resize (auto-grow) the table, which is expensive (for narr > 0, LuaJIT automatically adds one to the user-supplied value since LuaJIT stores the index 0 in the array part of tables too).

jdesgats:
Thank you for your feedback.

Yes, the changes have to be ported to `resty.core` too. All the lua-nginx-modules tests pass with core too (aI added some more test cases directly on core). Here is the relevant PR: https://github.com/openresty/lua-resty-core/pull/19

agentzh:
@jdesgats Applied a modified version of your patch to master, mainly doing the same change to the captures table fed into `ngx.re.sub` and `ngx.re.gsub`'s user function callbacks as well. Thanks for your contribution!

PR #660, opened 2016-01-24
the `ngx.req.get_body_data()` API is valid for log_by_lua phase

membphis:
the `ngx.req.get_body_data()` API is valid for log_by_lua phase


agentzh:
@membphis Patch applied in master. Thanks a lot!

PR #663, opened 2016-01-28
bugfix: code cache off does not work in balancer_by_lua_file

doujiang24:


agentzh:
@doujiang24 Merged a slightly modified version of your patch. Thanks a lot!

Issue #254, opened 2013-07-03
run log_by_lua before the access log


calind:
I think a better behavior of log_by_lua would be to run **before** access_log so that it can set some variables which can be written afterward by access_log.

For example one can do a json econde of data he wants to be logged and then let access log to write the log. This way you can simply have your data in logs already in json format and you don't need to to a translation from log format to json.

PR #667, opened 2016-02-03
doc: fix log_by_lua, runs before nginx access log and after

bjoe2k4:
According to #254 and commit e549fc2df0c03648a4553c8310c90c61db46aa42 log_by_lua runs before classic nginx access logs. This was missed in the docs.


agentzh:
@bjoe2k4 Nice catch. Thanks!

Issue #668, opened 2016-02-09
nginx 1.9.11 incompatibility


nzt4567:
Hi,

new nginx version removes ngx_udp_connection_t which is used in lua module -> it will not compile :'(

Thanks,
Tomas

teward:
I was about to poke here, too, about this issue.  Heh.

agentzh:
The compilation error should be gone now.

sumantdas:
seems the issue is coming again in the current master branch

agentzh:
@sumantdas Fixed. Thanks for the report.

BTW, why are you using such an old nginx core? We no longer do thorough testing against such old nginx cores.

BTW2: it's highly recommended to use OpenResty to make sure the best compatibility and stability:

http://openresty.org/en/download.html

sumantdas:
@agentzh - I dont think i have mentioned anything related to using old nginx core.

agentzh:
@sumantdas It's definitely older than 1.11.2. Mind you, the latest nginx core version is 1.13.7:

http://nginx.org/

PR #669, opened 2016-02-09
Nginx 1.9.11 ngx_udp_connection_t => ngx_resolver_connection_t

charlesportwoodii:
Hi,

In Nginx 1.9.11, the ```ngx_udp_connection_t``` structure has been renamed to ```ngx_resolver_connection_t``` in ```src/core/ngx_resolver.h``` of Nginx core. (https://github.com/nginx/nginx/commit/1945fff7bfee1eef23933faa0ccfbe05445dcfe1).

Consequently, compilation fails in ```lua-nginx-module/src/ngx_http_lua_socket_udp.h``` and ```lua-nginx-module/src/ngx_http_lua_socket_udp.c``` due to references of ```ngx_udp_connection_t```. The compilation failure is as follows:

```
modules/lua-nginx-module/src/ngx_http_lua_util.c
In file included from modules/lua-nginx-module/src/ngx_http_lua_util.c:38:0:
modules/lua-nginx-module/src/ngx_http_lua_socket_udp.h:34:5: error: unknown type name ‘ngx_udp_connection_t’
     ngx_udp_connection_t             udp_connection;
```

I've attached a simple patch that corrects the issue and allows Nginx 1.9.11 to compile.

----
UPDATE: See #668 for others reporting this issue.

### DIFF
--- a/src/ngx_http_lua_socket_udp.c
+++ b/src/ngx_http_lua_socket_udp.c
@@ -54,7 +54,7 @@ static void ngx_http_lua_socket_udp_read_handler(ngx_http_request_t *r,
     ngx_http_lua_socket_udp_upstream_t *u);
 static void ngx_http_lua_socket_udp_handle_success(ngx_http_request_t *r,
     ngx_http_lua_socket_udp_upstream_t *u);
-static ngx_int_t ngx_http_lua_udp_connect(ngx_udp_connection_t *uc);
+static ngx_int_t ngx_http_lua_udp_connect(ngx_resolver_connection_t *uc);
 static int ngx_http_lua_socket_udp_close(lua_State *L);
 static ngx_int_t ngx_http_lua_socket_udp_resume(ngx_http_request_t *r);
 static void ngx_http_lua_udp_resolve_cleanup(void *data);
@@ -170,7 +170,7 @@ ngx_http_lua_socket_udp_setpeername(lua_State *L)
     ngx_url_t                    url;
     ngx_int_t                    rc;
     ngx_http_lua_loc_conf_t     *llcf;
-    ngx_udp_connection_t        *uc;
+    ngx_resolver_connection_t   *uc;
     int                          timeout;
     ngx_http_lua_co_ctx_t       *coctx;
 
@@ -246,7 +246,7 @@ ngx_http_lua_socket_udp_setpeername(lua_State *L)
             return 2;
         }
 
-        if (u->udp_connection.connection) {
+        if (u->udp_connection.udp) {
             ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                            "lua udp socket reconnect without shutting down");
 
@@ -642,7 +642,7 @@ ngx_http_lua_socket_resolve_retval_handler(ngx_http_request_t *r,
 {
     ngx_http_lua_ctx_t              *ctx;
     ngx_http_lua_co_ctx_t           *coctx;
-    ngx_udp_connection_t            *uc;
+    ngx_resolver_connection_t       *uc;
     ngx_connection_t                *c;
     ngx_http_cleanup_t              *cln;
     ngx_http_upstream_resolved_t    *ur;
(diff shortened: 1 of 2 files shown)


nzt4567:
Great work!

src/ngx_http_lua_socket_udp.c
```diff
@@ -170,7 +170,7 @@ ngx_http_lua_socket_udp_setpeername(lua_State *L)
     ngx_url_t                    url;
     ngx_int_t                    rc;
     ngx_http_lua_loc_conf_t     *llcf;
-    ngx_udp_connection_t        *uc;
+    ngx_resolver_connection_t    *uc;
```
agentzh: Style: the identifier names should be aligned up vertically. I think the `ngx-releng` tool in the nginx-devel-utils repos automatically catches coding style issues like this:

https://github.com/openresty/nginx-devel-utils

Will you please run this tool and fix any coding style issues? Thanks!

src/ngx_http_lua_socket_udp.h
```diff
@@ -31,7 +31,7 @@ struct ngx_http_lua_socket_udp_upstream_s {
     ngx_http_lua_loc_conf_t         *conf;
     ngx_http_cleanup_pt             *cleanup;
     ngx_http_request_t              *request;
-    ngx_udp_connection_t             udp_connection;
+    ngx_resolver_connection_t       connection;
```
agentzh: I think we should still keep the original `udp_connection` field name here. Also the identifier is not aligned up properly.

src/ngx_http_lua_socket_udp.c
```diff
@@ -54,7 +54,7 @@ static void ngx_http_lua_socket_udp_read_handler(ngx_http_request_t *r,
     ngx_http_lua_socket_udp_upstream_t *u);
 static void ngx_http_lua_socket_udp_handle_success(ngx_http_request_t *r,
     ngx_http_lua_socket_udp_upstream_t *u);
-static ngx_int_t ngx_http_lua_udp_connect(ngx_udp_connection_t *uc);
+static ngx_int_t ngx_http_lua_udp_connect(ngx_resolver_connection_t *uc);
```
agentzh: This change will break builds with nginx cores older than 1.9.11, I'm afraid. We need macros to support both.

agentzh: A closer look at the definition of `ngx_resolver_connection_t` reveals that this type does not serve our purpose well since it contains unused fields like `read_buf` and `write_buf`. Maybe it's time to roll out our own `ngx_http_lua_udp_connection_t` type instead?

charlesportwoodii:
@agentzh,

Sorry about the style. I've rebased an update that has the alignment issues correct.

Changes in this rebase:
- Changed ```connection``` back to ```udp_connection```
- Styles

I'm not exactly sure what should be done regarding ```ngx_resolver_connection_t```. In Nginx core, ```ngx_udp_connection_t``` was simply renamed to ```ngx_resolver_connection_t``` with some slight variation to the struct (```connection``` was renamed to ```udp```). 

It may be beneficial to to create something like a ```ngx_http_lua_udp_connection_t``` struct, as that would be an easier place to put the macros in. My C is a little bit rusty though, and I just whipped to get Nginx to compile again = ).

agentzh:
@charlesportwoodii Thanks for your work on this. I'll continue thinking about the details here.

charlesportwoodii:
@agentzh,

Glad to help out what little I can. Sorry for not being much more useful than this though. Preprocessor macros will be a pretty simple solution for the backwards compatibility aspect, but that would make a mess of ```src/ngx_http_lua_socket_udp.c``` and I wanted to avoid that as much as possible.

Let me know if I can help out in any other way.

agentzh:
@charlesportwoodii Thanks for your help thus far :) I'll look into this in the next couple of weeks. The next OpenResty release will be based on NGINX 1.9.11 anyway.

agentzh:
The compilation error should be gone now. Thanks for the report!

PR #671, opened 2016-02-10
Small documentation fix to sample code

typhonius:
Was playing with `sock:receive` and realised the sample code was outputting the wrong variable.


agentzh:
Nice catch. Thanks!

PR #672, opened 2016-02-10
optimize: use lua_pushliteral for string-literal.

cubicdaiya:
A bit optimization.


agentzh:
@cubicdaiya Nice catch. Applied. Thanks!

PR #673, opened 2016-02-11
HTTP 2.0 support in ngx.req.http_version() function

r--w:
HTTP 2.0 support in ngx.req.http_version() function

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -52,6 +52,10 @@ ngx_http_lua_ngx_req_http_version(lua_State *L)
     case NGX_HTTP_VERSION_11:
         lua_pushnumber(L, 1.1);
         break;
+        
+    case NGX_HTTP_VERSION_20:
+        lua_pushnumber(L, 2.0);
+        break;
 
     default:
         lua_pushnil(L);


agentzh:
@r--w Thanks for the patch! Will you please add some test cases to the existing test suite too?

agentzh:
@r--w Okay, the existing test scaffold lacks HTTP 2.0 support. Never mind. I'll merge this anyway.

src/ngx_http_lua_headers.c
```diff
@@ -52,6 +52,10 @@ ngx_http_lua_ngx_req_http_version(lua_State *L)
     case NGX_HTTP_VERSION_11:
         lua_pushnumber(L, 1.1);
         break;
+
```
agentzh: Style: this line has superfluous whitespace characters.

src/ngx_http_lua_headers.c
```diff
@@ -52,6 +52,10 @@ ngx_http_lua_ngx_req_http_version(lua_State *L)
     case NGX_HTTP_VERSION_11:
         lua_pushnumber(L, 1.1);
         break;
+        
+    case NGX_HTTP_VERSION_20:
```
agentzh: We also need to support older NGINX cores without this macro defined.

agentzh:
@r--w I've applied a modified version of your patch to git master with some fixes and documentation updates. Thanks for your contribution!

PR #675, opened 2016-02-13
Fix a typo in Readme

pra85:
Remove extra `the`


agentzh:
@pra85 Thanks for the catch. I've applied a modified version of your patch to git master. BTW, the `README.markdown` file is automatically generated from `doc/HttpLuaModule.wiki` which should be edited manually instead :)

pra85:
:+1:

PR #676, opened 2016-02-15
docs(readme) add get_phase() result for ssl_cert context

thibaultCha:
I believe the result of `ngx.get_phase()` in the SSL context was missing.


agentzh:
@thibaultCha Applied your patch with some minor fixes. Thanks a lot!

PR #683, opened 2016-02-23
docs: add get_phase() result in "balancer" context

thibaultCha:
Was just browsing code and docs when I noticed it was missing too :o


agentzh:
Thanks!

Issue #687, opened 2016-02-26
Buffer split can cause block parsing to produce incorrect errors


pushrax:
The fix for https://github.com/openresty/lua-nginx-module/issues/611 did not take into account that sometimes a single value can be split across the config file buffer boundary. In certain cases, this can cause the block to be prematurely closed, and throw errors for valid configs. We ran into this bug in production under a very subtle case, but the problem can be captured with the following simplified cases:

Example 1 (must be in its own file and start at byte 0):

```
[...]

PR #688, opened 2016-02-26
bugfix: handle case in block parsing where token is split across buffers

pushrax:
This is one possible way to fix https://github.com/openresty/lua-nginx-module/issues/687. The approach is to always move the start of the current token to the start of the buffer before lexing. This ensures that the full value is always seen by the lexer.

I included one of the tests I came up with, but couldn't come up with a good way to include the one that requires exact file positioning.


agentzh:
@pushrax Well, I understand the problem here but I don't quite like your approach since it's too expensive (always moving data in the buffer and filling the gap with more data by reading the file handle after *every* token read). Also it won't cover the case where a single token is larger than the buffer size. I think the ultimate way of solving this issue is to add proper streaming parsing support to the sregex DFA compiler, re.pl.

pushrax:
I agree that this approach isn't efficient, I guess the main reason I submitted this PR is to demonstrate a test for the problem and prove the root cause of the issue.

How difficult do you think it would be to add stremaing support to sregex? I imagine it will mainly involve moving the state machine from locals into a struct, and allocating an instance of that struct in `ngx_http_lua_conf_read_lua_token`. If that is the extent of it, I can probably take a stab at it tomorrow since we need a fix for this ASAP as it is affecting our production infrastructure.

agentzh:
@pushrax Yeah, I appreciate your test case.

Regarding the streaming support in sregex DFA's re.pl, it shouldn't be very difficult. We just need to

1. distinguish eof and end of the current data buffer (we can require the caller pass a boolean "eof" argument to the match C function), similar to the C code generated by Ragel. In case of end of buffer and not eof, it needs to return the current state to the caller for later invocations (see below).
1. allow the caller to pass the current state via the C function argument instead of always assuming the start state and resume executing from the state possibly left by the previous invocation.
1. the offset used in the submatch captures need to be absolute as to the whole data stream, instead of the current buffer.

Well, it's best to look at the C code generated by Ragel and the existing Pike engine of sregex for reference :)

Patches welcome :)

PR #689, opened 2016-02-26
ngx.req.get_post_args: return error message instead of raising an exception

yejingx:
return error message instead of raising an exception when request body is in temp file.

### DIFF
--- a/src/ngx_http_lua_args.c
+++ b/src/ngx_http_lua_args.c
@@ -111,8 +111,6 @@ ngx_http_lua_ngx_req_get_uri_args(lua_State *L)
 
     ngx_http_lua_check_fake_request(L, r);
 
-    lua_createtable(L, 0, 4);
-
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_args.c
```diff
@@ -209,9 +210,13 @@ ngx_http_lua_ngx_req_get_post_args(lua_State *L)
 
     buf = ngx_palloc(r->pool, len);
     if (buf == NULL) {
-        return luaL_error(L, "no memory");
```
agentzh: This is vital enough to be an exception. Better leave this as is.

yejingx: @agentzh thx, I have rebased my commit.

agentzh:
@yejingx Applied a slightly modified version of your patch to git master. Thanks!

PR #699, opened 2016-03-07
Fix conf sample syntax error in README

othree:
The sample with incorrect ending of if block will cause 500 error.


agentzh:
Nice catch. Thanks!

PR #700, opened 2016-03-08
New Feature: added timeout support for balancer_by_lua*

toruneko:
the set_timeout API for balancer_by_lua*

Related: https://groups.google.com/forum/#!topic/openresty/oFN8TfUUgq4/discussion


src/ngx_http_lua_balancer.c
```diff
@@ -489,6 +489,80 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+        *err = "send_timeout must be greater than zero";
+        return NGX_ERROR;
+    }
+
+    if (read_timeout < 0) {
+        *err = "read_timeout must be greater than zero";
+        return NGX_ERROR;
+    }
+
+    ucf = ngx_palloc(r->pool, sizeof(ngx_http_upstream_conf_t));
```
agentzh: Hah, allocating an upstream conf struct for every request is clever but it's a bit wasteful since the conf struct is a bit large (464 bytes). I wonder if we can save the memory a bit taking advantage that the rest of the conf struct fields should never change.

toruneko: I also want to do so but the upstream module add a timer who used the u->conf->xxx_timeout. It's really egg pain. Of cause, I will find a more reasonable way to do it, and do you have some suggest for me? thanks.

src/ngx_http_lua_balancer.c
```diff
@@ -489,6 +489,80 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+    ngx_memcpy(ucf, u->conf, sizeof(ngx_http_upstream_conf_t));
+
+    ucf->connect_timeout = connect_timeout;
+    ucf->send_timeout = send_timeout;
+    ucf->read_timeout = read_timeout;
+
+    u->conf = ucf;
+
+    return NGX_OK;
+}
```
agentzh: Style: 2 blank lines needed between function definitions. Better use the `ngx-releng` tool from the `nginx-devel-utils` repos to check your code.

agentzh:
Maybe we just need to patch the nginx core for OpenResty so we can use the more efficient code path and for standard NGINX core, we just adopt your approach as a (slower) fallback solution.

xbsura:
@agentzh  I think If not do some change in nginx core, and just change the global timeout conf is not safe as all request share this conf

agentzh:
@xbsura The change done by @toruneko does not modify the globally shared conf struct. Rather, it just changes the pointer inside `ngx_http_request_t` struct if you are looking more carefully :)

toruneko:
I'm sorry to reply this PR now. I was try to patch the nginx core few days ago, such as change the `ngx_http_upstream_t` struct (add three timeout variable), and modify `ngx_http_upstream.c` file where used `u->conf->xxx_timeout` (replace to u->xxx_timeout). But finally, i reset all. 

The meaning of @xbsura is that we do not alloc memory, and just change timeout value unsafely.

xbsura:
@agentzh 
Oh, he used a memcopy copyed all global conf
yes, this may cost a little more time

We have similar scene which need such a timeout control, I add such timeout var in request struct and just set these variables, may be faster

But this need change not only ngx_lua_module but nginx project, maybe slower or difficult to be accepted

xbsura:
@toruneko 
Maybe you are right, patching nginx core may be a better choice.

toruneko:
```
From 30476c4190ae7b290a0438b19fca17fa7ab01313 Mon Sep 17 00:00:00 2001
From: “toruneko” <toruneko@outlook.com>
Date: Fri, 25 Mar 2016 14:13:17 +0800
Subject: [PATCH] patch-timeout

---
 src/http/modules/ngx_http_proxy_module.c |  3 +++
 src/http/ngx_http_upstream.c             | 16 ++++++++--------
 src/http/ngx_http_upstream.h             |  3 +++
 3 files changed, 14 insertions(+), 8 deletions(-)

diff --git a/src/http/modules/ngx_http_proxy_module.c b/src/http/modules/ngx_http_proxy_module.c
index 8ee32f4..609eea4 100644
--- a/src/http/modules/ngx_http_proxy_module.c
+++ b/src/http/modules/ngx_http_proxy_module.c
@@ -709,6 +709,9 @@ ngx_http_proxy_handler(ngx_http_request_t *r)
     u->output.tag = (ngx_buf_tag_t) &ngx_http_proxy_module;
 
     u->conf = &plcf->upstream;
+    u->read_timeout = u->conf->read_timeout;
+    u->send_timeout = u->conf->send_timeout;
+    u->connect_timeout = u->conf->connect_timeout;
 
 #if (NGX_HTTP_CACHE)
     u->create_key = ngx_http_proxy_create_key;
diff --git a/src/http/ngx_http_upstream.c b/src/http/ngx_http_upstream.c
index 040bda1..6143dd4 100644
--- a/src/http/ngx_http_upstream.c
+++ b/src/http/ngx_http_upstream.c
@@ -1313,7 +1313,7 @@ ngx_http_upstream_connect(ngx_http_request_t *r, ngx_http_upstream_t *u)
     u->request_sent = 0;
 
     if (rc == NGX_AGAIN) {
-        ngx_add_timer(c->write, u->conf->connect_timeout);
+        ngx_add_timer(c->write, u->connect_timeout);
         return;
     }
 
@@ -1512,7 +1512,7 @@ ngx_http_upstream_send_request(ngx_http_request_t *r, ngx_http_upstream_t *u)
     }
 
     if (rc == NGX_AGAIN) {
-        ngx_add_timer(c->write, u->conf->send_timeout);
+        ngx_add_timer(c->write, u->send_timeout);
 
         if (ngx_handle_write_event(c->write, u->conf->send_lowat) != NGX_OK) {
             ngx_http_upstream_finalize_request(r, u,
@@ -1537,7 +1537,7 @@ ngx_http_upstream_send_request(ngx_http_request_t *r, ngx_http_upstream_t *u)
         c->tcp_nopush = NGX_TCP_NOPUSH_UNSET;
     }
 
-    ngx_add_timer(c->read, u->conf->read_timeout);
+    ngx_add_timer(c->read, u->read_timeout);
 
     if (c->read->ready) {
         ngx_http_upstream_process_header(r, u);
@@ -2182,7 +2182,7 @@ ngx_http_upstream_process_body_in_memory(ngx_http_request_t *r,
     }
 
     if (rev->active) {
-        ngx_add_timer(rev, u->conf->read_timeout);
+        ngx_add_timer(rev, u->read_timeout);
 
     } else if (rev->timer_set) {
         ngx_del_timer(rev);
@@ -2474,7 +2474,7 @@ ngx_http_upstream_send_response(ngx_http_request_t *r, ngx_http_upstream_t *u)
         p->cyclic_temp_file = 0;
     }
 
-    p->read_timeout = u->conf->read_timeout;
+    p->read_timeout = u->read_timeout;
     p->send_timeout = clcf->send_timeout;
     p->send_lowat = clcf->send_lowat;
 
@@ -2731,7 +2731,7 @@ ngx_http_upstream_process_upgraded(ngx_http_request_t *r,
     }
 
     if (upstream->write->active && !upstream->write->ready) {
-        ngx_add_timer(upstream->write, u->conf->send_timeout);
+        ngx_add_timer(upstream->write, u->send_timeout);
 
     } else if (upstream->write->timer_set) {
         ngx_del_timer(upstream->write);
@@ -2743,7 +2743,7 @@ ngx_http_upstream_process_upgraded(ngx_http_request_t *r,
     }
 
     if (upstream->read->active && !upstream->read->ready) {
-        ngx_add_timer(upstream->read, u->conf->read_timeout);
+        ngx_add_timer(upstream->read, u->read_timeout);
 
     } else if (upstream->read->timer_set) {
         ngx_del_timer(upstream->read);
@@ -2936,7 +2936,7 @@ ngx_http_upstream_process_non_buffered_request(ngx_http_request_t *r,
     }
 
     if (upstream->read->active && !upstream->read->ready) {
-        ngx_add_timer(upstream->read, u->conf->read_timeout);
+        ngx_add_timer(upstream->read, u->read_timeout);
 
     } else if (upstream->read->timer_set) {
         ngx_del_timer(upstream->read);
diff --git a/src/http/ngx_http_upstream.h b/src/http/ngx_http_upstream.h
index 9f96c50..b213a4f 100644
--- a/src/http/ngx_http_upstream.h
+++ b/src/http/ngx_http_upstream.h
@@ -283,6 +283,9 @@ struct ngx_http_upstream_s {
     ngx_chain_writer_ctx_t           writer;
 
     ngx_http_upstream_conf_t        *conf;
+    ngx_msec_t                       connect_timeout;
+    ngx_msec_t                       send_timeout;
+    ngx_msec_t                       read_timeout;
 
     ngx_http_upstream_headers_in_t   headers_in;
 
-- 
2.3.8 (Apple Git-58)
```

xbsura:
@toruneko 
Yes, this is absolutely what I did
But I am afraid for nginx, timeout for upstream but not for request may be more resonable
whatever, I hope this patch will be accepted too as I really dont want to use a personnal branch, which is difficult to upgrade in the future

src/ngx_http_lua_balancer.c
```diff
@@ -537,6 +537,72 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+        *err = "no upstream peer data found";
+        return NGX_ERROR;
+    }
+
+    if (timeout < 0) {
+        *err = "timeout must be equals or greater than zero";
+        return NGX_ERROR;
+    }
+
+    if (ngx_strncmp("read", type, 4) == 0) {
```
agentzh: C-land string comparison is much more expensive than Lua land and we should avoid it.

src/ngx_http_lua_balancer.c
```diff
@@ -537,6 +537,72 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+        u->read_timeout = timeout;
+        return NGX_OK;
+    }
+
+    if (ngx_strncmp("send", type, 4) == 0) {
+        u->send_timeout = timeout;
```
agentzh: Instead of assuming the nginx core is patched with these new fields, we should branch the code via special macros. Otherwise ngx_lua is now failing to build against the standard nginx cores, which is unacceptable.

src/ngx_http_lua_balancer.c
```diff
@@ -537,6 +537,72 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+    lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
+
+    bp = lmcf->balancer_peer_data;
+    if (bp == NULL) {
+        *err = "no upstream peer data found";
+        return NGX_ERROR;
+    }
+
+    if (timeout < 0) {
+        *err = "timeout must be equals or greater than zero";
```
agentzh: There's a grammar error in the message text.  You can just say "negative timeout".

src/ngx_http_lua_balancer.c
```diff
@@ -537,6 +537,72 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
 
 
 int
+ngx_http_lua_ffi_balancer_set_peer_timeout(ngx_http_request_t *r,
+    const u_char *type, int timeout, char **err)
```
agentzh: Instead of using `int` for the timeout argument, I think `long` is better, which is usually larger than `int` on 64-bit systems. And timeout values could be large.

agentzh:
@toruneko @xbsura Please review my version of the nginx core patch:

https://github.com/openresty/openresty/blob/master/patches/nginx-1.9.15-upstream_timeout_fields.patch

I introduced the `HAVE_UPSTREAM_TIMEOUT_FIELDS` macro in the patch so that ngx_lua can work with both the patched nginx and the stock nginx :)

agentzh:
Closing this in favor of #812.

PR #705, opened 2016-03-10
bugfix: ngx.req.append_body may go to endless loop

doujiang24:
when ngx.req.init_body have not specified the buffer_size and
(client_body_buffer_size = 0 or Content Length = 0)

And now we honor the `client_body_in_file_only` in `ngx.req.append_body` too, I'm not sure if this is ok, if yes, then I should add some test case too.


agentzh:
@doujiang24 Patch applied with some minor modifications. Thanks a lot for the fix!

PR #712, opened 2016-03-18
feature: add tcpsock:bind api

doujiang24:
Just like the standard proxy_bind directive, this api makes the outgoing connection to a upstream server originate from the specified local IP address.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6438,6 +6438,7 @@ Here is an example for connecting to a TCP server from the specified local IP ad
  location /test {
      content_by_lua_block {
          local sock = ngx.socket.tcp()
+         -- assume "192.168.1.10" is the local ip address
          local ok, err = sock:bind("192.168.1.10")
          if not ok then
              ngx.say("failed to bind")
--- a/README.markdown
+++ b/README.markdown
@@ -2926,6 +2926,7 @@ Nginx API for Lua
 * [udpsock:settimeout](#udpsocksettimeout)
 * [ngx.socket.stream](#ngxsocketstream)
 * [ngx.socket.tcp](#ngxsockettcp)
+* [tcpsock:bind](#tcpsockbind)
 * [tcpsock:connect](#tcpsockconnect)
 * [tcpsock:sslhandshake](#tcpsocksslhandshake)
 * [tcpsock:send](#tcpsocksend)
@@ -6383,6 +6384,7 @@ ngx.socket.tcp
 
 Creates and returns a TCP or stream-oriented unix domain socket object (also known as one type of the "cosocket" objects). The following methods are supported on this object:
 
+* [bind](#tcpsockbind)
 * [connect](#tcpsockconnect)
 * [sslhandshake](#tcpsocksslhandshake)
 * [send](#tcpsocksend)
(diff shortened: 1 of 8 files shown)


JakSprats:
I complied the patch and used some simple scripts to test them, everything looks good.

[bind_patch.zip](https://github.com/openresty/lua-nginx-module/files/254115/bind_patch.zip)

(Not included in files) Additionally, I tested binding to a bad address, test failed (correct behavior)

agentzh:
I'll look into this as soon as I can manage :)

JakSprats:
I do have one comment, but it is not directly related to sock:bind() but I found it while testing sock:bind(), so I am posting it here. This applies to all sock:* calls and is related to the luaL_error() call.

There is a different behavior for different types of errors.

For instance if I try to bind to the ip "A.B.C.D" I get a error as the 2nd return value of the call
e.g. local ok, err = sock:bind("A.B.C.D");

Whereas if I put some garbage arguments on the end of the sock:bind() call
e.g. local ok, err = sock:bind("127.0.0.1", 8888, 1, 2, 3);
The error propagates directly to the logs/error.log level and the script is killed
logs/error.log:
2016/05/11 13:24:26 [error] 3023#0: *1 lua entry thread aborted: runtime error: nginx/./locations/test.lua:23: ngx.socket bind: expecting at least 2 arguments (including the object) but seen 6
stack traceback:
coroutine 0:
	[C]: in function 'bind'

It just seems like there should be only one error path

Anyways, not a big deal, but it's a problem I ran into, and this is my (lazy) way of documenting it :)

alonbg:
@agentzh @JakSprats
I'd like to add the IP_TRANSPARENT socket option.
Per my understanding I just need to set `pc->transparent` and [nginx will take of the rest](https://github.com/nginx/nginx/blob/master/src/event/ngx_event_connect.c#L86)

where would be a better location to set this flag? adding a parameter to `udp.setpeername() / tcp.connect()` or to this new `bind()` api ?

@JakSprats, just for my understanding, why did you choose to implement a separate `bind()` and not do it all in one go with additional args to 'connect()` ?

Thanks!

agentzh:
@JakSprats We distinguish between bad argument errors due to API misuse and other kinds of errors. The former usually results in Lua exceptions so that the user cannot (easily) ignore it.

@alonbg I think we'd better add support for the `transparent` argument of the `setoption` method, similar to LuaSocket's same-name method:

http://w3.impa.br/~diego/software/luasocket/tcp.html#setoption

alonbg:
@agentzh Thanks!
btw, for your reference, some time ago, I opened an [issue@LuaSockets](https://github.com/diegonehab/luasocket/issues/176) for this.

iseriser:
@agentzh Please, review and add this code into main branch. Almost two years have passed.

kiwina:
love to see the bind option merged as well, it is really usefull

djaara:
What is the plan with this pull request?  We would like to leverage `bind()` to ensure right source IP for our requests.

Scott41382:
Stop

alerque:
@agentzh @alonbg Is anything in the linked LuaSockets issue still relevant? I'm one of the new maintainers but for that particular issue I'm a bit lost. I'm happy to facilitate if somebody has a contribution and concrete reason why it is needed, but there isn't enough to go on and move it forward as it stands. Just FYI.

zhuizhuhaomeng:
The feature has been merged from the command line. I forgot to close this PR.

PR #721, opened 2016-03-24
bug fix: balancer_by_lua make upstream_addr err when using more than one set_current_peer

xbsura:
copy a pointer of virtual peer name to nginx instead of directly return the virtual peer name, to solve the upstream_addr error when using more than one time set_current_peer

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -301,21 +301,22 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
         pc->sockaddr = bp->sockaddr;
         pc->socklen = bp->socklen;
 
-        /* return a copy pointer of bp->host to pc, in case upstream_addr
+        /* balancer module has no peer actually, it just create one virtual peer
+         * when necessary, 
+         * at the same time, because the peer_name is returned as a pointer, its
+         * content changed after using balancer_by_lua, if we just returned the pointer,
+         * nginx may always use the last peer_name, so we return a copy pointer of
+         * bp->host to pc, in case upstream_addr
          * always use the last peer name
          */
         peer_name = ngx_palloc(r->pool, sizeof(ngx_str_t));
         if (peer_name == NULL) {
-            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                          "failed to malloc mem from request pool");
             return NGX_ERROR;
 
         } else {
             ngx_memzero(peer_name, sizeof(ngx_str_t));
             peer_name->data = ngx_palloc(r->pool, bp->host.len);
             if (peer_name->data == NULL) {
-                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                              "failed to malloc mem from request pool");
                 return NGX_ERROR;
 
             } else {
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -304,10 +304,22 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
          * always use the last peer name
          */
         ngx_str_t *peer_name = ngx_palloc(r->pool, sizeof(ngx_str_t));
-        peer_name->data = ngx_palloc(r->pool, bp->host.len);
-        ngx_memcpy(peer_name->data, bp->host.data, bp->host.len);
-        peer_name->len = bp->host.len;
-        pc->name = peer_name;
+        if (peer_name == NULL) {
+            pc->name = &bp->host;
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                "failed to malloc mem from request pool, upstream addr may not correct");
+        } else {
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                pc->name = &bp->host;
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                    "failed to malloc mem from request pool, upstream addr may not correct");
+            } else {
+                ngx_memcpy(peer_name->data, bp->host.data, bp->host.len);
+                peer_name->len = bp->host.len;
+                pc->name = peer_name;
+            }
+        }
 
         bp->rrp.peers->single = 0;
 
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -252,6 +252,7 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
     ngx_http_lua_srv_conf_t            *lscf;
     ngx_http_lua_main_conf_t           *lmcf;
     ngx_http_lua_balancer_peer_data_t  *bp = data;
+    ngx_str_t                          *peer_name;
 
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, pc->log, 0,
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +299,28 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
     if (bp->sockaddr && bp->socklen) {
         pc->sockaddr = bp->sockaddr;
         pc->socklen = bp->socklen;
-        pc->name = &bp->host;
+
+        /* return a copy pointer of bp->host to pc, in case upstream_addr
+         * always use the last peer name
+         */
+        ngx_str_t *peer_name = ngx_palloc(r->pool, sizeof(ngx_str_t));
```
agentzh: Please following nginx coding style. We never define new local variables in the middle of the code chunk.

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +299,28 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+        if (peer_name == NULL) {
```
agentzh: When allocation failed, we should return NGX_ERROR immediately without proceeding.

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +299,28 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+            pc->name = &bp->host;
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                "failed to malloc mem from request pool, upstream addr may not correct");
+        } else {
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                pc->name = &bp->host;
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                    "failed to malloc mem from request pool, upstream addr may not correct");
+            } else {
```
agentzh: Style: need a blank line before `} else {`. Please check the rest of the code base for the right coding style. Thank you.

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +299,28 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+        if (peer_name == NULL) {
+            pc->name = &bp->host;
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                "failed to malloc mem from request pool, upstream addr may not correct");
+        } else {
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                pc->name = &bp->host;
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                    "failed to malloc mem from request pool, upstream addr may not correct");
```
agentzh: Style: alignment problem. I suggest you run the `ngx-releng` script from the `nginx-devel-utils` GitHub repo to check your code.

agentzh:
@xbsura Another thing: will you please add some corresponding test cases to the existing test suite? (It's fine to add tests to `lua-resty-core`'s test suite if it is easier for you.) Thanks!

src/ngx_http_lua_balancer.c
xbsura: all fixed and checked done with ngx-releng
Thakns

xbsura:
@agentzh 
Ok, I will try to add a case to test suite

xbsura:
@agentzh 
Is there anything else I can do for this pull?

thanks

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +300,31 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+
+        } else {
+            ngx_memzero(peer_name, sizeof(ngx_str_t));
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "failed to malloc mem from request pool");
+                return NGX_ERROR;
+
+            } else {
```
agentzh: We do not need the `else` part here since the branch above never falls through.

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +300,31 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "failed to malloc mem from request pool");
+            return NGX_ERROR;
+
+        } else {
+            ngx_memzero(peer_name, sizeof(ngx_str_t));
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                              "failed to malloc mem from request pool");
```
agentzh: We usually do not log memory allocation failures in the nginx world.

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +300,31 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+        /* return a copy pointer of bp->host to pc, in case upstream_addr
+         * always use the last peer name
+         */
+        peer_name = ngx_palloc(r->pool, sizeof(ngx_str_t));
+        if (peer_name == NULL) {
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "failed to malloc mem from request pool");
+            return NGX_ERROR;
+
+        } else {
```
agentzh: Same as below.

agentzh:
@xbsura Made some more comments. Will you take care of them? Thank you very much!

xbsura:
@agentzh 
ok, I removed the err log and add some comments
But I am not sure the alloc will always success, maybe it is reasonable to check all alloc result
thanks

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +300,32 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+        if (peer_name == NULL) {
+            return NGX_ERROR;
+
+        } else {
+            ngx_memzero(peer_name, sizeof(ngx_str_t));
+            peer_name->data = ngx_palloc(r->pool, bp->host.len);
+            if (peer_name->data == NULL) {
+                return NGX_ERROR;
+
+            } else {
```
agentzh: We do not need this `else` since the previous branch always returns.

xbsura: oh... I understand, else removed, thanks

src/ngx_http_lua_balancer.c
```diff
@@ -299,7 +300,30 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+         * nginx may always use the last peer_name, so we return a copy pointer of
+         * bp->host to pc, in case upstream_addr
+         * always use the last peer name
+         */
+        peer_name = ngx_palloc(r->pool, sizeof(ngx_str_t));
+        if (peer_name == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_memzero(peer_name, sizeof(ngx_str_t));
```
agentzh: We do not need this since we explicitly initialize both of the fields of `ngx_str_t` later anyway.

agentzh:
Merged with minor modifications. Thanks!

Issue #719, opened 2016-03-22
does the 5th argument of  ngx.re.match make the 4th argument ctx.pos invalid?


iorichina:
the code:
`
local ctx, m = { pos = 1 }, {}
local _, err = ngx.re.match("2016-11-3 03:07:09", [=[^(\d\d\d\d)]=], "", ctx)
print(m[0], ' ', _[0], ' ', ctx.pos)
-- will output:nil 2016 5

local ctx, m = { pos = 1 }, {}
local _, err = ngx.re.match("2016-11-3 03:07:09", [=[^(\d\d\d\d)]=], "", ctx, m)
print(m[0], ' ', _[0], ' ', ctx.pos)
-- will output:2016 2016 1

`
openresty info:
`# /usr/local/openresty-1.9.7.2/bin/resty -V
resty 0.03
nginx version: openresty/1.9.7.2
built by gcc 4.4.7 20120313 (Red Hat 4.4.7-11) (GCC) 
built with OpenSSL 1.0.0s 11 Jun 2015
TLS SNI support enabled
`

agentzh:
@iorichina This looks like a bug to me. Will you contribute a patch for it? Thanks!

iorichina:
春哥吩咐，义不容辞，我会try，但不做承诺，lol

agentzh:
@iorichina Thanks. But please use English here. Thank you :)

iorichina:
@agentzh Sir, yes sir! lol

agentzh:
The fix has been committed to git master.

PR #726, opened 2016-03-29
fix issue#719 ngx_http_lua_ngx_re_match_helper bug

iorichina:
issue https://github.com/openresty/lua-nginx-module/issues/719


agentzh:
@iorichina Nice catch. Will you add a corresponding test case to the existing test suite as well? Thanks!

PR #727, opened 2016-03-29
fix issue #719

rako9000:
check nargs >= 4 to support nargs == 5

### DIFF
--- a/t/034-match.t
+++ b/t/034-match.t
@@ -733,7 +733,7 @@ done
     GET /re
 --- response_body
 pos: 1
-m: 
+m:
 --- no_error_log
 [error]
 
@@ -1169,3 +1169,21 @@ hello
 hello
 false
 
+
+=== TEST 50: Bug #719 5th arg blind 4th arg
+--- config
+    location /re {
+        content_by_lua '
+            local ctx, m = { pos = 5 }, {};
+            local _, err = ngx.re.match("20172016-11-3 03:07:09", [=[(\d\d\d\d)]=], "", ctx, m);
+            if m then
+                ngx.say(m[0], " ", _[0], " ", ctx.pos)
+            else
+                ngx.say("not matched!")
+            end
+        ';
+    }
+--- request
+    GET /re
+--- response_body
+2016 2016 9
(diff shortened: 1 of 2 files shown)


agentzh:
@rako9000 Seems like being duplicate with #726 ?

rako9000:
#726 changes one line of code, and I think there should be two places to be changed

rako9000:
If change only one line like #726, this case will not work

rako@t450s:/usr/local/openresty/bin$ ./resty -e 'local ctx, m = { pos = 5 }, {};local _, err = ngx.re.match("20172016-11-3 03:07:09", [=[(\d\d\d\d)]=], "", ctx);print(m[0], " ", _[0], " ", ctx.pos)'
nil 2016 9
rako@t450s:/usr/local/openresty/bin$ ./resty -e 'local ctx, m = { pos = 5 }, {};local _, err = ngx.re.match("20172016-11-3 03:07:09", [=[(\d\d\d\d)]=], "", ctx, m);print(m[0], " ", _[0], " ", ctx.pos)'
2017 2017 5

agentzh:
@rako9000 Then I think you should close #726 and keep this one. There is no reason to keep both :)

agentzh:
@rako9000 Will you add the test cases to the existing test suite? Thanks!

rako9000:
@agentzh 
1.I have no right to close #726 because it belongs to @iorichina, am i right?
2.I am a newbie to this project, and I will try to add this to the test suite, but may take some time
3.There are lots of issues which are not bugs and will not be discussed, why keep them alive?

agentzh:
@rako9000 Sorry I got confused with the creators of #727 and #726. Thanks for your test case! I'll look into this.

iorichina:
good job @rako9000 I will try to close issue #719 & #726

iorichina:
I think i should just close pull#726 and i had reopen the issue#719. lol

t/034-match.t
```diff
@@ -733,7 +733,7 @@ done
     GET /re
 --- response_body
 pos: 1
-m: 
+m:
```
agentzh: BTW, this change breaks this unrelated test case.

agentzh:
I've merged a modified version of this patch to git master since this PR breaks existing tests as is. Thank you both!

PR #733, opened 2016-04-09
Implement 418 as an HTTP status constant

p0pr0ck5:
Everybody loves teapots.


agentzh:
@p0pr0ck5 Hmm, I'm afraid the real world use cases of it does not justify a name in the `ngx.` table :) Better hard-code the number 418 in your Lua code instead.

p0pr0ck5:
Awww, where's your sense of fun? ;)

agentzh:
@p0pr0ck5 Sorry, I'm more concerned about performance penalty :)

PR #736, opened 2016-04-13
[feature] ngx.stub_status API

ruoshan:
ngx.stub_status() return a table containing the metrics from the [http_stub_status_module](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html).

I find it quite useful for monitoring the Nginx's connection status without parsing the `stub_status` output.

:)


agentzh:
@ruoshan Better make this a standalone nginx C module similar to the existing ngx_http_upstream_lua module? See

https://github.com/openresty/lua-upstream-nginx-module

We are trying hard to keep the ngx_lua core minimal :) Thanks for your understanding.

ruoshan:
will do, thanks.

PR #740, opened 2016-04-22
cppcheck fixes (I think we could add cppcheck as a CI step)

chipitsine:
[src/ngx_http_lua_socket_tcp.c:1184]: (warning) Logical conjunction always evaluates to false: n < 1 && n > 4.
[src/ngx_http_lua_util.c:916]: (warning) Either the condition 'ctx==0' is redundant or there is possible null pointer dereference: ctx.


agentzh:
@chipitsine Thanks for the catch. I've applied your patch with some modifications (a new test case exercising the issue and reworded the commit log message a bit).

agentzh:
@chipitsine We could add cppcheck to the Travis CI config if there's no false positives :)

PR #741, opened 2016-04-23
skip ipv6 if not supported

chipitsine:
travis ci does not support ipv6, the best we can do, skip it

### DIFF
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -1,5 +1,6 @@
 # vim:set ft= ts=4 sw=4 et fdm=marker:
 
+
 use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
@@ -3298,6 +3299,7 @@ failed to receive a line: closed []
 close: 1 nil
 --- no_error_log
 [error]
+--- skip_eval: 3: system("ping6 -c 1 ::1 >/dev/null 2>&1") ne 0
 
 
 


t/058-tcp-socket.t
```diff
@@ -1,5 +1,6 @@
 # vim:set ft= ts=4 sw=4 et fdm=marker:
 
+
```
agentzh: Why this change?

t/058-tcp-socket.t
```diff
@@ -3298,6 +3299,7 @@ failed to receive a line: closed []
 close: 1 nil
 --- no_error_log
 [error]
+--- skip_eval: 3: system("ping6 -c 1 ::1 >/dev/null 2>&1") ne 0
```
agentzh: Maybe we should make this nicer via `--- use_ipv6` in `Test::Nginx::Socket`?

t/058-tcp-socket.t
chipitsine: I've forgotten to have a look at "git diff", do not pay attention to that line

t/058-tcp-socket.t
chipitsine: do you mean skip all tests in 058 ? only single test fails.

chipitsine: anyway, I'm not very familiar with Test:Nginx (I'm just studying that framework), you can use whatever you think is better

agentzh: @chipitsine Other nginx modules of openresty have test cases against IPv6 as well. So better abstract it away? We can check if `--with-ipv6` is present in the `nginx -V` output instead of calling `ping6`.

agentzh: @chipitsine `--- use_ipv6` is just my idea. It does not exist yet. Needs patches for `Test::Nginx` :)

chipitsine: sounds good

chipitsine: can we merge my dirty hack for now?
we can polish it later if we will have enough time and will

agentzh:
Merged with modifications.

Issue #748, opened 2016-04-27
*_block directive parsing edge-case (unexpected lua closing long-bracket)


thibaultCha:
Hey there,

I did not find any discussion related to this topic already. I recently ran into an 'edge case' of the new `*_by_lua_block` directives validations by `ngx_http_lua_directive` with OpenResty `1.9.7.4`. I see the same code being used in the streaming module too.

Here is a minimal reproducible example:
```lua
content_by_lua_block {
    local t = {1}
    ngx.say(t[t[1]])
}
```

This example triggers this error https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_directive.c#L1428 because of the double closing brackets `]]` not matching opening ones. However, we are dealing with valid Lua code, handled correctly by all interpreters. A test in `t/132-lua-blocks.t` successfully prevents the execution of code blocks such as:

```lua
]]
```

but the parsing also prevents legit Lua code from running (of course one can find a workaround by caching `t[1]` in a local, but still). I believe the intent here is to fail early, in order to avoid runtime syntax errors. I also do not see such strict parsing with the old `*_by_lua` forms, which does fail at runtime on this.

I am not sure this can be fixed with the current parser, because it seems to me its lacking the semantics of the code being analyzed. And if it cannot, **should we favor valid Lua code over strong validation, or the other way around, aka keep the current behavior?**

agentzh:
@thibaultCha Yeah, this is a bug and should get fixed. Thanks for the report. I think the easiest way is to silently ignore such unexpected left-long-brackets-like sequences. Will you contribute a patch for that? It should be relatively easy I suppose. We surely want to accept any valid Lua code snippet in such blocks :)

thibaultCha:
Yes in fact I already worked on a patch for dropping the support, when I looked into this. Will do :)

agentzh:
@thibaultCha Nice. Thanks!

PR #749, opened 2016-04-27
bugfix: ignore unexpected closing long-brackets in '*_by_lua_block' directives

thibaultCha:
Fix for #748 dropping support for detecting unexpected closing long brackets.

Should I propose the same patch against the ngx_lua_stream module or will you apply it there too?

### DIFF
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -1417,20 +1417,8 @@ ngx_http_lua_conf_lua_block_parse(ngx_conf_t *cf, ngx_command_t *cmd)
             break;
 
         case FOUND_LBRACKET_STR:
-
-            break;
-
         case FOUND_LBRACKET_CMT:
-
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_directive.c
```diff
@@ -1417,20 +1417,8 @@ ngx_http_lua_conf_lua_block_parse(ngx_conf_t *cf, ngx_command_t *cmd)
             break;
 
         case FOUND_LBRACKET_STR:
-
-            break;
-
         case FOUND_LBRACKET_CMT:
-
-            break;
```
thibaultCha: I hope that's fine, I just figured those could all be handled by a single `break` too.

agentzh: Fine.

agentzh:
@thibaultCha Just merged your patch to master. Thanks a lot!

PR #750, opened 2016-04-28
Support for passphrase protected key

guanglinlv:
read a private key from a BIO using the specify passphrase.


zhuizhuhaomeng:
some tweak to this https://github.com/openresty/lua-nginx-module/pull/1991/files.
will merge the latter.

PR #751, opened 2016-04-28
bugfix: changing peers in balancer_by_lua* might lead to stale values of $upstream_addr

ops-dev-cn:
Do we need to create local variable peer_name?

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -29,7 +29,7 @@ struct ngx_http_lua_balancer_peer_data_s {
     struct sockaddr                    *sockaddr;
     socklen_t                           socklen;
 
-    ngx_str_t                           host;
+    ngx_str_t                           *host;
     in_port_t                           port;
 
     int                                 last_peer_state;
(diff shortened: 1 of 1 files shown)


agentzh:
@ops-dev-cn Yeah, your approach is much simpler and better. Merged with fixes. Thank you!

src/ngx_http_lua_balancer.c
```diff
@@ -325,30 +325,7 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
         pc->socklen = bp->socklen;
         pc->cached = 0;
         pc->connection = NULL;
-
```
guanglinlv: @ops-dev-cn 

the definition of `peer_name` variable should be sanitized together.

src/ngx_http_lua_balancer.c
```diff
@@ -549,7 +526,7 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
     if (url.addrs && url.addrs[0].sockaddr) {
         bp->sockaddr = url.addrs[0].sockaddr;
         bp->socklen = url.addrs[0].socklen;
-        bp->host = url.addrs[0].name;
+        bp->host = &url.addrs[0].name;;
```
guanglinlv: @ops-dev-cn , it must be a typo, double semicolon.

agentzh: @guanglinlv No worries. I've already fixed these tiny things myself :)

PR #760, opened 2016-05-12
Add native ngx_crc32_update to lua.

Moezen:
To implement ngx_consistent_hash with lua, and make sure its output is as same as nginx, I require native ngx_crc32_update. But it isn't supported in ngx_lua now. Use zlib to implement ngx_crc32_update will cause lower efficiency beacuse of the additional xor operation.


agentzh:
@Moezen We want to keep the ngx_lua core minimal nowadays. Better add such Lua API to lua-resty-string instead. See

https://github.com/openresty/lua-resty-string

Issue #757, opened 2016-05-06
Build fails with OpenSSL 1.1


neilstuartcraig:
Hi

I have a project in which I am doing a custom NGINX build and want to use this module. The build is an RPM build (currently on CentOS 7) and it's failing with NGINX (1.10.0) with OpenSSL 1.1-pre4 (pre5 doesn't compile with NGINX):

```
In file included from /tmp/tmp44TVHu/BUILD/nginx-1.10.0/openssl-1.1.0-pre4/.openssl/include/openssl/ssl.h:912:0,
                 from src/event/ngx_event_openssl.h:15,
                 from src/core/ngx_core.h:83,
                 from /tmp/tmp44TVHu/BUILD/nginx-1.10.0/lua-nginx-module/src/ddebug.h:13,
                 from /tmp/tmp44TVHu/BUILD/nginx-1.10.0/lua-nginx-module/src/ngx_http_lua_ssl_certby.c:10:
/tmp/tmp44TVHu/BUILD/nginx-1.10.0/lua-nginx-module/src/ngx_http_lua_ssl_certby.c: In function 'ngx_http_lua_ffi_ssl_get_tls1_version':
/tmp/tmp44TVHu/BUILD/nginx-1.10.0/openssl-1.1.0-pre4/.openssl/include/openssl/tls1.h:182:20: error: dereferencing pointer to incomplete type
                 ((s->version >> 8) == TLS1_VERSION_MAJOR ? s->version : 0)
                    ^
/tmp/tmp44TVHu/BUILD/nginx-1.10.0/lua-nginx-module/src/ngx_http_lua_ssl_certby.c:524:18: note: in expansion of macro 'TLS1_get_version'
     return (int) TLS1_get_version(ssl_conn);
                  ^
/tmp/tmp44TVHu/BUILD/nginx-1.10.0/openssl-1.1.0-pre4/.openssl/include/openssl/tls1.h:182:61: error: dereferencing pointer to incomplete type
                 ((s->version >> 8) == TLS1_VERSION_MAJOR ? s->version : 0)
                                                             ^
/tmp/tmp44TVHu/BUILD/nginx-1.10.0/lua-nginx-module/src/ngx_http_lua_ssl_certby.c:524:18: note: in expansion of macro 'TLS1_get_version'
     return (int) TLS1_get_version(ssl_conn);
                  ^
[...]

PR #761, opened 2016-05-12
Initial OpenSSL 1.1.0 support

ghedo:
This fixes some errors when building with OpenSSL master as reported in #757.

Note that this is still not enough to make the build succeed, since changes to OpenSSL are also needed (and currently in review).

### DIFF
--- a/t/129-ssl-socket.t
+++ b/t/129-ssl-socket.t
@@ -1129,7 +1129,7 @@ SSL reused session
             sock:settimeout(2000)
 
             do
-                local ok, err = sock:connect("iscribblet.org", 443)
+                local ok, err = sock:connect("openresty.org", 443)
                 if not ok then
                     ngx.say("failed to connect: ", err)
                     return
@@ -1137,7 +1137,7 @@ SSL reused session
 
                 ngx.say("connected: ", ok)
 
-                local session, err = sock:sslhandshake(nil, "iscribblet.org")
+                local session, err = sock:sslhandshake(nil, "openresty.org")
                 if not session then
                     ngx.say("failed to do SSL handshake: ", err)
                     return
@@ -1145,7 +1145,7 @@ SSL reused session
 
                 ngx.say("ssl handshake: ", type(session))
 
-                local req = "GET / HTTP/1.1\\r\\nHost: iscribblet.org\\r\\nConnection: close\\r\\n\\r\\n"
+                local req = "GET /en/ HTTP/1.1\\r\\nHost: openresty.org\\r\\nConnection: close\\r\\n\\r\\n"
                 local bytes, err = sock:send(req)
                 if not bytes then
                     ngx.say("failed to send http request: ", err)
@@ -1174,7 +1174,7 @@ GET /t
 --- response_body
 connected: 1
 ssl handshake: userdata
-sent http request: 59 bytes.
+sent http request: 61 bytes.
 received: HTTP/1.1 200 OK
 close: 1 nil
 
@@ -1185,8 +1185,8 @@ qr/^lua ssl save session: ([0-9A-F]+)
 lua ssl free session: ([0-9A-F]+)
 $/
 --- error_log
-lua ssl server name: "iscribblet.org"
-SSL: TLSv1.2, cipher: "ECDHE-RSA-RC4-SHA SSLv3
+lua ssl server name: "openresty.org"
+SSL: TLSv1.2, cipher: "ECDHE-RSA-AES128-GCM-SHA256
 --- no_error_log
 SSL reused session
 [error]
@@ -1199,7 +1199,7 @@ SSL reused session
 --- config
     server_tokens off;
     resolver $TEST_NGINX_RESOLVER ipv6=off;
-    lua_ssl_ciphers RC4-SHA;
+    lua_ssl_ciphers ECDHE-RSA-AES128-GCM-SHA256;
     location /t {
         #set $port 5000;
         set $port $TEST_NGINX_MEMCACHED_PORT;
@@ -1266,7 +1266,7 @@ lua ssl free session: ([0-9A-F]+)
 $/
 --- error_log
 lua ssl server name: "iscribblet.org"
-SSL: TLSv1.2, cipher: "RC4-SHA SSLv3
+SSL: TLSv1.2, cipher: "ECDHE-RSA-AES128-GCM-SHA256
 --- no_error_log
 SSL reused session
 [error]
@@ -1346,7 +1346,7 @@ lua ssl free session: ([0-9A-F]+)
 $/
 --- error_log
 lua ssl server name: "iscribblet.org"
-SSL: TLSv1, cipher: "ECDHE-RSA-RC4-SHA SSLv3
+SSL: TLSv1
 --- no_error_log
 SSL reused session
 [error]


src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -490,7 +490,6 @@ ngx_http_lua_ffi_ssl_set_ocsp_status_resp(ngx_http_request_t *r,
 
     dd("set ocsp resp: resp_len=%d", (int) resp_len);
     (void) SSL_set_tlsext_status_ocsp_resp(ssl_conn, p, resp_len);
-    ssl_conn->tlsext_status_expected = 1;
```
agentzh: This line is needed for at least OpenSSL 1.0.2 to work properly.

agentzh: We need to find the OpenSSL 1.1 equivalent for this.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -5270,9 +5268,8 @@ ngx_http_lua_ssl_free_session(lua_State *L)
 
     psession = lua_touserdata(L, 1);
     if (psession && *psession != NULL) {
-        ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
-                       "lua ssl free session: %p:%d", *psession,
-                       (*psession)->references);
```
agentzh: I think we need to find the OpenSSL 1.1 equivalent of `(*psession)->references` for this. This info is very useful.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1547,9 +1546,8 @@ ngx_http_lua_ssl_handshake_retval_handler(ngx_http_request_t *r,
     } else {
         *ud = ssl_session;
 
-       ngx_log_debug2(NGX_LOG_DEBUG_HTTP, c->log, 0,
-                      "lua ssl save session: %p:%d", ssl_session,
-                      ssl_session->references);
```
agentzh: Like below.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1281,9 +1281,8 @@ ngx_http_lua_socket_tcp_sslhandshake(lua_State *L)
                     return 2;
                 }
 
-                ngx_log_debug2(NGX_LOG_DEBUG_HTTP, c->log, 0,
-                               "lua ssl set session: %p:%d",
-                               *psession, (*psession)->references);
+                ngx_log_debug1(NGX_LOG_DEBUG_HTTP, c->log, 0,
+                               "lua ssl set session: %p", *psession);
```
agentzh: Like below.

agentzh:
@ghedo Thanks for the looking into this. But I think we should try to find the OpenSSL 1.1 equivalents for those things instead of simply killing them just to make the build pass :)

src/ngx_http_lua_ssl_ocsp.c
ghedo: Can you please be more specific on how this is needed?

The status_cb is called *after* the cert_cb (meaning that when the status_cb is called the ocsp response is already set), The logic that OpenSSL uses to decide whether to send the response is: if callback returns OK (which in out case is always) and if the OCSP response is not null, then send it. The logic is the same in v1.0.1, v1.0.2 and master:
https://github.com/openssl/openssl/blob/OpenSSL_1_0_1-stable/ssl/t1_lib.c#L1971
https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/ssl/t1_lib.c#L3077
https://github.com/openssl/openssl/blob/master/ssl/t1_lib.c#L2801

Of course I could be wrong, will need to do some more testing I guess (too bad the test suite doesn't support OCSP stapling).

src/ngx_http_lua_socket_tcp.c
ghedo: There is no 1.1 equivalent. As the tlsext_status_expected field above, these are intended to be internal APIs and applications are not supposed to use them. I don't see OpenSSL adding new interfaces for this any time soon TBH.

src/ngx_http_lua_ssl_ocsp.c
agentzh: @ghedo Yeah, better conduct some tests with both 1.0.2.x and 1.1.x. I cannot recall the details myself.

chipitsine:
you might also want to edit .travis.yml (and specify openssl version there), right now it uses 1.0.2

https://travis-ci.org/openresty/lua-nginx-module/jobs/136683814

src/ngx_http_lua_ssl_ocsp.c
kroeckx: I think this is a variable you shouldn't be touching directly.  If you think this "fixes" something, please file a bug against OpenSSL instead.

ctrochalakis:
The following patch is also need on later nginx-lua versions: https://github.com/wikimedia/operations-software-nginx/commit/e6785d912c992cae676593a8bd266e8c486b098d

src/ngx_http_lua_ssl_ocsp.c
kroeckx: So if this wasn't clear, the correct way to set this variable is having the status callback function return SSL_TLSEXT_ERR_OK, which seems it always does anyway. It looks like setting this variable here doesn't do anything useful in any version of OpenSSL and obviously doesn't work anymore in 1.1.0. So I believe this part of the patch is correct.

ghedo:
@ctrochalakis that's already in https://github.com/openresty/lua-nginx-module/pull/761/commits/1769df4649ed54b6cfef97de6cd7a79753be7a43

agentzh:
@ghedo Sorry for the delay on my side. Will you please rebase to the latest master branch and make the Travis CI tests pass in your branch here? Many thanks!

ghedo:
@agentzh rebased and fixed tests. I also tested locally with OpenSSL git master + my patch (https://github.com/openssl/openssl/pull/1588) and tests seem to be passing as well.

agentzh:
@ghedo Thanks for your quick updates! We'll look into this.

@doujiang24 Will you have a look at this PR for me? Many thanks!

t/143-ssl-session-fetch.t
```diff
@@ -1036,7 +1048,8 @@ qr/\S+:\d+: ssl fetch sess by lua is running!/s
         server_name   test.com;
         ssl_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
         ssl_certificate_key $TEST_NGINX_CERT_DIR/cert/test.key;
-        ssl_protocols SSLv3;
+        ssl_protocols TLSv1.2;
```
chipitsine: why do we need to specify ssl_protocol here ?

chipitsine:
if openssl-1.1.0 is supported now, we should add to travis-ci build matrix as well

agentzh:
Yeah, it's better to add OpenSSL 1.1 to the travis ci matrix (while still continue testing OpenSSL 1.0.2 as well).

@ghedo Will you edit `.travis.yml` to add OpenSSL 1.1 to the matrix as well? I think we need to add the OpenSSL patch in openssl/openssl#1588 to the openresty/openresty github repository first. Will you create a PR for that as well? Many thanks!

chipitsine:
I can add it later, after this PR is accepted

t/143-ssl-session-fetch.t
ghedo: Good point, since we use `ssl_session_tickets off` now there's no point in setting `ssl_protocols`. Removed now.

agentzh:
@chipitsine Cool, appreciated.

doujiang24:
I added OpenSSL 1.1.0c to the matrix in my local branch (https://github.com/doujiang24/lua-nginx-module/commits/openssl), and found some problem.

1) nginx 1.11.* still can not complied with OpenSSL 1.1 when without this patch, http://hg.nginx.org/nginx/rev/1891b2892b68
2) `t/129-ssl-socket.t` failed, https://travis-ci.org/doujiang24/lua-nginx-module/jobs/176400775

@ghedo Does this PR only works fine with openssl master branch?

PS: 
1> I added the nginx patch and @ghedo's https://github.com/openssl/openssl/pull/1588 in the openresty repo: https://github.com/doujiang24/openresty/commit/75b06a1fcf90d11d02166cd2dc45fec5cf0d12f2?diff=unified
2> a patch for openresty-devel-utils: https://github.com/doujiang24/nginx-devel-utils/commit/9d742f10698ee11461b127de226b36496482852b

chipitsine:
RC4 is deprecated, I guess special effort should be made if we want to compile openssl-1.1.0 with rc4 support

https://travis-ci.org/doujiang24/lua-nginx-module/jobs/176400778#L1138

actually, do we need it ?

ghedo:
I've been using NGINX 1.11.5 and it built for me.

Re: tests failing, I need to check that, but yeah, it seems that we hardcode RC4 so I'll need to fix that as well (unlikely that this actually passed when I was testing, but I don't remember and I might just have missed it).

Re: using non-master OpenSSL, I haven't tried with any other so I don't really know, I'll try that as well.

doujiang24:
@ghedo Yeah, I have checked NGINX 1.11.5 & 1.11.6 both have applied that patch.
But I didn't see any change log related to OpenSSL from 1.11.2 to 1.11.5, and I didn't check them yesterday, sorry for that.

doujiang24:
I think we can support openssl-1.1.0 with nginx-1.11.6.
Shall we create a new branch like `openssl-1.1.0` that the matrix can be below ?
```
  matrix:
    - NGINX_VERSION=1.9.15 OPENSSL_VER=1.0.2j
    - NGINX_VERSION=1.11.2 OPENSSL_VER=1.0.2j
    - NGINX_VERSION=1.11.6 OPENSSL_VER=1.0.2j
    - NGINX_VERSION=1.11.6 OPENSSL_VER=1.1.0c
```
@agentzh @ghedo What do you think about this ?

doujiang24:
@chipitsine Thanks for your point, I personally think it's fine without RC4 :)

ctrochalakis:
@doujiang24 Note that nginx 1.10.2 (stable) also has backported openssl 1.1.0 support (https://nginx.org/en/CHANGES-1.10), perhaps it can be also added to the matrix.

The new debian stable release (stretch) will actually ship nginx 1.10.2 build against OpenSSL 1.1.0.

chipitsine:
Actually, it is not about vanilla nginx. Test suite requires openresty patches. If 1.10.2 is not supported by openresty, so, we cannot do that

t/129-ssl-socket.t
```diff
@@ -349,13 +349,13 @@ sent http request: 59 bytes.
...
-qr/^lua ssl save session: ([0-9A-F]+):2
-lua ssl set session: \1:2
-lua ssl save session: \1:3
-lua ssl free session: \1:2
-lua ssl free session: \1:1
+qr/^lua ssl save session: ([0-9A-F]+)
+lua ssl set session: \1
+lua ssl save session: \1
+lua ssl free session: \1
+lua ssl free session: \1
```
agentzh: Alas, this test is weaker than before. We can no longer track the reference count :(

t/129-ssl-socket.t
```diff
@@ -1569,14 +1569,14 @@ ssl handshake: userdata
...
-lua ssl save session: \1:4
-lua ssl free session: \1:4
-lua ssl free session: \1:3
-lua ssl free session: \1:2
+qr/^lua ssl save session: ([0-9A-F]+)
+lua ssl save session: \1
+lua ssl save session: \1
+lua ssl free session: \1
+lua ssl free session: \1
+lua ssl free session: \1
```
agentzh: Ditto.

agentzh:
It looks good to me though I'm not very happy with the lack of capability of tracking session reference counts.

Guys, any objections to merging this PR?

doujiang24:
@agentzh LGTM:)

agentzh:
I've noticed that the `t/129-ssl-socket.t` test file is still failing with OpenSSL 1.1.0c. We need to avoid using the RC4-SHA cipher in those tests. Patches welcome!

ghedo:
Ok, so, I just pushed a commit to fix the RC4 tests which should work with both 1.0.2 and 1.1.0 now. However I noticed that several other tests are failing with OpenSSL 1.1.0 (namely tests 9, 10, 17 and 23), but I haven;t looked into that yet (test 17 uses `SSLv2` so that's probably the reason why it fails, but I don't how to fix it).

doujiang24:
@ghedo which version of openssl do you based on?
143-ssl-session-fetch.t still failed even with your patch for openssl-1.1.0c (openssl/openssl#1588)
https://travis-ci.org/openresty/lua-nginx-module/jobs/179391221#L1274

doujiang24:
@ghedo I have fixed test 17 & 23 in https://github.com/openresty/lua-nginx-module/pull/922
Seems OpenSSL 1.1 return the different error code against 1.0

kroeckx:
On Mon, Nov 28, 2016 at 06:54:35PM -0800, doujiang wrote:
> @ghedo I have fixed test 17 & 23 in https://github.com/openresty/lua-nginx-module/pull/922
> Seems OpenSSL 1.1 return the different error code against 1.0

You might want to see https://github.com/openssl/openssl/issues/1903, and the documentation change in the pull requests.

defanator:
Thanks for everyone's efforts on this! Upcoming Debian 9 "stretch" is using openssl 1.1.0 out of the box (and more systems will do, undoubtedly).

neilstuartcraig:
Hi all
Is there anything preventing this being merged now? I'd really love to be able to compile against openssl 1.1 :-)

Cheers

agentzh:
@neilstuartcraig Some tests in this module's test suite are still failing with OpenSSL 1.1, as per @doujiang24.

agentzh:
@neilstuartcraig Being able to compile does not necessarily mean working correctly. If it does not work properly, I'd rather it does not compile. Otherwise users will report runtime issues that are very hard to debug.

agentzh:
@neilstuartcraig The OpenSSL 1.0.2 series is still updating with new security fixes and etc. So supporting the new 1.1 series is not of a priority.

agentzh:
@jirutka If you really want it happen sooner, feel free to pick up @doujiang24's pull request and make all the tests pass with OpenSSL 1.1. It's that simple.

jirutka:
@agentzh I’m not saying that it’s simple. I didn’t mean 👎 as a criticism. Also I don’t really care about OpenSSL, but LibreSSL. I’m currently trying to untangle irrelevant external dependencies from lua-nginx-module tests, to run them during build of Alpine package.

chipitsine:
I see the openssl-1.1.0 roadmap as a following

nginx itself support openssl 1.1.0 since 1.11.7, while current openresty bundle is 1.11.2
lua-nginx-module is core part of openresty bundle and many tests require openresty patches

so, I think we should upgrade openresty to ... 1.13 ?
I can spend some time on that if I can be sure it will be merged

currently, I made PR https://github.com/openresty/no-pool-nginx/pull/5 , which did not get any attention
(openresty patches for 1.11.10 still "work in progress", some tests fail)

openssl-1.1.0 is released with Fedora and Arch, so I think it's time to move toward openssl-1.1.0

agentzh:
@chipitsine It's not about the NGINX core, actually. It's about incompatible changes in OpenSSL 1.1.0's own API and behaviors. See the detailed discussions and patches in this thread for details.

chipitsine:
I followed the thread of course. I was working on patches to nginx 1.11.10 and I was rinning tests against openssl-1.0.2 (some of them failed)

I suggest to move to nginx-1.13.0 first
Then, to work on openssl-1.1.0 patches, because nginx-1.11.2 does not support openssl-1.1.0 natively

agentzh:
@chipitsine NGINX core will indeed be upgraded in OpenResty. It is stated in OpenResty 1.11.2.3's announcement. But as I said, it's not the showstopper we are talking about in this thread.

agentzh:
@chipitsine We'll still stick with OpenSSL 1.0.2 series even after we move to the 1.13.x cores if the real showstopper is not fully resolved. BTW, IIRC, ngx_lua already found bugs in OpenSSL 1.1.0.

johnny564:
Debian 9 "stretch" released as new stable debian version. With native openssl 1.1.0, build of openresty fails.

agentzh:
@johnny564 FWIW, we do provide official OpenResty pre-built packages for Debian stretch:

https://openresty.org/en/linux-packages.html

tpimh:
@johnny564 Debian 9 still provides older version of OpenSSL, you can just install `libssl1.0-dev` instead of `libssl-dev` and it will compile just fine.

mtorromeo:
@tpimh except that nginx should be compiled against that same version of openssl or you can't build this as a dynamic module for debian's packaged nginx

tpimh:
@mtorromeo I'm talking about building openresty from source according to the [instructions](https://openresty.org/en/installation.html#building-from-source).

mtorromeo:
I know but I'm pointing out that this is only one module and you shouldn't be forced to use it like that.

agentzh:
Superseded by #1213, which is already merged into master :) Thanks!

PR #768, opened 2016-05-15
fixed an error (we could exit 0 on failure), also some cleanup

chipitsine:


agentzh:
@chipitsine The commit log messages should have the prefix "travis-ci: " to be more specific about the context :)

agentzh:
@chipitsine I've fixed the commit log messages myself and merged into master. Thanks.

PR #769, opened 2016-05-15
enable clang

chipitsine:
I use PATH=$PATH because clang is located at /usr/local/clang-3.5.0/bin and without specifiyng PATH it is not found after sudo


agentzh:
@chipitsine Please avoid merge commit in such simple PRs so that we can have a nicer git history. Thank you :) Use `git rebase` in your own branches whenever possible.

agentzh:
Applied your patch to git master.

PR #790, opened 2016-06-04
tests: Add regression test for buffer split in block parsing (#687)

pushrax:
It looks like #687 was fixed. I added the regression test I had previously included in #688 and confirmed that it fails without 70f708071534857d2b60ad2998d8d05d91483a0d and passes with.

Can you briefly explain what the patch does? To me it doesn't look like it fixes the issue generally, but maybe there's a reason why that doesn't matter?


agentzh:
@pushrax Thanks for the test case! Merged into master.

PR #803, opened 2016-06-18
travis-ci: refactor package management, i.e. use travis-ci "apt" plugin, cache packages

chipitsine:


.travis.yml
```diff
@@ -107,6 +115,7 @@ script:
   - cd ..
   - export PATH=$PWD/work/nginx/sbin:$PWD/nginx-devel-utils:$PATH
   - export NGX_BUILD_CC=$CC
+  - if [ "$CC" == "clang" ]; then export CFLAGS="-Qunused-arguments"; fi
```
agentzh: Will you document why we need this? Looks strange.

.travis.yml
```diff
@@ -49,10 +60,8 @@ services:
  - memcache
  - redis-server
 
-before_install:
-  - sudo apt-get install -y axel cpanminus libtest-base-perl libtext-diff-perl liburi-perl libwww-perl libtest-longstring-perl liblist-moreutils-perl libgd-dev
-
 install:
+  - mkdir ~/bin && ln -s $(which ccache) ~/bin/$CC && export PATH=$HOME/bin:$PATH
```
chipitsine: using "ln" is recommended way, it is documented in "man ccache", anyway, we save more on openssl, than nginx

.travis.yml
chipitsine: I've got several "unused option -I...." like described here http://stackoverflow.com/questions/13238511/

agentzh: @chipitsine Will you document this as a code comment around this line, maybe?

.travis.yml
agentzh: @chipitsine I reread the manpage of ccache and I failed to see any propounding reason of using the masquerading approach other than convenience when ccache'ing all the compilation in the current session. Well, given that Travis CI is using a VM approach and invasive settings are common practice, so maybe it also makes things more convenient if not really elegant.

chipitsine: yes, if we will want to get rid of ccache, we can just remove that line

agentzh: @chipitsine I'm fine with that.

chipitsine:
it seems ccache is not stable in case of openssl (sometimes it helps to remove "-j" option in make)
I'll need time to investigate it

agentzh:
@chipitsine I'm not feeling comfortable with removing `-jN` from openssl's `make` command. It looks like something wrong with ccache. And if ccache is not really trust worthy, there is a potential risk of using it in our Travis CI builds.

chipitsine:
I will live just "apt-get" cache for now. Will remove ccache (and will investigate it later)

agentzh:
@chipitsine Sounds good to me :)

agentzh:
Merged to master. Thanks!

PR #812, opened 2016-07-01
feature: added pure C API for setting upstream request timeouts

agentzh:
This supersedes PR #700.

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -33,6 +33,10 @@ struct ngx_http_lua_balancer_peer_data_s {
     in_port_t                           port;
 
     int                                 last_peer_state;
+
+#if !(HAVE_UPSTREAM_TIMEOUT_FIELDS)
+    unsigned                            cloned_upstream_conf;  /* :1 */
+#endif
 };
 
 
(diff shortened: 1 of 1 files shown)


agentzh:
This PR can optionally take advantage of this nginx core patch: https://github.com/openresty/openresty/blob/master/patches/nginx-1.9.15-upstream_timeout_fields.patch

Still we can work with the stock nginx core, just requiring a bit more memory per request.

agentzh:
@xbsura @toruneko Please review this PR. Thanks!

src/ngx_http_lua_balancer.c
```diff
@@ -537,6 +541,101 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+            return NGX_ERROR;
+        }
+
+        ngx_memcpy(ucf, u->conf, sizeof(ngx_http_upstream_conf_t));
+
+        u->conf = ucf;
+        bp->cloned_upstream_conf = 1;
+
+    } else {
+        ucf = u->conf;
```
toruneko: hah good job, it means we can invoke this function twice，but  only alloc memory one times。

toruneko:
I review this PR, and not find any problem. @xbsura ？

src/ngx_http_lua_balancer.c
agentzh: @toruneko Yup. That's right :)

guanglinlv: @agentzh , someone maybe want to set different timeout for different peer when retrying the next one. how about  keep it left to the user?

xbsura:
oh, good
meeting pre work, not in work, sorry

toruneko:
I find a problem that if we do not use this api to set timeouts, and the u->*_timeout will be blank if patched    nginx-1.9.15-upstream_timeout_fields.patch.
we should patches ngx_http_proxy_handler function after 
`u->conf = &plcf->upstream; ` like
```
u->conf = &plcf->upstream;
u->read_timeout = u->conf->read_timeout;
u->send_timeout = u->conf->send_timeout;
u->connect_timeout = u->conf->connect_timeout;
```

agentzh:
@toruneko It's already covered in OpenResty's nginx core patch here:

https://github.com/openresty/openresty/blob/master/patches/nginx-1.9.15-upstream_timeout_fields.patch#L17

BTW, patching ngx_proxy for this is wrong since there are unlimited number of other upstream modules out there, including standard upstream modules like ngx_fastcgi and ngx_uwsgi, not to mention 3rd-party upstream modules like ngx_memc and ngx_redis2.

agentzh:
This PR is already merged into master. If you find further problems, please create separate github issues or PRs. Thank you.

PR #827, opened 2016-07-31
doc: ngx.worker.count() is available in init_worker_by_lua context

GUI:
The `ngx.worker.count` documentation was missing `init_worker_by_lua` as a valid context. This updates the documentation and adds a test to demonstrate that this method does in fact work in the `init_worker_by_lua` context.


agentzh:
@GUI Merged with minor edits. Many thanks!

PR #831, opened 2016-08-04
bugfix: incr on expired list type key will cause shdict memory leaks

doujiang24:


agentzh:
Merged manually. Thanks!

PR #840, opened 2016-08-18
bugfix:avoid the package.preload module not found while code cache off

rainingmaster:
I've got a tiny bug in  ngx_http_lua_new_state : in [https://github.com/rainingmaster/lua-nginx-module/blob/master/src/ngx_http_lua_util.c#L218](url) , there is not copy the packge.preload from parent_vm to the new one.

My nginx.conf:
`...
lua_code_cache off;
...`

My ngx_modules.c(ngx_http_lua_upstream_module behind ngx_http_lua_module)
`
&ngx_http_lua_module, 
&ngx_http_lua_upstream_module,
`

My code.lua
`local upstream = require "ngx.upstream"`

And it will report the 

> error：lua entry thread aborted: runtime error: code.lua:1: module 'ngx.upstream' not found:

### DIFF
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -47,12 +47,11 @@ ngx_http_lua_add_package_preload(ngx_conf_t *cf, const char *package,
         lua_pushcfunction(L, func);
         lua_setfield(L, -2, package);
         lua_pop(L, 2);
-
-        return NGX_OK;
     }
 
     /* L == NULL */
 
+    /* still store for a new vm init */
     if (lmcf->preload_hooks == NULL) {
         lmcf->preload_hooks =
             ngx_array_create(cf->pool, 4,


agentzh:
@rainingmaster Is it possible to add a test case for it?

rainingmaster:
ok, and I has moved the `ngx_http_lua_upstream_module` to the back of the `ngx_http_lua_module` in `ngx_modules.c` , to ensure that the first init ngx_http_lua_module, then init ngx_http_lua_upstream_module. And  execute [the code in if](https://github.com/rainingmaster/lua-nginx-module/blob/master/src/ngx_http_lua_api.c#L44) .

my nginx.conf:
```
http {
    server {
        listen 8080;
        
        location = / {
            lua_code_cache off;
            content_by_lua_block {
                local concat = table.concat
                local upstream = require "ngx.upstream"
                local get_upstreams = upstream.get_upstreams

                local us = get_upstreams()
            }
        }
    }
}
```

rainingmaster:
@agentzh , I find that lua-nginx-module is [HTTP_AUX_FILTER_MODULES](https://github.com/openresty/lua-nginx-module/blob/master/config#L530)， and lua-upstream-nginx-module
is [HTTP_MODULES](https://github.com/openresty/lua-upstream-nginx-module/blob/master/config#L11), so lua-upstream-nginx-module must be in front of lua-nginx-module.
Sorry,  I cant get a simple test case :(

agentzh:
@doujiang24 What do you think of this patch? ;)

src/ngx_http_lua_api.c
```diff
@@ -47,12 +47,11 @@ ngx_http_lua_add_package_preload(ngx_conf_t *cf, const char *package,
         lua_pushcfunction(L, func);
         lua_setfield(L, -2, package);
         lua_pop(L, 2);
-
-        return NGX_OK;
```
doujiang24: Seems we can still `return NGX_OK` when `lua_cache_code` is `on`?

doujiang24:
@agentzh This seems reasonable, but we still need to build a test case for that, I'll try it also.
@rainingmaster will you build another nginx C module that will added after `ngx_http_lua_module`?

rainingmaster:
I've push a new pull request in [here](https://github.com/openresty/lua-nginx-module/pull/912)

PR #844, opened 2016-08-22
bugfix: skip setting multi-value headers for bad requests

es:
When setting a multi-value header during bad requests, the value isn't added into the header's array. Despite failing to be set, the size of the array is [still incremented](https://github.com/openresty/lua-nginx-module/blob/7d242ff79a63bae15c8f5f68999dfe5f4da7cf67/src/ngx_http_lua_headers_in.c#L624). Later, if Nginx attempts to [iterate over the array](https://github.com/nginx/nginx/blob/8d0b75f716f3ef2c33d277f70f45fbc1e0ca31fe/src/http/ngx_http_parse.c#L1922-L1931), a segfault will occur.

e21d9b5 established a pattern of silently ignoring the header being set. I've moved the check for a bad request into `ngx_http_lua_set_input_header` so that setting a header value is ignored for both types of headers.

r: @agentzh

### DIFF
--- a/src/ngx_http_lua_headers_in.c
+++ b/src/ngx_http_lua_headers_in.c
@@ -263,11 +263,6 @@ retry:
 
 new_header:
 
-    if (r->headers_in.headers.last == NULL) {
-        /* must be 400 bad request */
-        return NGX_OK;
-    }
-
     h = ngx_list_push(&r->headers_in.headers);
 
     if (h == NULL) {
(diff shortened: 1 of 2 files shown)


pushrax:
LGTM

agentzh:
Looks fine. I'll merge this after OpenResty 1.11.2.1 formal is out. Thanks!

es:
@agentzh looks like [1.11.2.1 is out](https://github.com/openresty/openresty/releases/tag/v1.11.2.1). Should be good to merge now 😸

agentzh:
Will look into this soon.

agentzh:
Merged with minor edits in your commit log message. Thanks!

PR #850, opened 2016-08-25
feature: add a new api tcpsock:settimeouts(connect_timeout, send_timeout, read_timeout)

doujiang24:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -946,7 +946,7 @@ To run specific test files:
 
 To run a specific test block in a particular test file, add the line `--- ONLY` to the test block you want to run, and then use the `prove` utility to run that `.t` file.
 
-There are also various testing modes based on mockeagain, valgrind, and etc. Refer to the [Test::Nginx documentation](http://search.cpan.org/perldoc?Test::Nginx) for more details for various advanced testing modes. See also the test reports for the Nginx test cluster running on Amazon EC2: <http://qa.openresty.org>.
+There are also various testing modes based on mockeagain, valgrind, and etc. Refer to the [Test::Nginx documentation](http://search.cpan.org/perldoc?Test::Nginx) for more details for various advanced testing modes. See also the test reports for the Nginx test cluster running on Amazon EC2: <http://qa.openresty.org.>
 
 [Back to TOC](#table-of-contents)
 
@@ -2654,7 +2654,7 @@ lua_socket_connect_timeout
 
 **context:** *http, server, location*
 
-This directive controls the default timeout value used in TCP/unix-domain socket object's [connect](#tcpsockconnect) method and can be overridden by the [settimeout](#tcpsocksettimeout) method.
+This directive controls the default timeout value used in TCP/unix-domain socket object's [connect](#tcpsockconnect) method and can be overridden by the [settimeout](#tcpsocksettimeout) or [settimeouts](#tcpsocksettimeouts) method.
 
 The `<time>` argument can be an integer, with an optional time unit, like `s` (second), `ms` (millisecond), `m` (minute). The default time unit is `s`, i.e., "second". The default setting is `60s`.
 
@@ -2671,7 +2671,7 @@ lua_socket_send_timeout
 
 **context:** *http, server, location*
 
-Controls the default timeout value used in TCP/unix-domain socket object's [send](#tcpsocksend) method and can be overridden by the [settimeout](#tcpsocksettimeout) method.
+Controls the default timeout value used in TCP/unix-domain socket object's [send](#tcpsocksend) method and can be overridden by the [settimeout](#tcpsocksettimeout) or [settimeouts](#tcpsocksettimeouts) method.
 
 The `<time>` argument can be an integer, with an optional time unit, like `s` (second), `ms` (millisecond), `m` (minute). The default time unit is `s`, i.e., "second". The default setting is `60s`.
 
@@ -2703,7 +2703,7 @@ lua_socket_read_timeout
 
 **phase:** *depends on usage*
 
-This directive controls the default timeout value used in TCP/unix-domain socket object's [receive](#tcpsockreceive) method and iterator functions returned by the [receiveuntil](#tcpsockreceiveuntil) method. This setting can be overridden by the [settimeout](#tcpsocksettimeout) method.
+This directive controls the default timeout value used in TCP/unix-domain socket object's [receive](#tcpsockreceive) method and iterator functions returned by the [receiveuntil](#tcpsockreceiveuntil) method. This setting can be overridden by the [settimeout](#tcpsocksettimeout) or [settimeouts](#tcpsocksettimeouts) method.
 
 The `<time>` argument can be an integer, with an optional time unit, like `s` (second), `ms` (millisecond), `m` (minute). The default time unit is `s`, i.e., "second". The default setting is `60s`.
 
@@ -3113,6 +3113,7 @@ Nginx API for Lua
 * [tcpsock:receiveuntil](#tcpsockreceiveuntil)
 * [tcpsock:close](#tcpsockclose)
 * [tcpsock:settimeout](#tcpsocksettimeout)
+* [tcpsock:settimeouts](#tcpsocksettimeouts)
 * [tcpsock:setoption](#tcpsocksetoption)
 * [tcpsock:setkeepalive](#tcpsocksetkeepalive)
 * [tcpsock:getreusedtimes](#tcpsockgetreusedtimes)
@@ -6669,6 +6670,7 @@ Creates and returns a TCP or stream-oriented unix domain socket object (also kno
 * [receive](#tcpsockreceive)
 * [close](#tcpsockclose)
 * [settimeout](#tcpsocksettimeout)
+* [settimeouts](#tcpsocksettimeouts)
 * [setoption](#tcpsocksetoption)
 * [receiveuntil](#tcpsockreceiveuntil)
 * [setkeepalive](#tcpsocksetkeepalive)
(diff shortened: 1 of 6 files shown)


thibaultcha:
@doujiang24 Very nice then!

---

One concern: it is too bad that ngx_lua is often so strict about the arguments being received from the stack. In plain Lua, additional arguments to a function are often **discarded**, whether ngx_lua **errors out** on such calls. This limits possibilities in terms of abstractions we can create on top of ngx_lua, and can make backwards compatibility much harder for lua-resty libraries maintainers.

For example, given that `sock:settimeout()` would have been implemented like so:
```c
lua_settop(L, 2); // just discard additional arguments on the stack

// and check the given arguments (this is just an example,
// which could probably be done in a better way
if (!lua_isuserdata(L, 1)) {
    luaL_argerror(L, 1, "not a cosocket");

} else if (!lua_isnumber(L, 2) {
    luaL_argerror(L, 2, "expected timeout to be a number");
}
```

Instead of:
```c
n = lua_gettop(L);

if (n != 2) {
    return luaL_error(L, "ngx.socket settimout: expecting at least 2 "
                      "arguments (including the object) but seen %d",
                      lua_gettop(L));
}
```

(**side note**: the error message here is wrong, it should be:
```diff
-        return luaL_error(L, "ngx.socket settimout: expecting at least 2 arguments"
+        return luaL_error(L, "ngx.socket settimout: expecting 2 arguments "
```

Then this PR could have been implemented in a nice backwards-compatible way for all the `lua-resty-*` libraries currently using the cosocket API:
```
sock:settimeout(connect_timeout, send_timeout?, read_timeout?)
```

All of the libraries could add additional `send_timeout` and `read_timeout` arguments, that are disregarded for older versions of ngx_lua, and taken into consideration for ngx_lua versions supporting multiple timeouts.

I might not be aware of the reasons why ngx_lua is so strict about the arguments received from the stack, but I would be curious to know :)

doujiang24:
@thibaultcha Thanks :)

I was also confused about that ngx_lua is often so strict about the arguments.
And I figured out one thing by myself: it will make the API more safer for growth, for example:
(Maybe @agentzh will give more reason?)

API growth:
```lua
ngx.shdict.DICT.incr(key, value) -- old version
ngx.shdict.DICT.incr(key, value, init?) -- new version
```
then, the code will behave in different way when we just upgrade OpenResty.
```lua
local foo = ngx.shared.foo
foo:incr("key", 1, 100)
```

agentzh:
Yeah, strict argument checks were for forward API compatibility. In new Lua implementations of the API in lua-resty-core, however, I've been relaxing it a little bit due to performance considerations. It's safer for the user to never feed extra arguments into the API functions since it may break in the future.

README.markdown
```diff
@@ -6997,6 +6999,24 @@ This feature was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+tcpsock:settimeouts
+-------------------
+**syntax:** *tcpsock:settimeouts(connect_timeout, send_timeout, read_timeout)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, ngx.timer.&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;*
+
+Set the timeout value(connect, send and read) in milliseconds for subsequent socket operations ([connect](#tcpsockconnect), [send](#tcpsocksend), [receive](#tcpsockreceive), and iterators returned from [receiveuntil](#tcpsockreceiveuntil)).
```
agentzh: "Set" => "Sets"

README.markdown
```diff
@@ -6997,6 +6999,24 @@ This feature was first introduced in the `v0.5.0rc1` release.
...
+
+Settings done by this method takes priority over those config directives, i.e., [lua_socket_connect_timeout](#lua_socket_connect_timeout), [lua_socket_send_timeout](#lua_socket_send_timeout), and [lua_socket_read_timeout](#lua_socket_read_timeout).
+
+You are recommended to use [settimeouts](#tcpsocksettimeouts) instead [settimeout](#tcpsocksettimeout).
```
agentzh: "instead" => "instead of".

src/ngx_http_lua_socket_tcp.c
```diff
@@ -137,8 +138,10 @@ static void ngx_http_lua_socket_tcp_close_connection(ngx_connection_t *c);
 
 enum {
     SOCKET_CTX_INDEX = 1,
-    SOCKET_TIMEOUT_INDEX = 2,
-    SOCKET_KEY_INDEX = 3
+    SOCKET_CONNECT_TIMEOUT_INDEX = 2,
+    SOCKET_SEND_TIMEOUT_INDEX = 3,
+    SOCKET_READ_TIMEOUT_INDEX = 4,
+    SOCKET_KEY_INDEX = 5
```
agentzh: I think it's better not to change existing indexes (like `SOCKET_KEY_INDEX`).

src/ngx_http_lua_socket_tcp.c
```diff
@@ -576,21 +588,37 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
 
     dd("lua peer connection log: %p", pc->log);
 
-    lua_rawgeti(L, 1, SOCKET_TIMEOUT_INDEX);
-    timeout = (ngx_int_t) lua_tointeger(L, -1);
-    lua_pop(L, 1);
+    lua_rawgeti(L, 1, SOCKET_CONNECT_TIMEOUT_INDEX);
+    lua_rawgeti(L, 1, SOCKET_SEND_TIMEOUT_INDEX);
+    lua_rawgeti(L, 1, SOCKET_READ_TIMEOUT_INDEX);
```
agentzh: Because we are adding more elements to the object table, maybe we should also increase the initial array-part size of the table?

doujiang24: Yes, already increased to 5 since we have recommended to use `settimeouts`.

doujiang24:
@agentzh Thanks for your review :)
All fixed & rebased master.

src/ngx_http_lua_socket_tcp.c
agentzh: @doujiang24 I'm not talking about the metatable, but the object table itself. Yes, I saw you increased the initial size of the metatables but I didn't see changes in the object table. Will you confirm that?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -389,7 +401,7 @@ ngx_http_lua_socket_tcp(lua_State *L)
                                | NGX_HTTP_LUA_CONTEXT_SSL_CERT
                                | NGX_HTTP_LUA_CONTEXT_SSL_SESS_FETCH);
 
-    lua_createtable(L, 3 /* narr */, 1 /* nrec */);
+    lua_createtable(L, 5 /* narr */, 1 /* nrec */);
```
agentzh: Okay, never mind, the initial size of the object table itself is already updated :)

src/ngx_http_lua_socket_tcp.c
agentzh: Never mind now :)

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2698,6 +2730,60 @@ ngx_http_lua_socket_tcp_settimeout(lua_State *L)
...
+{
+    int                     n;
+    ngx_int_t               connect_timeout, send_timeout, read_timeout;
+
+    ngx_http_lua_socket_tcp_upstream_t  *u;
+
+    n = lua_gettop(L);
+
+    if (n != 4) {
+        return luaL_error(L, "ngx.socket settimout: expecting at least 4 "
```
agentzh: Let's remove "at least ", which is wrong.

t/146-tcp-socket-settimeouts.t
```diff
@@ -0,0 +1,516 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
```
agentzh: @doujiang24 Please rename it on your side :) Thanks!

src/ngx_http_lua_socket_tcp.c
agentzh: Let's also fix the `tcp_settimeout` function from which you copy and paste :)

t/146-tcp-socket-settimeouts.t
```diff
@@ -0,0 +1,516 @@
...
+
+#log_level 'warn';
+log_level 'debug';
+
+no_long_string();
+#no_diff();
+run_tests();
+
+__DATA__
+
```
agentzh: Will you please use larger timeout values to ensure these tests also pass in the valgrind testing mode? Some tests are failing in the valgrind mode on my side :) Thanks!

doujiang24:
@agentzh Thanks, all updated :)

README.markdown
```diff
@@ -946,7 +946,7 @@ To run specific test files:
 
 To run a specific test block in a particular test file, add the line `--- ONLY` to the test block you want to run, and then use the `prove` utility to run that `.t` file.
 
-There are also various testing modes based on mockeagain, valgrind, and etc. Refer to the [Test::Nginx documentation](http://search.cpan.org/perldoc?Test::Nginx) for more details for various advanced testing modes. See also the test reports for the Nginx test cluster running on Amazon EC2: <http://qa.openresty.org>.
+There are also various testing modes based on mockeagain, valgrind, and etc. Refer to the [Test::Nginx documentation](http://search.cpan.org/perldoc?Test::Nginx) for more details for various advanced testing modes. See also the test reports for the Nginx test cluster running on Amazon EC2: <http://qa.openresty.org.>
```
agentzh: This is a regression. I think you need to update your local `nginx-devel-utils` via `git pull` :)

agentzh:
Merged with minor fixes. Thanks!

PR #853, opened 2016-08-31
bypass next_upstream_tries check when its value is 0

kingluo:

When the value of next_upstream_tries is 0, it means the limitation is turned off, then no need to check it.

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -682,7 +682,7 @@ ngx_http_lua_ffi_balancer_set_more_tries(ngx_http_request_t *r,
 #if (nginx_version >= 1007005)
     max_tries = r->upstream->conf->next_upstream_tries;
 
-    if (bp->total_tries + count > max_tries) {
+    if (max_tries && bp->total_tries + count > max_tries) {
         count = max_tries - bp->total_tries;
         *err = "reduced tries due to limit";
 


agentzh:
@kingluo Your patch looks like a micro optimization and it just saves a couple of memory accesses and an addition operation?

kingluo:
No, it's not a performance optimization.
In nginx, if next_upstream_tries is 0 (by default), it means no limit to set tries.
But the lua module does not allow to set more tries, because (bp->total_tries + count >0) is always true.

agentzh:
@kingluo Gotcha. Will you add corresponding test cases for this fix? Thanks!

kingluo:
How about this test case?

```

#use Test::Nginx::Socket 'no_plan';
use Test::Nginx::Socket::Lua 'no_plan';

#worker_connections(1014);
#master_on();
#workers(2);
#log_level('warn');

#repeat_each(2);

#plan tests => repeat_each() * (blocks() * 2);

#no_diff();
#no_long_string();
run_tests();

__DATA__

=== TEST 1: set_more_tries error
--- http_config
    lua_package_path "/home/vagrant/lualib/?.lua;;";
    upstream backend {
        server 0.0.0.1;
        balancer_by_lua_block {
            local _, err = balancer.set_more_tries(30)
            if err then
                print(err)
            end
        }
    }
--- config
    location = /t {
        proxy_next_upstream_tries 0;
        proxy_pass http://backend;
    }
--- request
    GET /t
--- error_code: 502
--- error_log
reduced tries due to limit
```

I still have two questions about this test case:
* maybe I should use `no_error_log` to ensure no error after bugfix?
* how to define lua library path relative to the test directory? how to place external libraries?

agentzh:
@kingluo 

> maybe I should use no_error_log to ensure no error after bugfix?

Yes. Please ensure that your test case would fail before your fix is applied. Otherwise it's not covering the fix :)

> how to define lua library path relative to the test directory? how to place external libraries?

The test case should be in the lua-resty-core repository. All the tests involved with `ngx.balancer` are there.

Thanks!

Tieske:
+1 for this.

Just got bitten by exactly this issue. Worked around it by ;

````
next_upstream_tries = 9999;
````

but that's just too ugly

agentzh:
I can confirm that this issue has been fixed in the latest master branch.

Tieske:
👍

PR #855, opened 2016-09-01
A typo in the doc

lution:
I think this might be a typo in the doc.

`environment variable` instead of `environ variable`


thibaultcha:
I think you should also re-generate the README.md with the [wiki2markdown.pl](https://github.com/openresty/nginx-devel-utils/blob/master/wiki2markdown.pl) tool in your patch.

agentzh:
I think this patch is already applied.

PR #861, opened 2016-09-03
bsd style receive method for ngx.socket.tcp

brg-liuwei:
@agentzh bsd style receive for ngx.socket.tcp in lua-nginx-module, and add a comment in the document. Maybe I need to add more test cases, yes?

@mike07026 bsd style receive for cosocket in http has accomplished. Good day.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6841,6 +6841,7 @@ If a number-like argument is specified (including strings that look like numbers
 If a non-number-like string argument is specified, then it is interpreted as a "pattern". The following patterns are supported:
 
 * `'*a'`: reads from the socket until the connection is closed. No end-of-line translation is performed;
+* `'*b'`: the BSD style receive, return anything once socket receives;
 * `'*l'`: reads a line of text from the socket. The line is terminated by a `Line Feed` (LF) character (ASCII 10), optionally preceded by a `Carriage Return` (CR) character (ASCII 13). The CR and LF characters are not included in the returned line. In fact, all CR characters are ignored by the pattern.
 
 If no argument is specified, then it is assumed to be the pattern `'*l'`, that is, the line reading pattern.
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -83,6 +83,7 @@ static int ngx_http_lua_socket_read_error_retval_handler(ngx_http_request_t *r,
 static int ngx_http_lua_socket_write_error_retval_handler(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u, lua_State *L);
 static ngx_int_t ngx_http_lua_socket_read_all(void *data, ssize_t bytes);
+static ngx_int_t ngx_http_lua_socket_read_bsd(void *data, ssize_t bytes);
 static ngx_int_t ngx_http_lua_socket_read_until(void *data, ssize_t bytes);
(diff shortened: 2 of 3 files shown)


t/146-tcp-socket-receive-bsd-style.t
```diff
@@ -0,0 +1,101 @@
...
+                    "CCCCC",
+                },
+                {
+                    "www",
+                    "xxx",
+                    "yyy",
+                    "zzz",
+                },
+            }
+            local i = math.random(2)
```
agentzh: Let's test the simplest thing that could possibly work. Complicated test cases are hard to debug when they fail. It took me a while to understand what this test case is trying to do. Better make it as straight forward as possible. No tricks please. Thanks!

README.markdown
```diff
@@ -6841,6 +6841,7 @@ If a number-like argument is specified (including strings that look like numbers
 If a non-number-like string argument is specified, then it is interpreted as a "pattern". The following patterns are supported:
 
 * `'*a'`: reads from the socket until the connection is closed. No end-of-line translation is performed;
+* `'*b'`: the BSD style receive, return anything once socket receives;
```
agentzh: I think a bit more about this. Maybe rename it to `*p` to mean receiving a "packet" makes more sense? Thoughts?

mike07026: @agentzh I think TCP has no "packet" concept unless you have defined a protocol, so "*p" maybe  not good enough. Consider that the receive behavior here is just returning anything the socket got, I suggest to  use `**` (anything) or `*s` (stream) instead.

t/146-tcp-socket-receive-bsd-style.t
```diff
@@ -0,0 +1,94 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use Test::Nginx::Socket::Lua 'no_plan';
```
spacewander: Better to avoid using `no_plan` in the lua-nginx-module test suite.
What need to do is just copy the run test number from `Bad plan` error message.

t/146-tcp-socket-receive-bsd-style.t
```diff
@@ -0,0 +1,94 @@
...
+GET /t
+--- response_body
+1
+22
+hello world
+--- no_error_log
+[error]
+--- error_log
+lua tcp socket read bsd
+
```
spacewander: The trailing line here is unnecessary.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1943,6 +1948,38 @@ ngx_http_lua_socket_read_all(void *data, ssize_t bytes)
...
+
+    b = &u->buffer;
+
+    dd("already read: %p: %.*s", u->buf_in,
+       (int) (u->buf_in->buf->last - u->buf_in->buf->pos),
+       u->buf_in->buf->pos);
+
+    dd("data read: %.*s", (int) bytes, b->pos);
+
+    u->buf_in->buf->last = ngx_cpymem(u->buf_in->buf->last, b->pos, bytes);
```
spacewander: `u->buf_in->buf` and `b` actually point to the same buffer. So there is no need to copy the content. Just increase the pointer position is enough.

README.markdown
agentzh: @mike07026 I think maybe it's better to do a `bsdrecv()` method instead.

Issue #860, opened 2016-09-02
tcp connect fails with nil as final option


Tieske:
This code;

```lua
local function myConnect(sock, host, port, opts)
  return sock:connect(host, port, opts)   --> args = 3, `opts = nil`
end

local ok, err = myConnect(myTcpSocket, "www.google.com", 80)  --> `opts` omitted
```

Will always fail. The internal implementation checks the number of arguments, but not whether it is `nil`. In case of `nil` it falls through and sets the `port` to the default 0 instead of the provided one.

So the code above will effectively try to connect to `www.google.com:0` instead of `www.google.com:80`.

current master still has the issue.

[This line](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L428) should be updated to something like this (untested and very rusty on my c code);
```c
n = lua_gettop(L);
while ( n>0 && lua_type(L, n) == LUA_TNIL) { --n; }
```

agentzh:
@Tieske Thanks for the report and test case! We'll look into this. I don't quite agree with your patch though :)

@doujiang24 Will you have a look at this? ;)

doujiang24:
@Tieske I opened a PR, will you take a look at it please?
https://github.com/openresty/lua-nginx-module/pull/862

Tieske:
on first glance the solution seems to work, but missing a testcase for the actual issue, eg. arguments; string, number, nil

as a more general comment; common Lua idiom is to ignore excess parameters, where openresty functions tend to throw an error. Maybe a consideration to drop those checks in the future?

> I don't quite agree with your patch though :)

Like I mentioned; rusty on my c, so I'm not even disappointed 😄

doujiang24:
@Tieske Sorry for the mistake, I have correct the test case, thanks :)

Tieske:
@doujiang24 looks good to me!

PR #862, opened 2016-09-04
[discuss] allow options_table is not a table for tcpsock:connect

doujiang24:
from https://github.com/openresty/lua-nginx-module/issues/860

Now we just ignore it when the `options_table` is not a table.
Or maybe we should return an error when the `options_table` is not a table nor nil ?


doujiang24:
@agentzh I have check the travis failed, seems like the DNS server response too faster?
Maybe we can change the resolver from `8.8.8.8` to openresty own DNS server？

Tieske:
lgtm!

agentzh:
Merged. Thanks!

PR #870, opened 2016-09-16
Added ngx.resp.set_status_reason function to set a custom reason phrase

jdesgats:

Allows to set the reason for HTTP status codes unknown from the nginx
core. Given the current implementation of nginx, it is safer and more
convenient to set the status at the same time.

Currently setting the status to an unknown response code causes the reason to be empty:

```
content_by_lua '
    ngx.status = 876
    ngx.say("custom response")
';
```

Causes the response to be:

```
< HTTP/1.1 876
< Date: Fri, 16 Sep 2016 20:06:33 GMT
< Content-Type: text/plain
< Transfer-Encoding: chunked
< Connection: keep-alive
<
custom response
```

The use cases can be to implement some of the error codes unknown from nignx (e.g. [RFC6585](https://tools.ietf.org/html/rfc6585)) or completely custom HTTP response codes.

There is still two open questions in my mind:
- Should we add an optional parameter in `ngx.exit` too?
- Do we add the `resty.core` wrapper now or in another PR?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -2985,6 +2985,7 @@ Nginx API for Lua
 * [ngx.status](#ngxstatus)
 * [ngx.header.HEADER](#ngxheaderheader)
 * [ngx.resp.get_headers](#ngxrespget_headers)
+* [ngx.resp.set_status_reason](#ngxrespset_status_reason)
 * [ngx.req.is_internal](#ngxreqis_internal)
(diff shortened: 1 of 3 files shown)


zhuizhuhaomeng:
We could add a third argument for ngx_http_lua_ffi_set_resp_status()

PR #878, opened 2016-09-30
feature: add support for building with BoringSSL

ghedo:

This should work with OpenSSL as well, since they cast the argument back
to (char *) in the macro.

Note that since BoringSSL doesn't implement OCSP, the ngx_lua OCSP support
is disabled when building with BoringSSL.


ghedo:
This is incomplete, so I'm closing for now.

PR #879, opened 2016-10-02
New directive lua_malloc_trim

agentzh:
feature: added new config directive "lua_malloc_trim N" to periodically call malloc_trim(1) every N requests when malloc_trim() is available.

By default, "lua_malloc_trim 1000" is configured.

This should fix the glibc oddity of holding too much freed memory when it fails to use brk() to allocate memory in the data segment (Linux only).

### DIFF
--- a/src/ngx_http_lua_logby.c
+++ b/src/ngx_http_lua_logby.c
@@ -72,14 +72,14 @@ ngx_http_lua_log_handler(ngx_http_request_t *r)
 {
 #if NGX_HTTP_LUA_HAVE_MALLOC_TRIM
     ngx_uint_t                   trim_cycle, trim_nreq;
-#endif
     ngx_http_lua_main_conf_t    *lmcf;
+#endif
     ngx_http_lua_loc_conf_t     *llcf;
     ngx_http_lua_ctx_t          *ctx;
 
+#if NGX_HTTP_LUA_HAVE_MALLOC_TRIM
     lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
 
-#if NGX_HTTP_LUA_HAVE_MALLOC_TRIM
     trim_cycle = lmcf->malloc_trim_cycle;
 
     if (trim_cycle > 0) {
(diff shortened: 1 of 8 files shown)


agentzh:
@yangshuxin @doujiang24 Please have a look. Many thanks!

src/ngx_http_lua_logby.c
```diff
@@ -67,9 +70,39 @@ ngx_http_lua_log_by_lua_env(lua_State *L, ngx_http_request_t *r)
 ngx_int_t
 ngx_http_lua_log_handler(ngx_http_request_t *r)
 {
+#if NGX_HTTP_LUA_HAVE_MALLOC_TRIM
+    ngx_uint_t                   trim_cycle, trim_nreq;
+#endif
+    ngx_http_lua_main_conf_t    *lmcf;
```
doujiang24: seems it's better to move line 76 to line 75? so does line 80?

doujiang24:
@agentzh LGTM :)

src/ngx_http_lua_logby.c
```diff
@@ -67,9 +70,39 @@ ngx_http_lua_log_by_lua_env(lua_State *L, ngx_http_request_t *r)
...
+        dd("cycle: %d", (int) trim_cycle);
+
+        trim_nreq = ++lmcf->malloc_trim_req_count;
+
+        if (trim_nreq >= trim_cycle) {
+            lmcf->malloc_trim_req_count = 0;
+
+            ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                           "malloc_trim(1) returned %d", malloc_trim(1));
+        }
```
yangshuxin: I could be wrong. if the '#if NGX_DEBUG' evaluates to false, none of the actual arguments to the ngx_log_debug1, meaning the malloc_trim() is not called at all. Other than this one, LGTM.

agentzh: Indeed you are right. Nice catch! Will fix it tomorrow. Thanks!

PR #885, opened 2016-10-07
doc: use *_by_lua_block {} in the examples.

philipjkim:

### DIFF
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -2620,7 +2620,7 @@ Undefined NGINX variables are evaluated to `nil` while uninitialized (but define
 This API requires a relatively expensive metamethod call and it is recommended to avoid using it on hot code paths.
 
 == Core constants ==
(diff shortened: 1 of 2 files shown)


doc/HttpLuaModule.wiki
```diff
@@ -2620,7 +2620,7 @@ Undefined NGINX variables are evaluated to `nil` while uninitialized (but define
 This API requires a relatively expensive metamethod call and it is recommended to avoid using it on hot code paths.
 
 == Core constants ==
-'''context:''' ''init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua, *log_by_lua*, ngx.timer.*, balancer_by_lua*, ssl_certificate_by_lua*, ssl_session_fetch_by_lua*, ssl_session_store_by_lua*''
+'''context:''' ''init_by_lua*, set_by_lua*, rewrite_by_lua*, access_by_lua*, content_by_lua*, header_filter_by_lua*, body_filter_by_lua, body_filter_by_lua_block, *log_by_lua*, ngx.timer.*, balancer_by_lua*, ssl_certificate_by_lua*, ssl_session_fetch_by_lua*, ssl_session_store_by_lua*''
```
agentzh: I think we should use `body_filter_by_lua*` instead of `body_filter_by_lua, body_filter_by_lua_block` here (and we also need to cover `body_filter_by_lua_file` anyway).

Please fix other similar places as well.

philipjkim: @agentzh I also thought that would be better. Please review the latest commit.

agentzh:
Merged. Thanks!

Issue #905, opened 2016-11-02
Lua VM crash when using ngx.exec(), unsafe uri


Jayce:
See crash point: [ngx_http_lua_control.c](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_control.c#L108-L112).

this code `return NGX_HTTP_INTERNAL_SERVER_ERROR;` causes the VM to crash.

detail:
* version openresty-1.9.7.3 and ngx_lua-0.10.0
* system log
```sh
Oct 28 15:11:18 f02 kernel: nginx[14476]: segfault at 140d92420 ip 00007fe7f20ee40d sp 00007fff66396f20 error 4 in libluajit-5.1.so.2[7fe7f20e5000+c1000]
Oct 28 15:11:39 f02 kernel: nginx[14489]: segfault at 141380a48 ip 00007fe7f20ee40d sp 00007fff66396ea0 error 4 in libluajit-5.1.so.2[7fe7f20e5000+c1000]
Oct 28 15:11:59 f02 kernel: nginx[14486]: segfault at 141eb8f10 ip 00007fe7f20ee40d sp 00007fff66396ea0 error 4 in libluajit-5.1.so.2[7fe7f20e5000+c1000]
```

* nginx log
```sh
2016/11/02 17:24:22 [error] 21115#0: *60 unsafe URI "/error_page/../" was detected *******
```

* configure
```sh
        location / {
            content_by_lua_block {
                ngx.log(ngx.ERR, "/error_page/ ")
                return ngx.exec("/error_page/../")
            }
        }
```

doujiang24:
@Jayce Thanks for your report. Already replied in https://groups.google.com/d/msg/openresty/R4TuNOYqO2o/wgvRoaKZCgAJ

Will you please create a PR, please don't forgot to add test case correspond to it :)
Thanks again :)

PR #907, opened 2016-11-02
bugfix: Lua VM crash when using ngx.exec(), unsafe uri #905

Jayce:
This PR is about to the issue #905


agentzh:
@Jayce Thanks for the patch! Applied with minor edits.

PR #912, opened 2016-11-14
[bugfix] Avoid the package.preload module not found while code cache off

rainingmaster:
Sorry, this pull request is the follow-up for [this](https://github.com/openresty/lua-nginx-module/pull/840), I try to add a test case in it, I hope it work :)


agentzh:
Merged with some minor adjustments. Thanks!

PR #917, opened 2016-11-18
optimze: reduce the ngx.req.socket object size & reduce ngx.socket.udp object size.

doujiang24:
Seems we only need to save the `timeout` values to the `ngx_http_lua_socket_tcp_upstream_t u` for the connected `request socket`.
And the `udp socket` don't have to save the key for keepalive.

### DIFF
--- a/t/087-udp-socket.t
+++ b/t/087-udp-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (3 * blocks() + 13);
+plan tests => repeat_each() * (3 * blocks() + 14);
 
 our $HtmlDir = html_dir;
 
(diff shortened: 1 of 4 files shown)


t/087-udp-socket.t
```diff
@@ -1103,3 +1103,54 @@ qr/runtime error: content_by_lua\(nginx\.conf:\d+\):14: bad request/
...
+
+=== TEST 20: udp object size
+--- config
+    server_tokens off;
+    location /t {
+        #set $port 5000;
+        set $port $TEST_NGINX_MEMCACHED_PORT;
+        #set $port 1234;
+
+        content_by_lua '
```
agentzh: New tests better use `_by_lua_block {...}` :)

doujiang24: Sorry for missing this, fixed :)

PR #919, opened 2016-11-21
codestyle: for get the index of the top element in the stack from lua

detailyang:

Hello @agentzh .

When i'm reading the codebase I have noticed that when we want to get the index of the top element in the stack from lua `lua_gettop(L);` for check lua number of arguments (It's common operate), some code like [ngx_http_lua_shdict.c#L463](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_shdict.c#L463) but other like [ngx_http_lua_socket_tcp.c#L2744](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L2744).

By the way, there are some similar questions in other files. Now the PR just change the [lua_socket_tcp.c](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c)

How do you think of this :)

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -376,12 +376,14 @@ ngx_http_lua_inject_req_socket_api(lua_State *L)
 static int
 ngx_http_lua_socket_tcp(lua_State *L)
 {
+    int                      n;
     ngx_http_request_t      *r;
     ngx_http_lua_ctx_t      *ctx;
 
-    if (lua_gettop(L) != 0) {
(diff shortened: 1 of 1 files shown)


detailyang:
It looks like the tests do not past:(. I will look at this

agentzh:
@detailyang I don't like these changes. Since it introduced an extra `n` local variable and the only other uses of this `n` variable is on a very cold error handling code path, which does not worth it.

detailyang:
@agentzh thx for you suggestion to point out 'this n variable is on a very cold error handling code path' 👍 . i will close the PR

PR #935, opened 2016-12-21
support multiple rewrite directives

membphis:
Its only support `rewrite` directive now. I'll add more directives in later.


src/ngx_http_lua_common.h
```diff
@@ -236,6 +236,17 @@ union ngx_http_lua_srv_conf_u {
...
 
 typedef struct {
+    ngx_int_t                is_handler_inline;
+    u_char                  *rewrite_chunkname;
+    ngx_http_complex_value_t rewrite_src;    /*  rewrite_by_lua
+                                                inline script/script
+                                                file path */
+
+    u_char                  *rewrite_src_key; /* cached key for rewrite_src */
+} ngx_http_lua_rewrites_t;
```
agentzh: Seems like we can reuse this struct for other phases as well? So we do not need to encode "rewrite" into this struct's names.

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,20 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         return NGX_CONF_ERROR;
     }
 
-    if (llcf->rewrite_handler) {
-        return "is duplicate";
+    if (p_rewrites == NULL) {
+        p_rewrites = ngx_array_create(cf->pool, 8,
```
agentzh: Multiple same-phase handlers should be rare. We can set the initial size to 1 instead of 8.

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,20 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+                                      sizeof(ngx_http_lua_rewrites_t));
+
+        if (p_rewrites == NULL) {
+            return "not enough memory";
```
agentzh: The common practice is to simply return `NGX_CONF_ERROR` in this case. See other similar places in this C function.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +164,62 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+
+    rc = NGX_DECLINED;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    p_sets = llcf->rewrites;
+
+    for (i = 0; i < p_sets->nelts; ++i)
+    {
+        rewrite = (ngx_http_lua_rewrites_t *)((char *)p_sets->elts
```
agentzh: We usually do not index an element of `ngx_array_t` object in the nginx world. Please use the common idiom like below:

```
elem_type_t   *foo;
foo = array->elts;
for (i = 0; i < array->nelts; i++) {
    /* then index the element like this: */
    foo[i]
}
```

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +164,62 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
-ngx_http_lua_rewrite_handler_inline(ngx_http_request_t *r)
+ngx_http_lua_rewrite_handler_sets(ngx_http_request_t *r)
 {
-    lua_State                   *L;
     ngx_int_t                    rc;
     ngx_http_lua_loc_conf_t     *llcf;
+    ngx_array_t                 *p_sets;    /*   rewrite sets, inline
+                                                script/script file path */
+    ngx_uint_t                   i;
+    ngx_http_lua_rewrites_t     *rewrite;
```
agentzh: This `rewrite` variable name is bad. It's not very descriptive.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +164,62 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
-    dd("rewrite by lua inline");
+    dd("rewrite by lua handler sets");
+
+    rc = NGX_DECLINED;
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    p_sets = llcf->rewrites;
+
+    for (i = 0; i < p_sets->nelts; ++i)
```
agentzh: This does not look quite right. Since a rewrite phase handler can yield and resume. And your loop always starts running from the first lua rewrite phase handler, which is wrong (when the 2nd handler yields and resumes, for example).

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,101 @@
...
+        }
+        rewrite_by_lua_block { ngx.print("Hello, Lua!\n") }
+        content_by_lua return;
+    }
+--- request
+GET /lua
+--- response_body
+Hello, Lua!
+--- error_log
+rewrite_by_lua_block first
```
agentzh: You should add tests for the cases that some of the rewrite phase handlers *yield* (like `ngx.sleep(0.001)` and `ngx.location.capture`) or something. Your current implementation will fail :)

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,101 @@
...
+GET /lua
+--- response_body
+Hello, Lua!
+
+
+
+=== TEST 3: mix three styles
+--- config
+    location /lua {
+        rewrite_by_lua ' ngx.log(ngx.ERR, "rewrite_by_lua") ';
```
agentzh: Better avoid using the literal string form in newly added tests. Always use `rewrite_by_lua_block` for inlined Lua scripts :)

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,101 @@
...
+
+=== TEST 5: the first return status == 0
+--- config
+    location /lua {
+        rewrite_by_lua_block {
+          ngx.log(ngx.ERR, "rewrite_by_lua_block first")
+          ngx.exit(0)
+        }
+        rewrite_by_lua_block { ngx.print("Hello, Lua!\n") }
+        content_by_lua return;
```
agentzh: Ditto.

t/023-rewrite/duplicate.t
agentzh: Also, I think we need tests mixing `rewrite_by_lua_block` and `rewrite_by_lua_file` directives in the same scope.

agentzh: And we also need test cases for directives in different scopes so that we can test directive inheritance and overriding.

moonbingbing:
Do we need to limit the max rewrite directives?

agentzh:
@moonbingbing Yes, we do! How about a hard-coded limit like 5?

agentzh:
@moonbingbing Or 10 :)

membphis:
limit the max rewrite directives to 10 now.

src/ngx_http_lua_common.h
membphis: change `ngx_http_lua_rewrites_t` to `ngx_http_lua_phase_ctx_t`, `rewrite_src_key ` to `source_key `.

src/ngx_http_lua_directive.c
membphis: did

src/ngx_http_lua_directive.c
membphis: done

src/ngx_http_lua_rewriteby.c
membphis: change it to `phase_ctx`

src/ngx_http_lua_rewriteby.c
membphis: It's my fault, i have fixed it.

src/ngx_http_lua_rewriteby.c
membphis: :) copy this style to new code

t/023-rewrite/duplicate.t
membphis: done

t/023-rewrite/duplicate.t
membphis: I have pushed new code to support the yield (like `ngx.sleep(0.001)` and `ngx.location.capture`).

the `TEST 3: mix three different styles` have mixed `rewrite_by_lua`, `rewrite_by_lua_block`, `rewrite_by_lua_file`.

and add new test case for directives in different scopes.

src/ngx_http_lua_module.c
```diff
@@ -241,7 +241,7 @@ static ngx_command_t ngx_http_lua_cmds[] = {
       ngx_http_lua_rewrite_by_lua,
       NGX_HTTP_LOC_CONF_OFFSET,
       0,
-      (void *) ngx_http_lua_rewrite_handler_inline },
+      (void *) ngx_http_lua_rewrite_handler_sets },
```
agentzh: The `_sets` suffix looks strange. What does it mean? Settings? But "sets" are very different from "settings", I'm afraid.

src/ngx_http_lua_rewriteby.c
```diff
@@ -126,6 +126,13 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
 
                 return NGX_HTTP_OK;
             }
+            else {
```
agentzh: The coding style around `} else {` is wrong. Please see similar places in the existing code base. Please be very careful :)

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +171,73 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+    }
+
+    dd("entered rewrite by lua handler sets[%ld]", ctx->rewrite_index);
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    phase_ctx = llcf->rewrites->elts;
+
+    for (i = ctx->rewrite_index; i < llcf->rewrites->nelts; ++i)
+    {
```
agentzh: Coding style: `{` should be in the previous line if the previous line is not splitted into multiple lines itself.

src/ngx_http_lua_common.h
```diff
@@ -236,6 +240,16 @@ union ngx_http_lua_srv_conf_u {
 
 
 typedef struct {
+    ngx_int_t                is_handler_inline;
+    u_char                  *chunkname;
+    ngx_http_complex_value_t source;    /*  inline script/script
```
agentzh: Seems like we need more spaces between the type names and the field names. It's a bit too crowded I'm afraid :)

src/ngx_http_lua_common.h
```diff
@@ -236,6 +240,16 @@ union ngx_http_lua_srv_conf_u {
...
+                                            file path */
+
+    u_char                  *source_key; /* cached key for source */
+} ngx_http_lua_phase_ctx_t;
```
agentzh: Maybe `ngx_http_lua_phase_handler_t` is better?

src/ngx_http_lua_common.h
```diff
@@ -261,12 +275,8 @@ typedef struct {
 
     ngx_http_output_body_filter_pt         body_filter_handler;
 
-    u_char                  *rewrite_chunkname;
-    ngx_http_complex_value_t rewrite_src;    /*  rewrite_by_lua
-                                                inline script/script
-                                                file path */
-
-    u_char                  *rewrite_src_key; /* cached key for rewrite_src */
+    ngx_array_t             *rewrites;  /*   rewrite sets, inline script/script
```
agentzh: Maybe `rewrite_handlers` is better than `rewrites` here?

src/ngx_http_lua_common.h
```diff
@@ -526,6 +536,8 @@ typedef struct ngx_http_lua_ctx_s {
     unsigned         acquired_raw_req_socket:1;  /* whether a raw req socket
                                                     is acquired */
     unsigned         seen_body_data:1;
+
+    ngx_uint_t       rewrite_index;
```
agentzh: Maybe `current_rewrite_handler` is better than `rewrite_index` here?

src/ngx_http_lua_directive.c
```diff
@@ -450,6 +450,8 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
     ngx_str_t                   *value;
     ngx_http_lua_main_conf_t    *lmcf;
     ngx_http_lua_loc_conf_t     *llcf = conf;
+    ngx_array_t                 *p_rewrites = llcf->rewrites;
```
agentzh: We usually do not use the `p_blahblah` naming convention. So better avoid that here to be more consistent.

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,27 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+        if (p_rewrites == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        llcf->rewrites = p_rewrites;
+    }
+
+    if (p_rewrites->nelts >= NGX_HTTP_LUA_MAX_PHASE_COUNT) {
+        ngx_conf_log_error(NGX_LOG_ERR, cf, 0,
+                           "max rewrite directives limit to %d",
```
agentzh: This English sentence is not quite right. Maybe `"the number of rewrite_by_lua* directives exceeds %d"`?

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,27 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+                           NGX_HTTP_LUA_MAX_PHASE_COUNT);
+        return NGX_CONF_ERROR;
+    }
+
+    phase_ctx = ngx_array_push(p_rewrites);
+    if (phase_ctx == NULL) {
+        return "not enough memory";
```
agentzh: Again, we do not return such C strings upon allocation failures in the config handlers.

src/ngx_http_lua_module.c
```diff
@@ -1037,8 +1037,8 @@ ngx_http_lua_create_loc_conf(ngx_conf_t *cf)
     /* set by ngx_pcalloc:
      *      conf->access_src  = {{ 0, NULL }, NULL, NULL, NULL};
      *      conf->access_src_key = NULL
-     *      conf->rewrite_src = {{ 0, NULL }, NULL, NULL, NULL};
-     *      conf->rewrite_src_key = NULL
+     *      conf->source = {{ 0, NULL }, NULL, NULL, NULL};
+     *      conf->source_key = NULL
      *      conf->rewrite_handler = NULL;
```
agentzh: I'm afraid these comments are wrong because `conf` no longer take the `source` and `source_key` fields directly (but via your newly defined `struct` as an intermediate field).

src/ngx_http_lua_module.c
membphis: the `sets` means `group`, `suit` and `series` in here.

src/ngx_http_lua_common.h
membphis: There are `rewrite_handler` and `rewrite_handlers` in struct `ngx_http_lua_loc_conf_t`.  Will it cause any confusion?

src/ngx_http_lua_common.h
membphis: nice advice, change it later.

src/ngx_http_lua_module.c
membphis: fixed it right now.

src/ngx_http_lua_rewriteby.c
membphis: Got it.

src/ngx_http_lua_rewriteby.c
membphis: i'll pay more attention on code style.

src/ngx_http_lua_common.h
membphis: change the variable name to `current_rewrite_handler `

src/ngx_http_lua_directive.c
membphis: the current name is `rewrite_handlers`

src/ngx_http_lua_directive.c
membphis: changed to `"the number of rewrite_by_lua* directives exceeds %d"`

src/ngx_http_lua_directive.c
membphis: Do we need to change all the similar points?

eg: https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_directive.c#L578

src/ngx_http_lua_rewriteby.c
```diff
@@ -125,6 +125,12 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
                 }
 
                 return NGX_HTTP_OK;
+            } else {
```
doujiang24: style: blank line needed before this like :)

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +170,74 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+        }
+    }
+
+    return rc;
+}
+
+
+ngx_int_t
+ngx_http_lua_rewrite_handler_inline(ngx_http_request_t *r,
+    ngx_http_lua_phase_handler_t *phase_handler)
```
doujiang24: I personally think `handler`  is better than `phase_handler` here?

src/ngx_http_lua_rewriteby.c
membphis: nice point. fix it later.

src/ngx_http_lua_rewriteby.c
membphis: `handler` is more simple, love it.

src/ngx_http_lua_directive.c
```diff
@@ -446,10 +446,12 @@ ngx_http_lua_rewrite_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
...
-    u_char                      *p, *chunkname;
-    ngx_str_t                   *value;
-    ngx_http_lua_main_conf_t    *lmcf;
-    ngx_http_lua_loc_conf_t     *llcf = conf;
+    u_char                          *p, *chunkname;
+    ngx_str_t                       *value;
+    ngx_http_lua_main_conf_t        *lmcf;
+    ngx_http_lua_loc_conf_t         *llcf = conf;
+    ngx_array_t                     *rewrite_handlers = llcf->rewrite_handlers;
+    ngx_http_lua_phase_handler_t    *phase_handler;
```
agentzh: Local C variables sholud use short names (Huffman principle). How about use `ph` for `phase_handler` here?

src/ngx_http_lua_directive.c
```diff
@@ -446,10 +446,12 @@ ngx_http_lua_rewrite_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
...
 {
-    u_char                      *p, *chunkname;
-    ngx_str_t                   *value;
-    ngx_http_lua_main_conf_t    *lmcf;
-    ngx_http_lua_loc_conf_t     *llcf = conf;
+    u_char                          *p, *chunkname;
+    ngx_str_t                       *value;
+    ngx_http_lua_main_conf_t        *lmcf;
+    ngx_http_lua_loc_conf_t         *llcf = conf;
+    ngx_array_t                     *rewrite_handlers = llcf->rewrite_handlers;
```
agentzh: Like below, I think `handlers` is a good enough name for a local C variable.

src/ngx_http_lua_rewriteby.c
```diff
@@ -125,6 +125,13 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
                 }
 
                 return NGX_HTTP_OK;
+
+            } else {
```
agentzh: Seems like the previous branch always returns and never falls through, and we can save the `else` here.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +171,74 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+
+    dd("entered rewrite by lua handler sets[%ld]", ctx->current_rewrite_index);
+
+    llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    rewrite_handlers = llcf->rewrite_handlers;
+    handler = rewrite_handlers->elts;
+
+    for (i = ctx->current_rewrite_index; i < rewrite_handlers->nelts; ++i) {
+        ctx->current_rewrite_index = i + 1;
```
agentzh: This looks wrong to me since a single handler may yield multiple times. Do you have test cases for a single Lua handler yielding multiple times? Like a series of `ngx.sleep` calls in a single Lua handler?

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,303 @@
...
+        echo "internal";
+    }
+    location /t {
+        rewrite_by_lua_block {
+            ngx.log(ngx.ERR, "first rewrite before sleep")
+
+            local res = ngx.location.capture("/internal")
+            ngx.log(ngx.ERR, "status:", res.status, " body:", res.body)
+
+            ngx.log(ngx.ERR, "first rewrite after sleep")
```
agentzh: This is a copy&paste mistake. We are not doing sleeps here.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +171,74 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+    dd("ctx = %p", ctx);
+
+    if (ctx == NULL) {
+        ctx = ngx_http_lua_create_ctx(r);
+        if (ctx == NULL) {
+            return NGX_HTTP_INTERNAL_SERVER_ERROR;
+        }
+    }
+
+    dd("entered rewrite by lua handler sets[%ld]", ctx->current_rewrite_index);
```
agentzh: We need a comment indicating where we initialize this `current_rewrite_index` field.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +171,74 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+
+    llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    rewrite_handlers = llcf->rewrite_handlers;
+    handler = rewrite_handlers->elts;
+
+    for (i = ctx->current_rewrite_index; i < rewrite_handlers->nelts; ++i) {
```
agentzh: Style: do not use the form `++i` unless necessary. We usually use the `i++` form only in the nginx world.

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,303 @@
...
+            ngx.sleep(0.001)
+        }
+        rewrite_by_lua_block { ngx.log(ngx.ERR, "second rewrite") }
+        rewrite_by_lua_block { ngx.print("Hello, Lua!\n") }
+    }
+--- request
+GET /t
+--- response_body
+Hello, Lua!
+--- error_log
```
agentzh: I think it's better to use `grep_error_log` and `grep_error_log_out` here so that we can ensure there are no duplicate log entries for the same handler. It is a stronger check than `error_log`.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +171,74 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
 
 
 ngx_int_t
-ngx_http_lua_rewrite_handler_inline(ngx_http_request_t *r)
+ngx_http_lua_rewrite_handler_sets(ngx_http_request_t *r)
```
agentzh: I don't like the `_sets` suffix, which is too vague. Let's come up with a better name.

agentzh: How about `ngx_http_lua_run_rewrite_handlers`?

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,28 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    if (rewrite_handlers->nelts >= NGX_HTTP_LUA_MAX_PHASE_COUNT) {
+        ngx_conf_log_error(NGX_LOG_ERR, cf, 0,
+                           "the number of rewrite_by_lua* directives exceeds %d",
+                           NGX_HTTP_LUA_MAX_PHASE_COUNT);
+        return NGX_CONF_ERROR;
+    }
+
+    phase_handler = ngx_array_push(rewrite_handlers);
+    if (phase_handler == NULL) {
+        ngx_conf_log_error(NGX_LOG_ERR, cf, 0, "no memory");
```
agentzh: Again: we do not log no memory errors on allocation failures on the nginx C land.

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,303 @@
...
+--- response_body
+Hello /t
+--- error_log
+rewrite 1 at location
+rewrite 2 at location
+--- no_error_log
+rewrite 1 at http
+rewrite 2 at http
+rewrite 1 at server
+rewrite 2 at server
```
agentzh: Need a test case reading request bodies via `ngx.req.read_body()` in the rewrite phase handler. It might be a slightly different kind of "yielding" :)

src/ngx_http_lua_common.h
```diff
@@ -96,6 +96,10 @@ typedef struct {
 #define NGX_HTTP_LUA_MAX_ARGS 100
 #endif
 
+#ifndef NGX_HTTP_LUA_MAX_PHASE_COUNT
+#define NGX_HTTP_LUA_MAX_PHASE_COUNT 10
```
agentzh: I'm afraid this macro name can be confusing. We are not limiting the number of phases but rather the number of the handlers in *each* phase, right? Let's find a better name :)

src/ngx_http_lua_directive.c
membphis: fixed

src/ngx_http_lua_directive.c
membphis: fixed

src/ngx_http_lua_common.h
membphis: change it to `NGX_HTTP_LUA_MAX_HANDLER_EACH_PHASE`

src/ngx_http_lua_directive.c
membphis: sorry for this, it's my fault. fixed.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

src/ngx_http_lua_rewriteby.c
membphis: fixed

src/ngx_http_lua_rewriteby.c
membphis: The `duplicate.t/TEST 7` has included a series of `ngx.sleep` calls in a single Lua handler. it works fine.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

t/023-rewrite/duplicate.t
membphis: Great test feature :) all of the test case have beed fixed.

t/023-rewrite/duplicate.t
membphis: fixed

t/023-rewrite/duplicate.t
membphis: add new test case reading bodies via ngx.req.read_body() in the rewrite phase handler.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +170,77 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+        }
+    }
+
+    return rc;
+}
+
+
+ngx_int_t
+ngx_http_lua_rewrite_handler_inline(ngx_http_request_t *r,
+    ngx_http_lua_phase_handler_t *h)
```
agentzh: A single letter `h` might be a bit too short. How about `ph`?

src/ngx_http_lua_rewriteby.c
```diff
@@ -191,17 +250,15 @@ ngx_http_lua_rewrite_handler_inline(ngx_http_request_t *r)
 
 
 ngx_int_t
-ngx_http_lua_rewrite_handler_file(ngx_http_request_t *r)
+ngx_http_lua_rewrite_handler_file(ngx_http_request_t *r,
+    ngx_http_lua_phase_handler_t *h)
```
agentzh: Ditto.

src/ngx_http_lua_common.h
```diff
@@ -96,6 +96,10 @@ typedef struct {
 #define NGX_HTTP_LUA_MAX_ARGS 100
 #endif
 
+#ifndef NGX_HTTP_LUA_MAX_HANDLER_EACH_PHASE
+#define NGX_HTTP_LUA_MAX_HANDLER_EACH_PHASE 10
```
agentzh: Maybe `NGX_HTTP_LUA_MAX_HANDLERS_PER_PHASE` is a better name?

t/023-rewrite/duplicate.t
```diff
@@ -0,0 +1,350 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
```
agentzh: I think the file name `dup-cmds.t` is clearer than `duplicate.t`. What's your opinion? ;)

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,27 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+        if (handlers == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        llcf->rewrite_handlers = handlers;
+    }
+
+    if (handlers->nelts >= NGX_HTTP_LUA_MAX_HANDLER_EACH_PHASE) {
+        ngx_conf_log_error(NGX_LOG_ERR, cf, 0,
+                           "the number of rewrite_by_lua* directives exceeds %d",
```
agentzh: This line exceeds 80 columns. Please use the `ngx-releng` tool to check such things. Thank you :)

src/ngx_http_lua_directive.c
```diff
@@ -464,8 +466,27 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+
+        if (handlers == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        llcf->rewrite_handlers = handlers;
+    }
+
+    if (handlers->nelts >= NGX_HTTP_LUA_MAX_HANDLER_EACH_PHASE) {
+        ngx_conf_log_error(NGX_LOG_ERR, cf, 0,
```
agentzh: I think this error is important enough to use the `EMERG` log level.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

src/ngx_http_lua_common.h
membphis: yes, fixed.

t/023-rewrite/duplicate.t
membphis: yes, fixed.

src/ngx_http_lua_directive.c
membphis: ok, i'll pay more attention on these point.

src/ngx_http_lua_accessby.c
```diff
@@ -123,9 +123,16 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
                 }
 
                 return NGX_HTTP_OK;
+
+            }
+
+            rc = ngx_http_lua_run_access_handlers(r);
+
+            if (rc == NGX_DECLINED) {
+                rc = NGX_OK;
```
agentzh: If you are returning this constant anyway, why not just `return NGX_OK` here?

src/ngx_http_lua_common.h
```diff
@@ -538,6 +534,7 @@ typedef struct ngx_http_lua_ctx_s {
     unsigned         seen_body_data:1;
 
     ngx_uint_t       cur_rewrite_index;
+    ngx_uint_t       cur_access_index;
```
agentzh: I'm afraid it's wrong to add word-sized fields after bit-sized fields. It creates potential padding holes in the middle, which wastes some space.

src/ngx_http_lua_accessby.c
```diff
@@ -165,22 +172,74 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
...
+
+    if (ctx == NULL) {
+        ctx = ngx_http_lua_create_ctx(r);
+        if (ctx == NULL) {
+            return NGX_HTTP_INTERNAL_SERVER_ERROR;
+        }
+    }
+
+    /* ctx->cur_rewrite_index was initialized to 0 in
+     * ngx_http_lua_init_ctx */
```
agentzh: This comment should be in `ngx_http_lua_init_ctx` instead of here.

src/ngx_http_lua_accessby.c
```diff
@@ -165,22 +171,74 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
...
+
+    if (ctx == NULL) {
+        ctx = ngx_http_lua_create_ctx(r);
+        if (ctx == NULL) {
+            return NGX_HTTP_INTERNAL_SERVER_ERROR;
+        }
+    }
+
+    /* ctx->cur_rewrite_index was initialized to 0 in
+     * ngx_http_lua_init_ctx called by ngx_http_lua_create_ct */
```
agentzh: This comment should be in `ngx_http_init_ctx` instead of here.

src/ngx_http_lua_accessby.c
```diff
@@ -125,7 +125,13 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
                 return NGX_HTTP_OK;
             }
 
-            return NGX_OK;
+            rc = ngx_http_lua_run_access_handlers(r);
+
+            if (rc == NGX_DECLINED) {
+                rc = NGX_OK;
```
agentzh: Why not directly use `return NGX_OK` here?

src/ngx_http_lua_rewriteby.c
```diff
@@ -126,6 +126,12 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
 
                 return NGX_HTTP_OK;
             }
+
+            rc = ngx_http_lua_run_rewrite_handlers(r);
+
+            if (rc == NGX_OK) {
+                rc = NGX_DECLINED;
```
agentzh: Ditto.

src/ngx_http_lua_rewriteby.c
```diff
@@ -164,24 +170,77 @@ ngx_http_lua_rewrite_handler(ngx_http_request_t *r)
...
+
+    if (ctx == NULL) {
+        ctx = ngx_http_lua_create_ctx(r);
+        if (ctx == NULL) {
+            return NGX_HTTP_INTERNAL_SERVER_ERROR;
+        }
+    }
+
+    /* ctx->cur_rewrite_index was initialized to 0 in
+     * ngx_http_lua_init_ctx called by ngx_http_lua_create_ctx */
```
agentzh: Ditto.

src/ngx_http_lua_common.h
```diff
@@ -526,6 +532,9 @@ typedef struct ngx_http_lua_ctx_s {
     unsigned         acquired_raw_req_socket:1;  /* whether a raw req socket
                                                     is acquired */
     unsigned         seen_body_data:1;
+
+    ngx_uint_t       cur_rewrite_index;
+    ngx_uint_t       cur_access_index;
```
agentzh: It's wrong to specify bigger fields after smaller fields, which may create padding holes in between.

agentzh:
I suggest adding tests for different settings of the [satisfy](http://nginx.org/en/docs/http/ngx_http_core_module.html#satisfy) directives for multiple `access_by_lua*` directives in the same location.

agentzh:
All those `access_by_lua*` directives should follow the semantics of the standard [satisfy](http://nginx.org/en/docs/http/ngx_http_core_module.html#satisfy) directive.

src/ngx_http_lua_common.h
membphis: change it to `unsigned cur_rewrite_index:4;`, the max number is big enough in this case.

src/ngx_http_lua_accessby.c
membphis: fixed

src/ngx_http_lua_accessby.c
membphis: revert the code of `access` directives in this PR.

src/ngx_http_lua_rewriteby.c
membphis: fixed.

src/ngx_http_lua_rewriteby.c
membphis: fixed

membphis:
Revert the code of `access` directives in current PR. 

I'll submit a new PR for `access` directives including the test case of  `satisfy` directive.

membphis:
@agentzh @doujiang24

membphis:
do not need this feature now.

PR #942, opened 2016-12-29
ngx.shared.DICT.incr optional expire argument

splitice:
Rebased #358

I'll do a future commit changing all the shdict tests to _block.

### DIFF
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -1517,7 +1517,7 @@ setvalue:
         sd->expires = (uint64_t)tp->sec * 1000 + tp->msec
             + (uint64_t)(exptime * 1000);
 
-    } else if (exptime == 0) {
+    } else {
         dd("setting key to never expire");
         sd->expires = 0;
     }
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -1293,7 +1293,7 @@ ngx_http_lua_shdict_incr(lua_State *L)
             }
             has_init = 1;
         }
-        
+
         if (n >= 5) {
             if (!lua_isnil(L, 5)) {
                 exptime = luaL_checknumber(L, 5);
@@ -2763,8 +2763,8 @@ ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
         dd("setting expire time to %d", exptime);
 
         tp = ngx_timeofday();
-        sd->expires = (uint64_t)tp->sec * 1000 + tp->msec
-            + (uint64_t)(exptime * 1000);
+        sd->expires = (uint64_t) tp->sec * 1000 + tp->msec
+            + (uint64_t) (exptime * 1000);
 
     } else if (exptime == 0) {
         dd("setting key to never expire");
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -1273,18 +1273,18 @@ ngx_http_lua_shdict_incr(lua_State *L)
 
     if (n < 3) {
         return luaL_error(L, "expecting at least 3 arguments, but only seen %d",
-            n);
+                          n);
     }
-    
+
     if (n > 5) {
         return luaL_error(L, "expecting no more than 5 arguments, but %d seen",
-            n);
+                          n);
     }
 
     if (lua_type(L, 1) != LUA_TTABLE) {
         return luaL_error(L, "bad \"zone\" argument");
     }
-    
+
     if (n >= 4) {
         if (!lua_isnil(L, 4)) {
             init = luaL_checknumber(L, 4);
@@ -1404,8 +1404,8 @@ ngx_http_lua_shdict_incr(lua_State *L)
         dd("setting expire time to %d", exptime);
 
         tp = ngx_timeofday();
-        sd->expires = (uint64_t)tp->sec * 1000 + tp->msec
-            + (uint64_t)(exptime * 1000);
+        sd->expires = (uint64_t) tp->sec * 1000 + tp->msec
+            + (uint64_t) (exptime * 1000);
 
     } else if (exptime == 0) {
         dd("setting key to never expire");
@@ -1516,8 +1516,8 @@ setvalue:
         dd("setting expire time to %d", exptime);
 
         tp = ngx_timeofday();
-        sd->expires = (uint64_t)tp->sec * 1000 + tp->msec
-            + (uint64_t)(exptime * 1000);
+        sd->expires = (uint64_t) tp->sec * 1000 + tp->msec
+            + (uint64_t) (exptime * 1000);
 
     } else {
         dd("setting key to never expire");
@@ -2668,7 +2668,7 @@ ngx_http_lua_ffi_shdict_get(ngx_shm_zone_t *zone, u_char *key,
 
 int
 ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
-    size_t key_len, double *value, int exptime, char **err, int has_init, 
+    size_t key_len, double *value, int exptime, char **err, int has_init,
     double init, int *forcible)
 {
     int                          i, n;
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -1291,6 +1291,7 @@ ngx_http_lua_shdict_incr(lua_State *L)
             if (init < 0) {
                 return luaL_error(L, "bad \"init\" argument");
             }
+            
             has_init = 1;
         }
 
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -1257,23 +1257,49 @@ ngx_http_lua_shdict_incr(lua_State *L)
     ngx_http_lua_shdict_node_t  *sd;
     double                       num;
     double                       init = 0;
+    int                          has_init = 0;
     u_char                      *p;
     ngx_shm_zone_t              *zone;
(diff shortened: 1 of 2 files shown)


splitice:
Let me know if its good and I'll do stream & resty-core.

src/ngx_http_lua_shdict.c
```diff
@@ -1257,24 +1257,50 @@ ngx_http_lua_shdict_incr(lua_State *L)
...
     int                          forcible = 0;
     ngx_queue_t                 *queue, *q;
+    ngx_time_t                  *tp;
 
     n = lua_gettop(L);
 
-    if (n != 3 && n != 4) {
-        return luaL_error(L, "expecting 3 or 4 arguments, but only seen %d", n);
+    if (n < 3) {
+        return luaL_error(L, "expecting at least 3 arguments, but only seen %d", n);
```
membphis: This line exceeds 80 columns. Please use the [ngx-releng](https://github.com/openresty/openresty-devel-utils/blob/master/ngx-releng)  tool to check such things.

splitice:
done

t/043-shdict.t
```diff
@@ -2471,3 +2471,90 @@ error
...
 [emerg]
+
+
+
+=== TEST 94: incr expire test new key
+--- http_config
+    lua_shared_dict dogs 1m;
+--- config
+    location = /test {
+        content_by_lua '
```
membphis: Newly added test cases should use the *_by_lua_block form instead.

splitice: I'll do a future PR changing all the shdict tests to _block, not just these ones. It's an easy regex.

agentzh: @splitice I prefer keeping the old tests intact. Just the newly added ones should use `*_by_lua_block`.

src/ngx_http_lua_shdict.c
```diff
@@ -2712,6 +2759,18 @@ ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
...
     ngx_memcpy(p, (double *) &num, sizeof(double));
 
+    if (exptime > 0) {
+        dd("setting expire time to %d", exptime);
+
+        tp = ngx_timeofday();
+        sd->expires = (uint64_t) tp->sec * 1000 + tp->msec
+            + (uint64_t) (exptime * 1000);
+
+    } else if (exptime == 0) {
```
agentzh: Better make `exptime < 0` turn into `exptime == 0` as well. So better replace this `else if` with `else` directly?

agentzh: This should be the same as the CFunction implementation above.

src/ngx_http_lua_shdict.c
```diff
@@ -1257,25 +1257,53 @@ ngx_http_lua_shdict_incr(lua_State *L)
...
         return luaL_error(L, "bad \"zone\" argument");
     }
 
+    if (n >= 4) {
+        if (!lua_isnil(L, 4)) {
+            init = luaL_checknumber(L, 4);
+            if (init < 0) {
+                return luaL_error(L, "bad \"init\" argument");
+            }
+            has_init = 1;
```
agentzh: Style: need a blank line after the `}` line.

agentzh:
@splitice Have you created the corresponding PR for lua-resty-core's FFI-based API?

agentzh:
@splitice Thank you for your contribution! Will you please look at my comments and questions above? Thanks!

agentzh:
I think we also need updates in the documentation :)

agentzh:
Hmm, seems like we should permanently switch over to the FFI-based implementations to lua-resty-core. I'm already sick of maintaining two parallel implementations of the same Lua API :) I bet you can feel the pain as well :)

thibaultcha:
> mm, seems like we should permanently switch over to the FFI-based implementations to lua-resty-core.

I think so too. Having to provide two implementations of each new feature can also discourage a lot of contributors (I think it has been the case a few times already).

However, this does raise a few concerns on the long term: 
- if more and more new features of the API are only provided in lua-resty-core, is it still user-friendly to require from our users to manually require `resty.core` or the specific modules they need in every project?
- the same way, library authors might use some future functions that are only implemented in the lua-resty-core module, and thus, using such libraries would load lua-resty-core without the user being aware of it. This is maybe less of a concern that the first point.

agentzh:
@thibaultcha To answer your concerns:

1. We automatically require `resty.core` before running `init_by_lua*` unless `lua_use_resty_core off` is configured in the user `nginx.conf` (which is on by default).
2. It's not a problem IMHO, since ngx_lua will require lua-resty-core and lua-resty-lrucache unless `lua_use_resty_core` is explicitly turned off :)

agentzh:
Another upside of this is that the ngx_lua core will also be smaller :) The downside is that we now require lua-resty-core and lua-resty-lrucache for ngx_lua.

thibaultcha:
> We automatically require resty.core before running init_by_lua* unless lua_use_resty_core off

Is this an existing feature or a suggestion to solve those concerns? If so, I was about to suggest the same behavior as well. (and disable it by default when not using LuaJIT).

agentzh:
@thibaultcha It's a suggestion. I think we'd also drop the support for the standard Lua 5.1 interpreter in ngx_lua and openresty. LuaJIT is already diverged a lot from it and it's becoming a big burden for us due to the lack of FFI in the standard Lua 5.1 interpreter.

In addition, the user can selectively choose which `resty.core.*` submodules to load to reduce memory footprint and GC overhead even further, as in

```nginx
lua_use_resty_core regex shdict;  # only load the modules resty.core.regex and resty.core.shdict
```

By default, it could be `lua_use_resty_core all`. And the user can configure `lua_use_resty_core none` to disable everything :)

agentzh:
@thibaultcha Alas. We should have created a separate ticket for it.

agentzh:
@thibaultcha Let's move our discussion over to #949 to avoid off-topic :) Thanks!

src/ngx_http_lua_shdict.c
splitice: Those are different code paths. Not adjusting the exptime (<0 aka -1 the default for null) vs setting it to not expire (0)

t/043-shdict.t
splitice: Feel free to make adjustments.

splitice:
Like this commit there has been a PR for lua-resty-core for 2 years. It needs a rebase & for the argument order to be re-ordered. 

There is also a PR for the stream module that needs upgrading for the argument order same as this one.

agentzh:
@splitice For your pull requests to be merged, they have to pass the review process. Thank you for your understanding :)

src/ngx_http_lua_shdict.c
agentzh: @splitice Okay, got it.

t/043-shdict.t
agentzh: @splitice You want me to make adjustments in your branch? Sorry, no, I can't, obviously :)

agentzh: @splitice We can only help with the review process since we have other much higher priorities. Thanks for your understanding.

agentzh:
@splitice So please do not blame us for not merging your old PRs for years. They just never passed the review process like the lack of test cases and etc in your old PRs.

agentzh:
@splitice I do appreciate your contributions a lot and I also hope that you can respect our review process and PR merging policies. Thank you!

agentzh:
@splitice We're already trying our best to collaborate and cooperate with our contributors. I hope that you can see that from our active review feedback and etc.

agentzh:
@splitice Consider that these are all volunteering work.

splitice:
And I too am volunteering my time to merge upstream suitable features developed on projects I am involved with. We are all volunteers here. It's fine to do a review but the constant back and forth over small style fixes that would have taken you less time to fix than to comment on is just wasteful. Due to the time wasted by this not even being looked at that I am now investing time in rebasing & re-ordering the parameters (the new order is only required due to the time-lag, and IMHO less consistent ordering considering set/add).

At the end of the day its your project and you can accept changes however you please. But just take a look at the ancient PRs, plenty of good features abandoned there. Plenty that even look complete.

As this was submitted 2-3 years ago and never processed, none of our other patches have been even considered for pushing upstream. To give you an idea of the PRs that you have missed out on from this project alone:
 - ngx.redirect is missing support for HTTP See Other
 - user_flags in incr's return value consistent with ```set``` (quite useful for setting targets and some atomic structures)

Now back to this PR, you can quite easily make any minor style changes you so desire. You should know how, merge it into a feature branch, add a commit and then merge to master. Done and dusted. Going back and forth with minor style changes over years is only wasting both of our times. I'm sure it takes you longer to write your comment than to add a newline.

Now I could go change all the newly added tests, I could also upgrade the existing with the same regex as a separate PR but realistically it's a waste of time as it looks like this code will be replaced with resty-core soon so not worth the time investment. It will either require a re-base, or removed depending on how the code ends up structured. Either way.

Continuing to consider this a fork is fine for us at this point as the path of least resistance.

thibaultcha:
This function does not exist anymore since we removed the CFunction implementation (as originally discussed in this PR!) in https://github.com/openresty/lua-nginx-module/commit/947fa0088b7a0387f43eb016436a2e4462eb5746. The FFI implementation of this API has support for the `init_ttl` argument. Therefore, I will be closing this PR. Thanks for the patch!

splitice:
FYI init_ttl != expire for anyone using my fork (which I continue to maintain). init_ttl does what it name says while expire continues to bump the TTL (i.e for LRU pattern)

thibaultcha:
@splitice Good to know (and important distinction), thanks for clarifying.

PR #944, opened 2017-01-03
Remove duplicate ngx_http_request_t check in FFI var handlers

p0pr0ck5:
lua-resty-core already checks for a non-null 'r' value, so we
don't need to do it here as well.


agentzh:
I prefer not relying on the caller Lua code since these FFI C API functions may have other out-of-tree users.

p0pr0ck5:
Ok. Figured it might be worth some cleanup since there are other FFI C API functions that don't check for `r` (such as https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_ctx.c#L135). But not a big deal for me. Thanks :)

PR #948, opened 2017-01-09
Highlight the issue with escape sequences

solugebefola:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -5699,6 +5699,7 @@ ngx.re.match
 **context:** *init_worker_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
 
 Matches the `subject` string using the Perl compatible regular expression `regex` with the optional `options`.
+[Known Issue With Special Escaping Sequences](#special-escaping-sequences)
 
 Only the first occurrence of the match is returned, or `nil` if no match is found. In case of errors, like seeing a bad regular expression or exceeding the PCRE stack limit, `nil` and a string describing the error will be returned.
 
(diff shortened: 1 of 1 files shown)


agentzh:
@solugebefola Why not just use all those new `*_by_lua_block` directives which are *free* of any of such escaping issues? Those `*_by_lua` directives are obsolete and not recommended to use nowadays.

solugebefola:
I am using `*_by_lua_block`, and I still have to put in one backslash escape to use metacharacters (though not two as noted in the issue link).  I don't know if it is the module version I am using perhaps?  I installed through homebrew on macOS which is v.0.10.6

agentzh:
That's for Lua escaping rules, not the regex. You just need to use Lua's long brackets for regexes.

solugebefola:
Thanks for the tip.  I think some clarification may be useful in the docs for that.

agentzh:
@solugebefola There is an already a warning (in bold) at the beginning of every `*_by_lua` directive in the docs, for example,

https://github.com/openresty/lua-nginx-module#content_by_lua

To quote,

> **WARNING** Since the v0.9.17 release, use of this directive is discouraged; use the new content_by_lua_block directive instead.

thibaultcha:
Would it make sense to implement a log stating their deprecation when one uses the old `*_by_lua` directives? Or do we consider the warning in the documentation to be sufficient?

agentzh:
@thibaultcha We still have a lot of old enterprise code bases using the old directives. So it would be annoying to issue warnings directly to the error logs when these enterprise users upgrade OpenResty.

solugebefola:
@agentzh I didn't mean a warning about the use of `*_by_lua` directives but rather about using long brackets within `ngx.re.match` and other regex functions.

agentzh:
@solugebefola There is already a pointer in the ngx.re API's documentation, for example,

https://github.com/openresty/lua-nginx-module/#ngxrematch

To quote,

> Known Issue With Special Escaping Sequences

agentzh:
@solugebefola And it's more about Lua syntax basics and I don't think we should spread it all over ngx_lua module's documentation.

solugebefola:
Yah, that pointer was what I originally moved to the top of the API docs in this PR.

agentzh:
@solugebefola But that section is really more about the old `*_by_lua` directives. The use of long brackets are really Lua standard syntax and has nothing to do with this module.

solugebefola:
I think maybe, more than a discussion of Lua syntax, maybe an example of using a metacharacter either double-escaped or with long brackets in the doc itself?  The examples shown all have exact text matching.

agentzh:
@solugebefola I think one thing we could improve here is that we should always use long brackets for any regexes our own example Lua code. Are you willing to propose a patch for that?

solugebefola:
Sure I will try to do that.  Woot!

solugebefola:
Just to confirm (as I am new to Lua), long brackets take the place of quotes.

agentzh:
@solugebefola Yeah, good idea! We should also have examples using regexes with `\d` or `\w` or `\s`, for example :)

agentzh:
@solugebefola Yes. When in doubt:

http://www.lua.org/manual/5.1/manual.html#2.1

For example,

```lua
local res = ngx.re.match("hello 1234", [[ (\w+) \s+ (\d+) ]], "jox")
```

solugebefola:
ok, great.  I will get on this in the next couple of days.

agentzh:
A runnable example from the command line:

```console
$ resty -e 'local res = ngx.re.match("hello 1234", [[ (\w+) \s+ (\d+) ]], "jox") print(res[1]) print(res[2])'
hello
1234
```

agentzh:
BTW, `resty` is the command-line utility provided by the `resty-cli` project or the OpenResty bundle.

Issue #951, opened 2017-01-12
coroutine.resume and error logging


bungle:
As far as I know, `coroutine.resume` should behave about the same as `pcall` regarding to errors, but this code gives different results:

```lua
local function a()
    error "ohno"
end
print(coroutine.resume(coroutine.create(a)))
print(pcall(a))
```

Here are different outputs.

```bash
bungle@mbp13tb:~$ lua a.lua 
false	a.lua:2: ohno
false	a.lua:2: ohno
bungle@mbp13tb:~$ luajit a.lua 
false	a.lua:2: ohno
false	a.lua:2: ohno
bungle@mbp13tb:~$ resty a.lua 
2017/01/12 22:43:20 [error] 19716#0: *2 lua coroutine: runtime error: a.lua:2: ohno
stack traceback:
coroutine 0:
	[C]: in function 'error'
	a.lua:2: in function <a.lua:1>
coroutine 1:
	[C]: in function 'resume'
	a.lua:5: in function 'file_gen'
	init_worker_by_lua:36: in function <init_worker_by_lua:34>
	[C]: in function 'pcall'
	init_worker_by_lua:43: in function <init_worker_by_lua:41>, context: ngx.timer
false	a.lua:2: ohno
false	a.lua:2: ohno
```

The last example with `resty` writes to error log when using `coroutine.resume` but doesn't write to error log when using `pcall`. Is there some specific reason for that? I think it is better if the `resty`(and OpenResty) just does the same as with `pcall` when calling `coroutine.resume` aka no logging.

agentzh:
@bungle Yeah, it's better to behave the same. Mind to propose a pull request for this? ;)

bungle:
@agentzh, yes, but do you know if this should behave differently with "main" coroutine?

agentzh:
@bungle You mean the main light thread, right? For all light threads (not just the main ones), they should not capture any exceptions due to their asynchronous nature.

bungle:
@agentzh,

The problem is here:
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_util.c#L1440-L1441

But if I comment that out, this code doesn't log anything:

```lua
local function a()
    error "this is just a return value for coroutine.resume & pcall"
end
print(coroutine.resume(coroutine.create(a)))
print(pcall(a))
error 'this should be logged as an error in nginx logs'
```

I think the last line here should log normally that error.

[...]

PR #952, opened 2017-01-12
No error logging on coroutine.resume (pull-request for issue #951)

bungle:

See more details on #951.

### DIFF
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -1437,9 +1437,6 @@ user_co_done:
 
             ctx->cur_co_ctx = next_coctx;
 
-            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                          "lua coroutine: %s: %s\n%s", err, msg, trace);
-
             /* try resuming on the new coroutine again */
             continue;
         }
--- a/t/091-coroutine.t
+++ b/t/091-coroutine.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 3 + 5);
+plan tests => repeat_each() * blocks() * 3 + 4;
 
 $ENV{TEST_NGINX_RESOLVER} ||= '8.8.8.8';
 
--- a/t/091-coroutine.t
+++ b/t/091-coroutine.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * blocks() * 3 + 4;
+plan tests => repeat_each() * (blocks() * 3 + 2);
 
 $ENV{TEST_NGINX_RESOLVER} ||= '8.8.8.8';
 
@@ -281,7 +281,7 @@ GET /lua
               end)
             end
 
-            N = 10
+            N = 10 
             x = gen(N)		-- generate primes up to N
             while 1 do
               local n = x()		-- pick a number until done
@@ -470,13 +470,13 @@ done
 --- config
     location /lua {
         content_by_lua '
-            local f = function(cr) coroutine.resume(cr) end
+            local f = function(cr) return coroutine.resume(cr) end
             -- emit a error
             local g = function() unknown.unknown = 1 end
             local l1 = coroutine.create(f)
             local l2 = coroutine.create(g)
-            coroutine.resume(l1, l2)
-            ngx.say("hello")
+            local ok = coroutine.resume(l1, l2)
+            ngx.say(ok and "hello" or "error")
         ';
     }
 --- request
@@ -1315,3 +1315,4 @@ co yield: 2
 
 --- no_error_log
 [error]
+
--- a/t/091-coroutine.t
+++ b/t/091-coroutine.t
@@ -475,8 +475,8 @@ done
             local g = function() unknown.unknown = 1 end
             local l1 = coroutine.create(f)
             local l2 = coroutine.create(g)
-            local ok = coroutine.resume(l1, l2)
-            ngx.say(ok and "hello" or "error")
+            local o, k, e = coroutine.resume(l1, l2)
+            ngx.say(o and not k and "hello" or e)
         ';
     }
 --- request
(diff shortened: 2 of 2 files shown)


agentzh:
@bungle Will you add some tests to the existing test suite to justify this change? Thank you!

agentzh:
@bungle Also, please ensure the Travis CI checks are passing in this PR.

bungle:
Yes, I will do that tomorrow.

bungle:
@agentzh, it seems to pass the tests now. I don't think we need another tests for this as I did change the original tests to work with the patch, and they also test that now error is written to log on `coroutine.resume` where resumed `thread` errors.

t/091-coroutine.t
```diff
@@ -483,8 +483,8 @@ done
 GET /lua
 --- response_body
 hello
---- error_log eval
-["stack traceback:", "coroutine 0:", "coroutine 1:", "coroutine 2:"]
+--- no_error_log
+[error]
```
agentzh: Hmm, the indentation of this test case is to test the Lua land backtrace. After your change, we no longer have any tests designed for this purpose.

I'm not sure if silencing the error logging is a good idea. Maybe we can make it a lower log level like "warn" or even "info"?

t/091-coroutine.t
```diff
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 3 + 5);
+plan tests => repeat_each() * blocks() * 3 + 4;
```
agentzh: Better keep the parens otherwise we need to update this plan every time we change the repeat_each number. Also, the repeat each number is automatically increased on the "check leak" test mode of Test::Nginx::Socket, and your change will break the test plan in that test mode.

t/091-coroutine.t
agentzh: Also, better check the return values of `coroutine.resume` for better test coverage.

t/091-coroutine.t
```diff
@@ -281,7 +281,7 @@ GET /lua
               end)
             end
 
-            N = 10 
+            N = 10
```
agentzh: Better avoid unrelated changes to make your changeset atomic. It's better for version control.

t/091-coroutine.t
```diff
@@ -1315,4 +1315,3 @@ co yield: 2
 
 --- no_error_log
 [error]
-
```
agentzh: This is an unrelated change.

t/091-coroutine.t
bungle: I think my editor did it automatically.

t/091-coroutine.t
bungle: Yes, the editor did this automatically. I will add the new line back.

t/091-coroutine.t
bungle: `coroutine.resume` is a `pcall` and its purpose is to catch the errors aka silence the errors and let users handle them. I will make that test better so it checks the return value.

bungle:
One thing though, is this code path only used with when user's code calls `coroutine.resume`, and not for say . Nginx managed light threads e.g. `ngx.thread` or for the main light thread.

src/ngx_http_lua_util.c
```diff
@@ -1437,9 +1437,6 @@ ngx_http_lua_run_thread(lua_State *L, ngx_http_request_t *r,
 
             ctx->cur_co_ctx = next_coctx;
 
-            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
-                          "lua coroutine: %s: %s\n%s", err, msg, trace);
```
agentzh: As I mentioned earlier, should we just lower the log level here? Like `info`?

agentzh:
@thibaultcha @doujiang24 What do you think about this?

src/ngx_http_lua_util.c
bungle: If we do so, this will be different than what Lua / LuaJIT does. They don't write anything. I mean, OpenResty doesn't log anything when wrapping code with `pcall`. If you want to log errors even when they are raised in a code wrapped with `pcall`, then I would consider having this logged as well, and of course we need to change `pcall` then as well. This would differ from what users' expect, though. My only concern is that I'm not sure if that removed log entry removes error logging also from some case where it shouldn't. That's why I asked do anyone of you know if it is only handling when users' code calls `coroutine.resume` (that's the only place where this logging should be silenced).

agentzh: @bungle `pcall` is a more explicit `try/catch` in the Lua world while many Lua programmers may overlook the error handling in `coroutine.resume`.

bungle: @agentzh, I agree, but that's how the Lua is. Those that use `coroutine.resume` should know that it is effectively a `pcall`. There is some history to this as `pcall` has not been fully compatible with coroutines (http://lua-users.org/wiki/PcallAndCoroutines).

thibaultcha:
Sorry for the delay, I was on a trip.

I have spent some time testing this change as well as giving some thoughts about @bungle's concerns, and so far here is what I gathered:

- As we try to stay as close as possible as the standard Lua/LuaJIT's behavior, I believe this change to be aligned with this goal. We should not unexpectedly write to the error logs when Lua would not write to stderr itself.
- Ultimately, lowering the log level would be a compromise, but as with all compromises, it would leave both parties unsatisfied :)
- As for @bungle's concern in #951 about not providing a stacktrace to the `error()` call, I tried some cases on my side and it would seem that this behavior is particular to resty-cli, which executes all user code in a `pcall()`, as seen [here](https://github.com/openresty/resty-cli/blob/master/bin/resty#L353). I would be in favor of replacing this with an `xpcall` in order to provide a full stacktrace to the user and avoid confusions, but that is a different topic than this discussion.

So far with this patch, `error()` writes the appropriate errors and stacktraces from the various contexts I tried it in (`content`, `log`, `timer` and user threads) and indeed removes the logging in a coroutine. I feel like more testing would be necessary though.

agentzh:
Okay, I'm fine with removing the logging in this case completely. We just need to ensure that a backtrace is still generated in other cases, i.e., outside a user `coroutine.resume()` call. In addition, we can avoid generating the backtrace in the `coroutine.resume()` case if we can since the backtrace is later discarded completely anyway and the generation is a little bit expensive.

thibaultcha:
Done as part of #1239, closing this PR, thanks!

PR #957, opened 2017-01-16
feature: add client SSL certificiate support

detailyang:

@agentzh This feature is talked at  [openresty/issues/222](https://github.com/openresty/openresty/issues/222) and be listed in [TODO](https://github.com/openresty/lua-nginx-module#todo) list

Now I just add the api and tests but lack document. When we think it can be merged to master, i will supplement the api usage later.

Please help me review this PR, many thanks :grin:


t/129-ssl-socket.t
```diff
@@ -2618,3 +2676,410 @@ qr/\[error\] .* ngx.socket sslhandshake: expecting 1 ~ 5 arguments \(including t
 --- no_error_log
 [alert]
 --- timeout: 5
+
+
+
+=== TEST 33: setsslcert, too many arguments
```
agentzh: These new tests should go to a separate test file. This `129-ssl-socket.t` file is already quite large.

t/129-ssl-socket.t
```diff
@@ -2618,3 +2676,410 @@ qr/\[error\] .* ngx.socket sslhandshake: expecting 1 ~ 5 arguments \(including t
...
+                return content
+            end
+
+            local sock = ngx.socket.tcp()
+            sock:settimeout(3000)
+
+            local cert = read_file("$TEST_NGINX_HTML_DIR/client.crt")
+            local key = read_file("$TEST_NGINX_HTML_DIR/client.unsecure.key")
+
+            local ok, err = sock:setsslcert(cert, key)
```
agentzh: I'm thinking about a different Lua API than this. Loading the raw certificates upon every SSL cosocket is quite expensive and the client-side certificates tend to be reused over and over again. I think we can introduce an opaque SSL ctx object that loads the certificates only once and for each new SSL cosocket, we just feed the SSL ctx object to it directly. Other custom SSL configurations added in the future can be configured in a similar way. What do you think?

t/129-ssl-socket.t
detailyang: well, how about 151-ssl-client-setcert.t?

t/129-ssl-socket.t
detailyang: @agentzh I agree ` Loading the raw certificates upon every SSL cosocket is quite expensive and the client-side certificates tend to be reused over and over again.` So how about the new directive `lua_ssl_client_key` and `lua_ssl_client_certificate` to feed  SSL ctx object to it directly? Also can we provide this Lua API to allow user to set different certificates on the fly?

So sorry to long time  reply:(

detailyang:
@agentzh so sorry to reply.

I know you are thinking a different Lua API.  Can you explain to me what's the different Lua API so that to change this codebase? Now I have enough time to follow your minds.

By the way, happy new year 2017:)

agentzh:
@detailyang I'm not exactly sure at this point. Maybe something like this?

```lua
local ssl = require "ngx.ssl"
local ctx = ssl.create_ctx{
    client_certificate = "pem-data",
}

-- we can cache the ctx object in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx)
sock:sslhandshake(...)
```

@lziest @ghedo @doujiang24 What do you think?

detailyang:
@agentzh well, it looks like not bad. I will attempt to implement it in localhost

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +33,112 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+    if (cbio == NULL) {
+        ngx_ssl_error(NGX_LOG_ERR, log, 0, "BIO_new_mem_buf() failed");
+        goto done;
+    }
+
+    /*
+    * Reading the PEM-formatted certificate from memory into an X509
+    */
+
+    x509 = PEM_read_bio_X509(cbio, NULL, 0, NULL);
```
ghedo: Given that we already have PEM (and DER) parsing functions, should users maybe use those and pass the already parsed chain to this function?

ghedo:
@agentzh yes, that looks better, since it would avoid setting the certificate for the global SSL_CTX (so other SSLs would get the same configuration). An alternative API would be having a `ngx_http_lua_socket_tcp_sslhandshake_with_ctx()` which would be passed an ctx explicitly, and then reimplement `ngx_http_lua_socket_tcp_sslhandshake()` so that it would only call that. Would still need something to create the SSL_CTX in the first place.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +33,112 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+        goto done;
+    }
+
+    pbio = BIO_new_mem_buf((char *)priv_key->data, priv_key->len);
+    if (pbio == NULL) {
+        ngx_ssl_error(NGX_LOG_ERR, log, 0, "BIO_new_mem_buf() failed");
+        goto done;
+    }
+
+    pkey = PEM_read_bio_PrivateKey(pbio, NULL,
```
ghedo: As above. We don't yet support parsing keys with passwords, but that should be added to the existing API instead IMO, so it would be more generally useful (or better yet, add an additional function (e.g. `ngx_http_lua_ffi_parse_pem_priv_key_with_password()`) to do just that.

t/129-ssl-socket.t
ghedo: Should be fine.

src/ngx_http_lua_ssl.c
ghedo: Also, I think this should be a separate PR, so this one would get smaller and easier to review.

t/129-ssl-socket.t
ghedo: I understand the loading part that @agentzh as actually parsing the certificates (which would be solved by re-using the existing API as I already commented).

However I also think that adding the certificate to the "global" SSL_CTX is wrong, since that would affect any other SSL objects created from it (past and present), so having an API to create a new SSL_CTX and apply configurations to it would be best.

agentzh:
@ghedo Agreed :)

detailyang:
@ghedo @agentzh Thanks for yours eager suggestion. I guess I need more time to finish this feature (so sorry for my level). So just want to  finish these easy parts pointed by @ghedo (parsing the private key with password), then do other parts.

I wil create another PR to add the existing API (ngx_http_lua_ffi_parse_pem_priv_key) or creating additional api (ngx_http_lua_ffi_parse_pem_priv_key_with_password)  for parsing the private key with password.

What do you thinks of this?

src/ngx_http_lua_ssl.c
detailyang: Yesterday I have try to reuse the already parsed chain. Finally I have found these data is type of Luajit's cdata and we cannot reference these data in Lua CFunction.:(

detailyang:
@agentzh 

These days I have try to implement the api like what you said:

```

local ssl = require "ngx.ssl"
local ctx = ssl.create_ctx{
    client_certificate = "pem-data",
}

```

The first version  I try to use Luajit FFI to implement this, But I found the Luajit cdata cannot be used in Lua cfunction `ngx_http_lua_socket_tcp_sslhandshake `. So I eliminated this way.

The second version I try to use Lua CFunction to implement this as the following:

```
local ssl_ctx = ngx.ssl.create_ctx({
    client_certificate = "pem-data",
})

```

The API `ngx.ssl.create_ctx` return the userdata point to the `SSL_CTX *`. To reuse the API `ngx_int_t ngx_ssl_create_connection(ngx_ssl_t *ssl, ngx_connection_t *c, ngx_uint_t flags)`  in `ngx_http_lua_socket_tcp_sslhandshake`, these are two ways to get the `ngx_ssl_t` as the following:
 
# Reuse u->conf->ssl

```

// if we have custom ssl_ctx
saved_ssl_ctx = u->conf->ssl->ctx;
// replace the u->conf->ssl ctx
u->conf->ssl->ctx = custom_ssl_ctx;

if (ngx_ssl_create_connection(u->conf->ssl, c,
                              NGX_SSL_BUFFER|NGX_SSL_CLIENT)
    != NGX_OK)
{
    // restore the ssl ctx
    u->conf->ssl->ctx = saved_ssl_ctx;
    lua_pushnil(L);
    lua_pushliteral(L, "failed to create ssl connection");
    return 2;
}

// restore the ssl ctx
u->conf->ssl->ctx = saved_ssl_ctx;

```

# Create ngx_ssl_t on stack

```

// if we have custom ssl_ctx

ngx_ssl_t ssl;
ssl.buffer_size = u->conf->ssl->buffer_size;
ssl.log = u->conf->ssl->log;
ssl.ctx = custom_ssl_ctx;

if (ngx_ssl_create_connection(&ssl, c,
                              NGX_SSL_BUFFER|NGX_SSL_CLIENT)
    != NGX_OK)
{
    lua_pushnil(L);
    lua_pushliteral(L, "failed to create ssl connection");
    return 2;
}

```
 
# Create ngx_ssl_t on heap by lua_newuserdata

```

```

 I have to create the `ngx_ssl_t`. So I can choose to the `ngx_ssl_t` on the heap or stack as the following:

1. stack:

detailyang:
@agentzh @lziest @ghedo


Hello, guys. I have successfully run a demo for Lua API to feed `SSL Secure Context` Object. For reduce waste yours time to review, I wish we can agree on the Lua API design.

Recently I I have try to  implement two versions Lua API to feed `SSL_CTX*` object as the following:

1. userdata version
```lua
local ctx = ngx.ssl.create_ctx{
    client_certificate = "pem-data",
}

-- we can cache the ctx object in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx)
sock:sslhandshake(...)
```

This version `ssl.create_ctx` will return `Userdata` from Lua stack, it is enough. But if we want to feed more arguments to `SSL_CTX *` in future,  we will have to add more option for  `options table`   and the `create_ctx` CFunction will become more and more large.


2. table version like ngx.socket.tcp

```lua
local ctx = ngx.ssl.ctx()
local ok, err = ctx:init({
    method = 'xxx', -- the openssl protocol methods
    cert = 'xxx', -- cert chains in PEM format 
    key = 'xxx', --  private keys in PEM format
    cacert = 'xxx'  -- the trusted CA certificates
})

if not ok then
    return error("init ssl ctx error: " .. err)
end

-- we can cache the ctx table in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx)
sock:sslhandshake(...)
```

This version `ssl.create_ctx` will return `Table` from Lua stack, it wraps `SSL_CTX *` to table. This is more extended. If we want to feed more arguments to `SSL_CTX *`, we just add new CFunction to ctx table metatable like the following:

```lua
local ctx = ngx.ssl.ctx()
local ok, err = ctx:init({
    method = 'xxx', -- the openssl protocol methods default SSLv23_method
    cert = 'xxx', -- cert chains in PEM format 
    key = 'xxx', --  private keys in PEM format
    cacert = 'xxx' -- the trusted CA certificates in PEM format
})

if not ok then
    return error("init ssl ctx error: " .. err)
end

--we can add more api in future as the following

ctx:set_key()
ctx:set_cert()
ctx:add_cacert()
ctx:add_crl()
ctx:set_ciphres()
ctx:set_ecdhcurve()
ctx:set_dhparam()

```

Also I refer to the [Node.js JS API Design](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options) and [Node.js C++ API Design](https://github.com/nodejs/node/blob/master/src/node_crypto.cc#278) as a guide to our Lua API.


Which one is better? Need yours feedback and  let us make a choice :D

detailyang:
Ping @agentzh @thibaultcha  @lziest @ghedo. Can you have a look at the above, If you are free now. Thanks :D

agentzh:
@detailyang Sorry for the delay on my side. Been massively distracted. Regarding the API design, I don't quite like the `ctx.init()` syntax. It should be in the `ngx.ssl` module of lua-resty-core. As I've demonstrated before:

```lua
local ssl = require "ngx.ssl"
local ctx = ssl.create_ctx{
    client_certificate = "pem-data",
}

-- we can cache the ctx object in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx)
sock:sslhandshake(...)
```

But yeah, we'll have to implement the `setsslctx` function atop FFI instead of as a CFunction.

The `ctx` data returned by `ssl.create_ctx()` is an opaque pointer as an FFI cdata object.

detailyang:
@agentzh Thanks for reply me. I have  try to implement the `setsslctx` function atop FFI. But I'm failed because the `tcp:sslhandshake` is the CFunction. I'm confused that how can we pass the `ctx` cdata to `tcp:sslhandshake` CFunction. Can you Give me a little hint ?

Issue #938, opened 2016-12-23
development proposal: ngx.print with index parameters


hkmo:
Hi,
My proposal is related to the following API function:

ngx.print
syntax: ok, err = ngx.print(...)

We are using this function in conent_by_lua phase.
I would like to be able to print starting from an index x to index x+k without having to create a new string.
Today we are doing this by using lua string.sub() and this seems to create a performance hit according to our testing.

I am going to look at the code and see how this could be added.

Please let me know if you have any comment.

Best Regards

agentzh:
@hkmo I suggest we introduce a new API function that simply takes an FFI cdata buffer as argument.

hkmo:
@agentzh what do you mean by FFI cdata buffer? How does this relate to the existing code? How would the ngx.lua interface look like?

agentzh:
@hkmo Something like below, maybe?

```lua
local output = require "ngx.output"  -- this Lua module can sit inside lua-resty-core.
local ffi = require "ffi"  -- this is the standard LuaJIT module

-- below is the cdata buffer which can be shared on the nginx worker level
local buf = ffi.new("char[256]")

local len = math.min(256, #some_lua_string)
ffi.copy(buf, some_lua_string, math.min(256, #some_lua_string))

-- the output.ffi_write() can be implemented atop FFI as well for maximum performance.
local ok, err = output.ffi_write(buf, len)

-- pointer arithmetic also works, for example.
local offset = 128
local ok, err = output.ffi_write(buf + offset, len)
```

hkmo:
@agentzh In your proposal I see a create buffer and a copy ffi.copy() from the lua string to the buffer. So, maybe the overhead is still similar to string.sub()? How is it algorithmically better? Is it because we are copying by chunk of 256 max?

I was thinking about simply adding an extra API function signature print_() or write() in ngx lua module that will accept two extra parameters; offset and length. 

Thanks

agentzh:
@hkmo You can avoid data copying on the Lua land altogether and pass the Lua strings directly with an offset, as in

```
local ffi = require "ffi"
local ffi_cast = ffi.cast
local const_char_ptr = ffi.typeof("const char *")

local ok, err = output.ffi_write(ffi_cast(my_lua_string, const_char_ptr) + offset, len)
```

But we can also add an offset parameter to make the caller code much cleaner, as in

```
local ok, err = output.ffi_write(my_lua_string, offset, len)
```

[...]

PR #959, opened 2017-01-18
Update ngx_http_lua_common.h

hkmo:
ffi_write API 
see https://github.com/openresty/lua-nginx-module/issues/938

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -119,8 +119,17 @@ typedef struct {
 
 
 #ifndef NGX_LUA_NO_FFI_API
-#define NGX_HTTP_LUA_FFI_NO_REQ_CTX         -100
-#define NGX_HTTP_LUA_FFI_BAD_CONTEXT        -101
+#define NGX_HTTP_LUA_FFI_EOF                110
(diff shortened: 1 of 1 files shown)


agentzh:
@hkmo I don't see how your changes in this PR are related to #938. Please elaborate.

hkmo:
@agentzh I added new constants for returned error codes by "ngx_http_lua_ffi_write()". 
But, we can use error messages instead that we pass back to the caller, similarly to other FFI functions like "balancer.set_current_peer()". I just wanted to avoid the overhead of error buffer.

agentzh:
@hkmo Constant error strings have *zero* buffer overhead. They are already intern'ed Lua strings, just like integer constants.

agentzh:
Also, constants defined in this modules' C header files are not visible nor usable by FFI and lua-resty-core anyway.

agentzh:
@hkmo I think you should stick with the existing NGINX error codes like `NGX_DECLINED`, `NGX_ERROR`, `NGX_BUSY`, and etc. Too many detailed error codes make little sense in the real world and also complicates the caller code significantly.

hkmo:
Didn't see these comments. I agree for the error codes. Already made these chsnges obsolete, andI was looking to cancel this PR #959.

PR #960, opened 2017-01-18
ffi_write new API: Update ngx_http_lua_output.c

hkmo:
ffi_write new API function.

This new FFI API function is used to print a string buffer on the client output starting from a given offset and ending at given length. This avoids the overhead of creating another string in the lua land.

See https://github.com/openresty/lua-nginx-module/issues/938 for more details.

### DIFF
--- a/src/ngx_http_lua_output.c
+++ b/src/ngx_http_lua_output.c
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
     ctx->flushing_coros--;
 }
 
+#ifndef NGX_LUA_NO_FFI_API
+int
+ngx_http_lua_ffi_write(ngx_http_request_t *r, const char* str, size_t offset, size_t len)
+{
+    ngx_http_lua_ctx_t          *ctx;
+    ngx_uint_t                  size;
+    ngx_buf_t                   *b;
+    ngx_chain_t                 *cl;
+    ngx_int_t                   rc;
+    const char*                 str_sub;
+    u_char                      err[128];
+    size_t                      errlen;
+
+    errlen = 64;
+
+    if (r == NULL) {
+        return NGX_HTTP_LUA_FFI_NO_REQUEST;
+    }
+
+    if (str == NULL) {
+        return NGX_HTTP_LUA_FFI_NULL_STRING;
+    }
+
+    size = ngx_strlen(str);
+
+    if (len == 0 || size == 0 || offset > size - 1) {
+        return NGX_HTTP_LUA_FFI_BAD_RANGE;
+    }
+
+    if (offset + len > size) {
+        len = size - offset;
+    }
+
+    str_sub = str + offset;
+
+    if (r->connection->fd == (ngx_socket_t) -1) {
+        return NGX_HTTP_LUA_FFI_BAD_CONTEXT;
+    }
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
+
+    if (ctx == NULL) {
+        return NGX_HTTP_LUA_FFI_NO_REQ_CTX;
+    }
+
+    rc = ngx_http_lua_ffi_check_context(ctx,
+                    NGX_HTTP_LUA_CONTEXT_REWRITE
+                    | NGX_HTTP_LUA_CONTEXT_ACCESS
+                    | NGX_HTTP_LUA_CONTEXT_CONTENT,
+                    &err[0], &errlen);
+
+    if (rc != NGX_OK) {
+        return NGX_HTTP_LUA_FFI_BAD_CONTEXT;
+    }
+
+    if (ctx->acquired_raw_req_socket) {
+        return NGX_HTTP_LUA_FFI_RAW_SOCKET;
+    }
+
+    if (r->header_only) {
+        ctx->eof = 1;
+        return NGX_HTTP_LUA_FFI_HEADER_ONLY;
+    }
+
+    if (ctx->eof) {
+        return NGX_HTTP_LUA_FFI_EOF;
+    }
+
+    ctx->seen_body_data = 1;
+
+    cl = ngx_http_lua_chain_get_free_buf(r->connection->log, r->pool,
+            &ctx->free_bufs, len);
+
+    if (cl == NULL) {
+        return NGX_HTTP_LUA_FFI_NO_MEMORY;
+    }
+
+    b = cl->buf;
+    b->last = ngx_copy(b->last, (u_char *) str_sub, len);
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+            "lua ffi_write response");
+
+    rc = ngx_http_lua_send_chain_link(r, ctx, cl);
+
+    if (rc == NGX_ERROR || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
+        return NGX_HTTP_LUA_FFI_OUTPUT_ERROR;
+    }
+
+    dd("downstream write: %d, buf len: %d", (int) rc,
+            (int) (b->last - b->pos));
+
+    return NGX_OK;
+}
+#endif /* NGX_LUA_NO_FFI_API */
+
 /* vi:set ft=c ts=4 sw=4 et fdm=marker: */
--- a/src/ngx_http_lua_output.c
+++ b/src/ngx_http_lua_output.c
@@ -802,30 +802,31 @@ ngx_http_lua_flush_cleanup(void *data)
     ctx->flushing_coros--;
 }
 
+
 #ifndef NGX_LUA_NO_FFI_API
 int
-ngx_http_lua_ffi_write(ngx_http_request_t *r, const char* str, size_t offset, size_t len)
+ngx_http_lua_ffi_write(ngx_http_request_t *r, const char *buf, size_t offset, size_t len)
 {
     ngx_http_lua_ctx_t          *ctx;
-    ngx_uint_t                  size;
+    ngx_uint_t                   size;
     ngx_buf_t                   *b;
     ngx_chain_t                 *cl;
-    ngx_int_t                   rc;
-    const char*                 str_sub;
-    u_char                      err[128];
-    size_t                      errlen;
-
+    ngx_int_t                    rc;
+    const char                  *str_sub;
+    u_char                       err[128];
+    size_t                       errlen;
+    
     errlen = 64;
 
     if (r == NULL) {
         return NGX_HTTP_LUA_FFI_NO_REQUEST;
     }
 
-    if (str == NULL) {
+    if (buf == NULL) {
         return NGX_HTTP_LUA_FFI_NULL_STRING;
     }
 
-    size = ngx_strlen(str);
+    size = ngx_strlen(buf);
 
     if (len == 0 || size == 0 || offset > size - 1) {
         return NGX_HTTP_LUA_FFI_BAD_RANGE;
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
     ctx->flushing_coros--;
 }
```
agentzh: Style: 2 blank lines needed here.

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
...
 
+#ifndef NGX_LUA_NO_FFI_API
+int
+ngx_http_lua_ffi_write(ngx_http_request_t *r, const char* str, size_t offset, size_t len)
```
agentzh: Style: `const char* str` should be `const char *str`. Be careful. Always be consistent with the rest of the code base.

agentzh: Also, I think `buf` is a better variable name than `str` here.

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
...
+{
+    ngx_http_lua_ctx_t          *ctx;
+    ngx_uint_t                  size;
+    ngx_buf_t                   *b;
+    ngx_chain_t                 *cl;
+    ngx_int_t                   rc;
+    const char*                 str_sub;
```
agentzh: Again, you put the `*` char in the wrong place.

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
...
 
+#ifndef NGX_LUA_NO_FFI_API
+int
+ngx_http_lua_ffi_write(ngx_http_request_t *r, const char* str, size_t offset, size_t len)
+{
+    ngx_http_lua_ctx_t          *ctx;
+    ngx_uint_t                  size;
+    ngx_buf_t                   *b;
+    ngx_chain_t                 *cl;
+    ngx_int_t                   rc;
```
agentzh: Bad variable alignment. See the rest of the code base.

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,100 @@ ngx_http_lua_flush_cleanup(void *data)
...
+    }
+
+    if (ctx->eof) {
+        return NGX_HTTP_LUA_FFI_EOF;
+    }
+
+    ctx->seen_body_data = 1;
+
+    cl = ngx_http_lua_chain_get_free_buf(r->connection->log, r->pool,
+            &ctx->free_bufs, len);
```
agentzh: Bad indentation.

agentzh:
@hkmo Please use the `ngx-releng` script to analyze coding style issues in your branch. See https://github.com/openresty/openresty-devel-utils/blob/master/ngx-releng

agentzh:
Also, please rebase to the latest master.

hkmo:
@agentzh Do you know why travis is complaining with: "Failed to get the version of the Nginx in PATH"?

agentzh:
@hkmo They were the consequences of the earlier compilation errors. To quote:

```
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c: In function ‘ngx_http_lua_ffi_write’:
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:823:16: error: ‘NGX_HTTP_LUA_FFI_NO_REQUEST’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_NO_REQUEST;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:823:16: note: each undeclared identifier is reported only once for each function it appears in
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:827:16: error: ‘NGX_HTTP_LUA_FFI_NULL_STRING’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_NULL_STRING;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:833:16: error: ‘NGX_HTTP_LUA_FFI_BAD_RANGE’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_BAD_RANGE;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:863:16: error: ‘NGX_HTTP_LUA_FFI_RAW_SOCKET’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_RAW_SOCKET;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:868:16: error: ‘NGX_HTTP_LUA_FFI_HEADER_ONLY’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_HEADER_ONLY;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:872:16: error: ‘NGX_HTTP_LUA_FFI_EOF’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_EOF;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:881:16: error: ‘NGX_HTTP_LUA_FFI_NO_MEMORY’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_NO_MEMORY;
                ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_output.c:893:16: error: ‘NGX_HTTP_LUA_FFI_OUTPUT_ERROR’ undeclared (first use in this function)
         return NGX_HTTP_LUA_FFI_OUTPUT_ERROR;
                ^
make[1]: *** [objs/addon/src/ngx_http_lua_output.o] Error 1
make[1]: Leaving directory `/home/travis/build/openresty/lua-nginx-module/buildroot/nginx-1.9.15'
make: *** [build] Error 2
```

hkmo:
Yes I figured that out yesterday and I made the corrections.

hkmo:
@agentzh I did a load test and I see a lot of "bad range" errors and very slow throughput compared to the normal print. Is it possible the fact we are not casting to const char* in the lua is giving us a bad buffers, or? 

     size = ngx_strlen(buf);      
     if (len == 0 || size == 0 || offset > size - 1) {
          *err = "bad buffer range";        
          return NGX_ERROR;     
      }

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,105 @@ ngx_http_lua_flush_cleanup(void *data)
...
+        *err = "no request object found";
+        return NGX_ERROR;
+    }
+
+    if (buf == NULL) {
+        *err = "no buffer found";
+        return NGX_ERROR;
+    }
+
+    size = ngx_strlen(buf);
```
agentzh: It's wrong to assume that the input data stream is a C string. It could contain binary data like image data.

src/ngx_http_lua_output.c
```diff
@@ -802,4 +802,105 @@ ngx_http_lua_flush_cleanup(void *data)
...
+
+    if (len == 0 || size == 0 || offset > size - 1) {
```
agentzh: Also, for 0-sized buffer, we should just return immediately instead of raising errors.

src/ngx_http_lua_output.c
agentzh: It's the caller's responsibility to ensure the `len` argument is valid.

hkmo: so, how would we check a zero-sized input buffer without strlen?
how would we check an offset outsize of the length of the buffer?

agentzh: @hkmo Just check the caller-supplied `len` argument.

hkmo: @agentzh there seem to be  a weird issue: For troubleshooting I am logging  string.len(lua_string), I see a number bigger than the len passed in argument, but, inside the C module, the strlen(lua_string) is giving a size that fails the check if (len == 0 || size == 0 || offset > size - 1). Any idea?

hkmo: @agentzh The ngx_strlen(lua_string) doesn't seem to work. I should probably remove that check. I will just check the offset and len supplied in the call. But, I will not check the ngx_strlen(lua_string).

agentzh: @hkmo As I've said, `ngx_strlen` is libc's `strlen` per se. It's wrong to assume the input buffer data contains a C string.

hkmo: @agentzh without knowing the size of the object, it would be hard to ffi_write from an offset.

agentzh: @hkmo You should just use the value of the `len` argument provided by the caller. As I've said, you should trust the `len` argument value since it's the caller's responsibility to ensure its validity.

hkmo: @agentzh I think there is a misunderstanding about the API in output.lua (same applies for ffi.cdef):
    
    ffi_write(lua_string, offset, len)
    
    lua_string: the lua string
    offset: the offset in the lua string
    len: the length to print starting from the offset
    
We are missing the length of the lua string. We should probably have the following API function instead:
     
     ffi_write(lua_string, len, from, to)
     
     lua_string: the lua string to write
     len: the length of the lua string
     from: the index from which we start to write
     to: the index to which we stop to write
     
or 
     
      ffi_write(lua_string, lua_string_len, offset, len)
      
      lua_string: the lua string
      lua_string_len: the length of the lua string
      offset: the offset in the lua string
      len: the length to print starting from the offset

hkmo: @agentzh or we can keep output.lua API unchanged and just change the ffi.cdef API by one of the above (add buf_size). This parameter can be passed by the output.lua using "#lua_string".

agentzh: @hkmo Nope, the first argument of `ffi_write` is defined to be a cdata object of the type `const char *` which represents the starting pointer of a C buffer. You can directly feed a Lua string there just because LuaJIT can do the implicit Lua string -> `const char *` cdata casting. The `ffi_` prefix means that this API is unsafe per se so it's the caller's responsibility to ensure the validity of all the function argument values.

agentzh: @hkmo No, the `len` argument is sufficient and the extra `buf_size` argument is useless.

agentzh: @hkmo We should not assume the first argument to `ffi_write` is a Lua string. It could be just a raw C buffer. This is a much more flexible arrangement.

hkmo: @agentzh is there anything I am missing here? we pass a buffer without passing its size and we pass an offset with a len and we assume the ffi_write will not go beyond the buffer size? that means we assume that: offset + len <= buffer size

hkmo: assumption for ffi_write will be: 
     
      **offset** + **len** <= size of **buf**

agentzh: @hkmo I'm not sure where you are from, but in the C world, it's natural that the caller ensures offset + len never exceeds the buffer boundary. Otherwise, the `len` makes little sense. You can check out the `send()` and `write()` syscalls of UNIX, for example.

agentzh: @hkmo It's strange that you keep making me repeat the same thing over and over again here. Alas.

hkmo: @agentzh thanks for the code review comment.

zhuizhuhaomeng:
This PR It's been a long time coming, let's resubmit it after we fix the relevant issues if necessary.

PR #961, opened 2017-01-19
feature: ssl: add FFI function to set SSL ciphers

ghedo:


ghedo:
@agentzh @doujiang24  ping? This seems pretty straightforward.

agentzh:
@ghedo Sorry for the delay on our side. This PR looks good to me :) @doujiang24 ?

agentzh:
@ghedo Have you created a PR for lua-resty-core yet?

ghedo:
Made https://github.com/openresty/lua-resty-core/pull/86 now.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1232,6 +1232,38 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+
+    ssl_conn = r->connection->ssl->connection;
+    if (ssl_conn == NULL) {
+        *err = "bad ssl conn";
+        return NGX_ERROR;
+    }
+
+    if (!SSL_set_cipher_list(ssl_conn, ciphers)) {
+        *err = "SSL_set_cipher_list() failed";
+        goto failed;
```
doujiang24: Seems we can omit the `goto` here？Just move the `failed` code here?

ghedo: Done.

ghedo:
@agentzh @doujiang24 ping? Is there anything else needed here?

ghedo:
@agentzh @doujiang24 it's now been almost 6 months since this and https://github.com/openresty/lua-resty-core/pull/86 were opened, any news?

agentzh:
@ghedo Sorry for the long delay on our side. We'll look into this as soon as we can manage. Thanks for your understanding!

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1236,6 +1236,33 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+        return NGX_ERROR;
+    }
+
+    ssl_conn = r->connection->ssl->connection;
+    if (ssl_conn == NULL) {
+        *err = "bad ssl conn";
+        return NGX_ERROR;
+    }
+
+    if (!SSL_set_cipher_list(ssl_conn, ciphers)) {
```
agentzh: @ghedo I wonder if you really need per-connection ciphers. Maybe you just need a per-domain cipher suite? If it's the latter, then we can just reuse a domain-level SSL ctx to avoid per-request cipher parsing and setting.

agentzh: Or maybe we can simply set a parsed SSL cipher suite C data structure to avoid the string parsing here on a hot code path?

dndx: Yes I agree that calling `SSL_set_cipher_list` on a per request basis is a terrible idea as it both has lots of `malloc` along the way and the cipher [search algorithm is linear](https://github.com/openssl/openssl/blob/master/ssl/ssl_ciph.c#L1060).

Better way of doing this is to store the cipher stack created by OpenSSL once and reuse them later in new connections with some kind of Lua handle (e.g. cdata).

There may be other way that's better but I haven't looked too much into other OpenSSL cipher related APIs.

ghedo: Ooops, missed these comments...

Anyway, the function is intended to be used in `ssl_cert_by_lua`, so per-connection, not per-request, and even then not all connections (otherwise you might as well use `ssl_protocols` in the config).

AFAICT there's no OpenSSL API that would allow you to parse a cipher string into an array or stack of SSL_CIPHER, without actually changing the cipher list on an SSL or SSL_CTX struct. And even then, there's no OpenSSL API to set a cipher list _from_ an array or stack of SSL_CIPHER. Did you have anything specific in mind?

In any case the overhead of `SSL_set_cipher_list()` is really not that bad, and even if it was too slow/heavy for whatever use case (which would likely prevent other OpenResty APIs from being used as well), nobody would be forced to use this.

agentzh: @ghedo Well, I think we are just trying to do it *right* :) BTW, have you actually measured the overhead of parsing and loading long cipher strings in this context?

Webster-Yang:
How to set ciphersuites dynamically？
Can the changes in this PR be used directly on openresty/1.27.1.2?

PR #966, opened 2017-01-26
travis-ci: bump openssl to 1.0.2k

thibaultcha:
Here is the OpenSSL announcement about the release: https://mta.openssl.org/pipermail/openssl-announce/2017-January/000091.html

Tests seem to be ok.


agentzh:
Merged. Thanks!

PR #967, opened 2017-01-27
Fix build compatibility with BoringSSL.

tmthrgd:
There are two issues that occur when lua-nginx-module is compiled against BoringSSL. They are:
+ |SSL_set_tlsext_host_name| changing from a macro that internally casts name to a char* to a function that expects a char*, and
+ |sk_X509_num| returning a size_t and no longer an int.

The first issue is fixed by changing the caller of |SSL_set_tlsext_host_name|. This is identical to the fix applied in nginx/nginx@9b8b33b.

The second issue is fixed by changing variable `i` to be a size_t rather than an int when built against BoringSSL, i.e. when OPENSSL_IS_BORINGSSL is defined.

This change only addresses build failures and warnings and does not attempt to address missing features or correctness.

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -1335,7 +1335,7 @@ ngx_http_lua_socket_tcp_sslhandshake(lua_State *L)
 #ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
 
                 if (SSL_set_tlsext_host_name(c->ssl->connection,
-                                             (char *)name.data)
+                                             (char *) name.data)
                     == 0)
                 {
                     lua_pushnil(L);
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -1334,7 +1334,8 @@ ngx_http_lua_socket_tcp_sslhandshake(lua_State *L)
 
 #ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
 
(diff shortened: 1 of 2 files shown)


tmthrgd:
I suppose this takes over from the, incomplete and closed, pull request #878.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1334,7 +1334,8 @@ ngx_http_lua_socket_tcp_sslhandshake(lua_State *L)
 
 #ifdef SSL_CTRL_SET_TLSEXT_HOSTNAME
 
-                if (SSL_set_tlsext_host_name(c->ssl->connection, name.data)
+                if (SSL_set_tlsext_host_name(c->ssl->connection,
+                                             (char *)name.data)
```
agentzh: Style: I think we need a space after `)` here.

tmthrgd: My bad. Fixed.

agentzh:
@ghedo What do you think of this? ;)

ghedo:
LGTM, but note that e.g. `ngx_http_lua_ffi_ssl_get_tls1_version()` (and related Lua function) won't work anymore because BoringSSL doesn't define `TLS1_get_version()`.

It might be worth replacing `TLS1_get_version()` with `SSL_version()`, although it would slightly change the semantics of the FFI function above. We wouldn't lose much IMO.

The SSL-related tests also won't work because of some small differences that BoringSSL introduces.

agentzh:
@ghedo Thanks for your review!

agentzh:
@tmthrgd Merged. Thanks!

PR #972, opened 2017-02-02
bugfix: cdata pointer should check the nullity by == nil

detailyang:

cdata pointer should check the nullity by == nil rather true or false


agentzh:
@detailyang Fixed with minor adjustments in your commit log message. Thanks!

PR #973, opened 2017-02-02
feature: support load private key with password

detailyang:

@agentzh This feature is separated at #957 . Now It adds the FFI API as the following:

```c
int
ngx_http_lua_ffi_priv_key_pem_to_der_with_password(const u_char *pem,
    size_t pem_len, const u_char *pwd, size_t pwd_len, u_char *der, char **err);

void *
ngx_http_lua_ffi_parse_pem_priv_key_with_password(const u_char *pem,
    size_t pem_len, const u_char *pwd, size_t pwd_len, char **err);
```

When we think it can be merged to master, I will create another PR for lua-resty-core to implment load private with password.


src/ngx_http_lua_ssl_certby.c
```diff
@@ -1102,6 +1147,40 @@ ngx_http_lua_ffi_parse_pem_priv_key(const u_char *pem, size_t pem_len,
 }
 
 
+void *
+ngx_http_lua_ffi_parse_pem_priv_key_with_password(const u_char *pem,
+    size_t pem_len, const u_char *pwd, size_t pwd_len, char **err)
+{
+    BIO         *in;
+    EVP_PKEY    *pkey;
+    ngx_str_t   password;
```
agentzh: Coding style issue. Please use the `ngx-releng` script to check your code. Thank you.

t/140-ssl-c-api.t
```diff
@@ -811,3 +817,700 @@ lua ssl server name: "test.com"
 --- no_error_log
 [error]
 [alert]
+
+
+
+=== TEST 6: simple cert + private key with password
```
agentzh: Better create a new `.t` file to avoid growing this test file from forever :)

agentzh: Also, we can simply test the Lua API here instead of the plain C API since lua-resty-core is already visible to this module's test suite.

Finally, I think you can use the new `add_block_preprocessor` feature of `Test::Nginx::Socket` to avoid the massive code duplication in these SSL related tests. See

http://search.cpan.org/dist/Test-Nginx/lib/Test/Nginx/Socket.pm#add_block_preprocessor

src/ngx_http_lua_ssl_certby.c
```diff
@@ -982,6 +982,51 @@ ngx_http_lua_ffi_priv_key_pem_to_der(const u_char *pem, size_t pem_len,
...
 
 
+int
+ngx_http_lua_ffi_priv_key_pem_to_der_with_password(const u_char *pem,
+    size_t pem_len, const u_char *pwd, size_t pwd_len, u_char *der, char **err)
+{
+    int          len;
+    BIO         *in;
+    EVP_PKEY    *pkey;
+    ngx_str_t   password;
```
thibaultcha: idem (missing space)

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +34,35 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
 }
 
 
+int
+ngx_http_lua_ssl_password_callback(char *buf, int size, int rwflag,
```
ghedo: This should probably be moved to `ngx_http_lua_ssl_certby.c` and declared static since it's only used there.

t/140-ssl-c-api.t
ghedo: FWIW, the pattern i've been using is testing the C FFI function in `lua-nginx-module` and test the Lua function in `lua-resty-core`. Otheriwse you create a co-dependency on both repos that makes the tests fail for both lua-nginx-module and lua-resty-core PRs until you merge both. (Assuming this is what you meant in the first paragraph).

src/ngx_http_lua_ssl_certby.c
detailyang: stupid error:(.

src/ngx_http_lua_ssl_certby.c
detailyang: stupid error:(.

src/ngx_http_lua_ssl.c
detailyang: maybe `ngx_http_lua_ssl_password_callback` is a useful function ? I mean that it will be used in other scene like `cosocket` in future.

t/140-ssl-c-api.t
detailyang: @agentzh  Thanks for you suggestion, I will take a look at  `add_block_preprocessor`.

@ghedo what's your meaning? My plan is to add the new api at first. When we let it  merge to master  then creating another PR to `lua-resty-core`. Is that wrong?

detailyang:
@agentzh @ghedo @thibaultcha 

Will you please look at the new codebase, Many thanks :D

t/141-ssl-c-api-password.t
```diff
@@ -0,0 +1,584 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
```
doujiang24: I think it's fine to use `t/151-ssl-c-api-password.t`, so that we don't need to rename `t/141-luajit.t` to `t/151-luajit.t`.
The num order is not important here:)

detailyang: gotcha

src/ngx_http_lua_ssl_certby.c
```diff
@@ -982,6 +982,51 @@ ngx_http_lua_ffi_priv_key_pem_to_der(const u_char *pem, size_t pem_len,
 }
 
 
+int
+ngx_http_lua_ffi_priv_key_pem_to_der_with_password(const u_char *pem,
+    size_t pem_len, const u_char *pwd, size_t pwd_len, u_char *der, char **err)
```
detailyang: @doujiang24 I have considered the `ngx_http_lua_ffi_priv_key_pem`. But there are some existing tests. I'm not sure it's worth to change the `ngx_http_lua_ffi_priv_key_pem`. So  just add the new api `ngx_http_lua_ffi_priv_key_pem_to_der_with_password`, then reimplement the Lua FFI API `ngx.ssl. priv_key_pem_to_der`

So if we almost agree to reimplement the `ngx_http_lua_ffi_priv_key_pem`, I will try to do :D

detailyang:
@doujiang24 @ghedo  @agentzh 

Now the only question is that we can choose to  reimplement `ngx_http_lua_ffi_priv_key_pem_to_der` and `ngx_http_lua_ffi_parse_pem_priv_key` to adding the password param but may introduce the risk because of  doing more change for some tests and `lua-resty-core` or adding the new API `ngx_http_lua_ffi_priv_key_pem_to_der_with_password` and `ngx_http_lua_ffi_parse_pem_priv_key_with_password`

What's your guys suggestion?

guanglinlv:
#750

agentzh:
@detailyang Huh, that is another reason why we shouldn't call into C functions directly through FFI in lua-nginx-module's test suite. We should really only test the lua-resty-core Lua API instead ;)

detailyang:
@agentzh Thanks.
Finally I do understand what's the meaning for 'simply test' when you are always to teach me 'simply test' in some code reviews :D

zhuizhuhaomeng:
we have merged this PR： https://github.com/openresty/lua-nginx-module/pull/1991

PR #974, opened 2017-02-02
move init_by_lua to init_module phase, ensure all shdict has inited

rainingmaster:

can we move init_by_lua to ngx_http_lua_module's init_module phase,  ensure all shdict has inited. in this way, we can use third part shdict without initialize in ngx_http_lua_module.

### DIFF
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -159,8 +159,6 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
     ngx_shm_zone_t              *ozone;
     void                        *odata;
 
-    ngx_int_t                    rc;
-    volatile ngx_cycle_t        *saved_cycle;
     ngx_http_lua_main_conf_t    *lmcf;
     ngx_http_lua_shm_zone_ctx_t *ctx;
     ngx_shm_zone_t              *zone;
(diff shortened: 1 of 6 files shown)


agentzh:
@rainingmaster Why don't you use ngx_http_lua's public C API to register your own shdict storage? It's much more reliable and elegant. See this example module in the test suite:

https://github.com/openresty/lua-nginx-module/tree/master/t/data/fake-shm-module

If you still believe there is a problem, please also add a test case to the existing test suite to cover it. Thank you.

rainingmaster:
@agentzh I have see ngx_http_lua's public C API, it truely useful.
But in some case there is a problem: when we want to [share memary between lua-stream and lua-http](https://github.com/rainingmaster/lua-shdict-nginx-module), we can not ensure every `share memary` has initialized in `init_by_lua` phase. So I think mayby it is better to `share memary` more independence to lua-* ?
On the other hand, maybe the meaning of  `init master` in `ngx_module_t` is more appropriate to `init_by_lua` phase ? :)

agentzh:
@rainingmaster Sharing shdicts between `ngx_stream_lua_module` and `ngx_http_lua_module` require a higher level module called `ngx_meta_lua_module`. Otherwise it's inherently unsafe. This has discussed before in the `ngx_stream_lua_module` github issues board. That's another topic.

And I don't think the `init_master` hook is actually used by grepping through the nginx source tree:

```console
agentzh@fedora64 ~/work/nginx-1.11.2 $ grep init_master -r src/
src/core/ngx_module.h:    ngx_int_t           (*init_master)(ngx_log_t *log);
```

agentzh:
@rainingmaster I agree `init_module` is a better hook for `init_by_lua*` than the current choice.

agentzh:
@doujiang24 @yangshuxin @helloyi Will you have a look at this patch please? Many thanks!

rainingmaster:
@agentzh Sorry, I've wrote the phase wrong. I've use `init_module` in my patch

doujiang24:
@agentzh LGTM :)

rainingmaster:
because of nginx will run `exit(1)` while `init_module`, we can not move `init_by_lua*` to `init_module` phase. as https://github.com/openresty/lua-nginx-module/pull/1002 said.  : (

PR #981, opened 2017-02-10
function declartion and extern variable postion

detailyang:

The [ngx_http_lua_ssl_init defination](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_ssl.c#L20) and [ngx_http_lua_ssl_ctx_index defination](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_ssl.c#L20) is in the 
`#if (NGX_HTTP_SSL)` block.
So The [ngx_http_lua_ssl_init declartion](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_ssl.h#L37) and [extern ngx_http_lua_ssl_ctx_index delcartion](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_ssl.h#L40) also should be in the `#if (NGX_HTTP_SSL)` block, even if it do not impact compile.

### DIFF
--- a/src/ngx_http_lua_ssl.h
+++ b/src/ngx_http_lua_ssl.h
@@ -12,6 +12,8 @@
 
 
 #if (NGX_HTTP_SSL)
+
+
 typedef struct {
     ngx_connection_t        *connection; /* original true connection */
     ngx_http_request_t      *request;    /* fake request */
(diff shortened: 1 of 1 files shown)


agentzh:
Merged. Thanks!

PR #982, opened 2017-02-11
Redo of doc edits to fit agentzh’s quirky setup

dakanji:
- Instances of “Warning” changed to more appropriate terms such as “Note” or “Caution”
- Removed informal terms such as “kinda”
- Converted several markdown backticks to mediawiki format

I hereby grant the copyright of the changes in this pull request
to the authors of the lua-nginx-module project.

### DIFF
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -520,13 +520,26 @@ instead of the old deprecated form:
 
 Here is the reason: by design, the global environment has exactly the same lifetime as the Nginx request handler associated with it. Each request handler has its own set of Lua global variables and that is the idea of request isolation. The Lua module is actually loaded by the first Nginx request handler and is cached by the <code>require()</code> built-in in the <code>package.loaded</code> table for later reference, and the <code>module()</code> builtin used by some Lua modules has the side effect of setting a global variable to the loaded module table. But this global variable will be cleared at the end of the request handler,  and every subsequent request handler all has its own (clean) global environment. So one will get Lua exception for accessing the <code>nil</code> value.
 
-Generally, use of Lua global variables is a really really bad idea in the context of ngx_lua because
+The use of Lua global variables is a generally inadvisable in the ngx_lua context as:
 
-# misuse of Lua globals has very bad side effects for concurrent requests when these variables are actually supposed to be local only,
-# Lua global variables require Lua table look-up in the global environment (which is just a Lua table), which is computationally expensive, and
+# the misuse of Lua globals has detrimental side effects on concurrent requests when such variables should instead be local in scope,
+# Lua global variables require Lua table look-ups in the global environment which is computationally expensive, and
 # some Lua global variable references may include typing errors which make such difficult to debug.
 
-It is therefore *highly* recommended to always declare such as "local" within an appropriate scope instead.
+It is therefore *highly* recommended to always declare such within an appropriate local scope instead.
+
+<geshi lang="lua">
+    -- Avoid
+    foo = 123
+    -- Recomended
+    local foo = 123
+
+    -- Avoid
+    function foo() return 123 end
+    -- Recomended
+    local function foo() return 123 end
+</geshi>
+
 
 To find all instances of Lua global variables in your Lua code, run the [https://github.com/openresty/nginx-devel-utils/blob/master/lua-releng lua-releng tool] across all <code>.lua</code> source files:
 <geshi lang="text">
@@ -565,7 +578,7 @@ will not work as expected.
 
 == Cosockets Not Available Everywhere ==
 
-Due the internal limitations in the nginx core, the cosocket API is disabled in the following contexts: [[#set_by_lua|set_by_lua*]], [[#log_by_lua|log_by_lua*]], [[#header_filter_by_lua|header_filter_by_lua*]], and [[#body_filter_by_lua|body_filter_by_lua]].
+Due to internal limitations in the nginx core, the cosocket API is disabled in the following contexts: [[#set_by_lua|set_by_lua*]], [[#log_by_lua|log_by_lua*]], [[#header_filter_by_lua|header_filter_by_lua*]], and [[#body_filter_by_lua|body_filter_by_lua]].
 
 The cosockets are currently also disabled in the [[#init_by_lua|init_by_lua*]] and [[#init_worker_by_lua|init_worker_by_lua*]] directive contexts but we may add support for these contexts in the future because there is no limitation in the nginx core (or the limitation might be worked around).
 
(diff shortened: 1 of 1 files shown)


agentzh:
@dakanji Thanks for the hard work on this! Really appreciated :)

@doujiang24 @moonbingbing @membphis @thibaultcha Will you have a look at this?

doc/HttpLuaModule.wiki
```diff
@@ -523,12 +523,12 @@ Here is the reason: by design, the global environment has exactly the same lifet
 Generally, use of Lua global variables is a really really bad idea in the context of ngx_lua because
 
 # misuse of Lua globals has very bad side effects for concurrent requests when these variables are actually supposed to be local only,
-# Lua global variables require Lua table look-up in the global environment (which is just a Lua table), which is kinda expensive, and
-# some Lua global variable references are just typos, which are hard to debug.
+# Lua global variables require Lua table look-up in the global environment (which is just a Lua table), which is computationally expensive, and
+# some Lua global variable references may include typing errors which make such difficult to debug.
 
-It's *highly* recommended to always declare them via "local" in the scope that is reasonable.
+It is therefore *highly* recommended to always declare such as "local" within an appropriate scope instead.
```
thibaultcha: To maintain context here, we could say:
```
It is therefore highly recommended to always declare variables and functions as "local" within an appropriate scope, instead of the default global scope assumed by Lua:

    -- don't
    foo = 123
    -- do
    local foo = 123

    -- don't
    function foo() return 123 end
    -- do
    local function foo() return 123 end
```

This provides a good summary for this paragraph (for non-attentive readers or readers who skipped the details) and provides a concrete, more natural example than the above `require` one. It also shows that this applies to functions (a mistake I have seen too often from contributors).

doc/HttpLuaModule.wiki
```diff
@@ -523,12 +523,12 @@ Here is the reason: by design, the global environment has exactly the same lifet
 Generally, use of Lua global variables is a really really bad idea in the context of ngx_lua because
```
thibaultcha: > a really really bad idea

It seems informal to repeat "really" here. If this PR's scope is to avoid informality, I believe this could be replaced with one of:

```
Generally, use of Lua global variables is [considered harmful] in the context of ngx_lua because:
Generally, use of Lua global variables is [considered a] bad idea in the context of ngx_lua because:
Generally, use of Lua global variables is [considered a very] bad idea in the context of ngx_lua because:
Generally, use of Lua global variables is a [really] bad idea in the context of ngx_lua because:
Generally, use of Lua global variables is a [very] bad idea in the context of ngx_lua because:
Generally, use of Lua global variables is a [really, really] bad idea in the context of ngx_lua because:
```

(note the colon at the end as well)

doc/HttpLuaModule.wiki
```diff
@@ -565,24 +565,23 @@ will not work as expected.
 
 == Cosockets Not Available Everywhere ==
 
-Due the internal limitations in the nginx core, the cosocket API are disabled in the following contexts: [[#set_by_lua|set_by_lua*]], [[#log_by_lua|log_by_lua*]], [[#header_filter_by_lua|header_filter_by_lua*]], and [[#body_filter_by_lua|body_filter_by_lua]].
+Due the internal limitations in the nginx core, the cosocket API is disabled in the following contexts: [[#set_by_lua|set_by_lua*]], [[#log_by_lua|log_by_lua*]], [[#header_filter_by_lua|header_filter_by_lua*]], and [[#body_filter_by_lua|body_filter_by_lua]].
```
thibaultcha: typo: "due **to** internal limitations in **the** nginx core"

Additionally, most occurrences of "Nginx" in this document are capitalizes (`Nginx` vs `nginx`), except in a few places, like this one. Should this be capitalized?

doc/HttpLuaModule.wiki
```diff
@@ -565,24 +565,23 @@ will not work as expected.
...
 
-PCRE sequences such as <code>\d</code>, <code>\s</code>, or <code>\w</code>, require special attention because in string literals, the backslash character, <code>\</code>, is stripped out by both the Lua language parser and by the Nginx config file parser before processing. So the following snippet will not work as expected:
+PCRE sequences such as <code>\d</code>, <code>\s</code>, or <code>\w</code>, require special attention because in string literals, the backslash character, <code>\</code>, is stripped out by both the Lua language parser and by the Nginx config file parser before processing if not within a <code>*_by_lua_block {}</code> directive. So the following snippet will not work as expected:
 
 <geshi lang="nginx">
     # nginx.conf
     ? location /test {
     ?     content_by_lua '
-    ?         local regex = "\d+"  -- THIS IS WRONG!!
+    ?         local regex = "\d+"  -- THIS IS WRONG OUTSIDE A *_by_lua_block DIRECTIVE!!
```
thibaultcha: typo: I believe this is missing an "of":
```
OUTSIDE OF A *_by_lua_block DIRECTIVE
```

Could also drop the "!!" to avoid informality.

doc/HttpLuaModule.wiki
```diff
@@ -658,6 +657,21 @@ Within external script files, PCRE sequences presented as long-bracketed Lua str
     -- evaluates to "1234"
 </geshi>
 
+As noted earlier, PCRE sequences presented within <code>*_by_lua_block {}</code> directives, available following the <code>v0.9.17</code> release, do not require modification.
```
thibaultcha: nitpick++: I'd use dashes (`- thing -`), or maybe even parenthesis instead of comma here, to not lose context. Very much a nitpick.

doc/HttpLuaModule.wiki
```diff
@@ -5447,7 +5454,7 @@ Fetch a list of the keys from the dictionary, up to <code><max_count></code>.
 
 By default, only the first 1024 keys (if any) are returned. When the <code><max_count></code> argument is given the value <code>0</code>, then all the keys will be returned even there is more than 1024 keys in the dictionary.
 
-'''WARNING''' Be careful when calling this method on dictionaries with a really huge number of keys. This method may lock the dictionary for quite a while and block all the nginx worker processes that are trying to access the dictionary.
+'''CAUTION''' Avoid calling this method on dictionaries with a very large number of keys as this method may lock the dictionary for a significant amount of time while blocking nginx worker processes that are trying to access the dictionary.
```
thibaultcha: very long sentence and yet important. Could use some punctuation maybe?
```
for a significant amount of time[, all the] while blocking Nginx worker processes
```

note: capitalized "Nginx" (if desired).

dakanji:
Hmmm, was taking a light touch and was not seeking to overhaul things as there are other items that could potentially be looked at but if tackled all at once, would make the PR too difficult to manage.

Done the most of the suggested changes though and other stuff could be looked at later with the conversion script run at some point afterwards. For instance, there is mixed usage of "NGINX", "Nginx" and "nginx" but trying to standardise this will touch too many lines.

I think it is best to deal with piecemeal. If/when this is merged, something else can be targeted.

thibaultcha:
@dakanji Agreed. Informality being the scope of this PR, there was one or two missing updates I believe. The Nginx capitalization thing... Just raising the concern, nothing has to be done about it urgently I believe :)

agentzh:
@dakanji Merged with minor edits in the commit log message :) Thanks!

@thibaultcha Thanks for your review!

PR #985, opened 2017-02-13
feature: Add FFI functions for ngx.req.get_post_args

p0pr0ck5:


src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,122 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+    }
+
+    if (count) {
+        if (max > 0 && count > max) {
+            count = max;
+            ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                           "lua hit post args limit %d", max);
+        }
+
+        return count;
```
thibaultcha: Should we exit early and move this condition + return statement inside the previous loop's else, to exit the loop asap, or is it better off here?

thibaultcha: Also, I don't think we need the `count = max` assignment in such a case. We can just `return max;`, probably.

src/ngx_http_lua_args.c
```diff
@@ -433,8 +438,8 @@ ngx_http_lua_ffi_req_get_uri_args_count(ngx_http_request_t *r, int max)
 
 
 int
-ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
-    ngx_http_lua_ffi_table_elt_t *out, int count)
+ngx_http_lua_ffi_req_args_helper(u_char *buf, ngx_http_lua_ffi_table_elt_t *out,
+    int count, int len)
```
p0pr0ck5: Ah, good point :) Will fix up!

src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,122 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
-#endif /* NGX_LUA_NO_FFI_API */
 
 
+int
+ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
+    ngx_http_lua_ffi_table_elt_t *out, int count)
+{
+    int len;
+
+    len = r->args.len;
```
p0pr0ck5: Ah, yep :) thanks!

src/ngx_http_lua_args.c
p0pr0ck5: This logic was copied from `ngx_http_lua_ffi_req_get_uri_args_count`, so to maintain consistency, we should stick with one or the other :) I'm good either way, though I'm a fan of the existing approach as it requires only one extra conditional check instead of `n` checks for `n` count args. Maybe worth some micro-benchmarking to figure out what's more performant? And would different compilers (or different compiler options) treat this the same way, or hoist some of the conditional checks into the loop anyhow? My lack of compiler knowledge is showing ;)

src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,122 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+        }
+
+        return count;
+    }
+
+    if (len) {
+        return 1;
+    }
+
+    return 0;
```
p0pr0ck5: Again, I think the more important goal is maintaining existing style. If the approach you suggest is more optimal, could we refactor both the existing function upon which the new function is based, and the new function, in a separate commit? What are your thoughts here?

src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,122 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+}
+
```
p0pr0ck5: Derp. Thanks :)

src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,122 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+int
+ngx_http_lua_ffi_req_get_post_args(ngx_http_request_t *r, u_char *buf,
+    ngx_http_lua_ffi_table_elt_t *out, int count, int len)
```
p0pr0ck5: Yep, good idea. Will tweak and update the lua-resty-core PR as well. Thanks!

src/ngx_http_lua_args.c
thibaultcha: Hm, I wouldn't say I have much more knowledge, but I do know a branch misprediction can be hurtful, so maybe it's better off this way :D

p0pr0ck5:
Tests using the lua-resty-core implementation will fail due to the typo fixed up in https://github.com/openresty/lua-nginx-module/pull/990 (travis CI should pass here though); I'll rebase once that's merged.

p0pr0ck5:
@agentzh what are you thoughts on this PR? It would be nice to have this (lua-resty-core sister PR at https://github.com/openresty/lua-resty-core/pull/87) in :)

p0pr0ck5:
@agentzh ping. any thoughts here?

src/ngx_http_lua_args.c
```diff
@@ -544,7 +547,119 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+
+    if (r->request_body->temp_file) {
+        return NGX_HTTP_LUA_FFI_POST_ARGS_IN_FILE;
+    }
+
+    if (r->request_body->bufs == NULL) {
+        return NGX_HTTP_LUA_FFI_POST_ARGS_EMPTY;
+    }
+
+    return NGX_HTTP_LUA_FFI_POST_ARGS_OK;
```
agentzh: We usually do not use a separate C function to do checkups. See other existing FFI C functions for error handling.

src/ngx_http_lua_args.c
```diff
@@ -14,6 +14,11 @@
 #include "ngx_http_lua_args.h"
 #include "ngx_http_lua_util.h"
 
+#define NGX_HTTP_LUA_FFI_POST_ARGS_OK 0
+#define NGX_HTTP_LUA_FFI_POST_ARGS_EMPTY 1
+#define NGX_HTTP_LUA_FFI_POST_ARGS_UNREAD 2
+#define NGX_HTTP_LUA_FFI_POST_ARGS_IN_FILE 3
```
agentzh: I'd rather not introduce new FFI constants unless absolutely necessary. If every FFI C function introduce its own error codes, it would be a huge amount of such constants in the top-level namespace.

p0pr0ck5:
Thanks, updated to remove the post args constants, and moved the extra verification.

p0pr0ck5:
Will clean up the git history here when ready for merge, or we need more changes.

agentzh:
@p0pr0ck5 Don't worry about the git history too much. I'll always squash before merging manually. Basically it is recommended to use incremental commits for PR submitters in their own branch. Generally speaking, squashing by the PR authors themselves will make my life harder instead of easier. The only requirement on PR authors from my side is to ensure the branch can be rebased to the latest master *cleanly*, but it does not require for the PR authors to actually rebase to the latest master periodically themselves (unless conflicts occur, of course). Sorry for the confusions!

p0pr0ck5:
@agentzh not to hassle, but any update?

agentzh:
@p0pr0ck5 Sorry for the delay on my side. This is still in my review queue. Will have another close look as soon as I can manage. Thanks!

p0pr0ck5:
@agentzh ping, any thoughts here?

agentzh:
@p0pr0ck5 Sorry, too busy with other higher priority stuff. Will look into this when I have a chance. Thanks for your understanding and sorry for the delay on my side.

src/ngx_http_lua_args.c
```diff
@@ -544,7 +542,109 @@ ngx_http_lua_ffi_req_get_uri_args(ngx_http_request_t *r, u_char *buf,
...
+    if (r->discard_body || r->request_body->bufs == NULL) {
+        return -1;
+    }
+
+    if (r->request_body == NULL) {
+        return -2;
+    }
+
+    if (r->request_body->temp_file) {
+        return -3;
```
agentzh: We should avoid magical numbers like -1, -2, and -3.

p0pr0ck5: @agentzh you asked me to avoid constants in this comment: https://github.com/openresty/lua-nginx-module/pull/985#discussion_r117308957

So what is the appropriate design in your eyes?

p0pr0ck5:
@agentzh any thoughts on this? I've gotten some conflicting feedback here, would love to be able to clean it up and see this functionality merged :)

PR #989, opened 2017-02-16
doc: added comment in to installation block

SmirnovW:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -298,6 +298,9 @@ Build the source with this module:
          --add-module=/path/to/ngx_devel_kit \
          --add-module=/path/to/lua-nginx-module
 
+# Note that you also should add arguments which are used in your current nginx  version.
+# You can get these arguments using command 'nginx -V'
+
  make -j2
  make install
 ```


README.markdown
```diff
@@ -298,6 +298,9 @@ Build the source with this module:
          --add-module=/path/to/ngx_devel_kit \
          --add-module=/path/to/lua-nginx-module
 
+# Note that you also should add arguments which are used in your current nginx  version.
```
agentzh: "also should" => "should also"
2 spaces between `nginx` and `version`, which should be one.
"arguments" => "options"

README.markdown
```diff
@@ -298,6 +298,9 @@ Build the source with this module:
...
+# You can get these arguments using command 'nginx -V'
```
agentzh: You should append ` (without quotes)` to the end of this line.

Also, `arguments` should be `options`.

agentzh:
Oh, BTW, `README.markdown` is automatically generated. Do not edit it manually! See the comment for details at the beginning of that file. You should edit the `doc/HttpLuaModule.wiki` file instead.

PR #997, opened 2017-02-21
feature: support ssl.create_ctx and tcp:setsslctx

detailyang:

This PR have been discussed at #957.  These FFI API as the following:
 
* `ngx_http_lua_ffi_ssl_ctx_init(const u_char *method, size_t method_len, char **err)`
* `ngx_http_lua_ffi_ssl_ctx_set_cert(void *cdata_ctx, void *cdata_cert, char **err)`
* `ngx_http_lua_ffi_ssl_ctx_set_priv_key(void *cdata_ctx, void *cdata_key, char **err)`
* `ngx_http_lua_ffi_ssl_ctx_free(void *cdata)`

I refer to the [Node.js C++](https://github.com/nodejs/node/blob/master/src/node_crypto.cc#278) as a guide. Now It exposes `set_cert` and `set_priv_key` to FFI and It's convenient that to feed more argument to `SSL_CTX` object in future if we want.

need sure we all agree on this API Design:)

These is a [sister pr]() about `lua-resty-core`. The API usage is as the following:

```lua
local ssl = require "ngx.ssl"
local ctx, err = ssl.create_ctx{
    cert = "pem-data", #parsed by ssl.parse_pem_cert
    priv_key = "pem-data", #parsed by ssl.parse_pem_priv_key
}

if ctx ~= nil then
    return error("create ssl ctx error" .. err)
end

-- we can cache the ctx object in a cache like lua-resty-lrucache

local sock = ngx.socket.tcp()
local ok, err = sock:setsslctx(ctx) #the setsslctx is a Lua function atop FFI whcich been implement on lua-resty-core
sock:sslhandshake(...)
```

@agentzh @thibaultcha @doujiang24  
have a look at this PR? Many thanks :D


ghedo:
It seems there is quite a lot of code duplication from NGINX which is unnecessary IMO.

Why not make `ngx_http_lua_ffi_ssl_ctx_init()` allocate and return the `ngx_ssl_t` instead of the `SSL_CTX`? This way you can simply call `ngx_ssl_create()` (with `llcf->ssl_protocols`) in there and avoid all the duplication.

Then you can use that `ngx_ssl_t` in `ngx_http_lua_ffi_socket_tcp_setsslctx()` instead of allocating a new one. This would also make the `method` parameter superfluous, which is a good thing IMO.

Also, the currently implemented tests are probably not enough to test the functionality.

detailyang:
@ghedo  Thanks for you review :D

>It seems there is quite a lot of code duplication from NGINX which is unnecessary IMO.

These duplicated code will appear because I cannot reuse `ngx_ssl_create ` function. I have consider to alloc ngx_ssl_t in FFI Function on `ngx_cycle->pool`, I'm not sure it is the best practice  to alloc memory in FFI Function on `ngx_cycle->pool`.  So I choose the `SSL_CTX`.

>This would also make the method parameter superfluous, which is a good thing IMO.

Since we have provide API to feed the `SSL_CTX` object, so I choose to provide the superfluous `method` to more purely control SSL negotiation protocols not default `llcf->ssl_protocols`.

>Also, the currently implemented tests are probably not enough to test the functionality.

Just simply test in `lua-nginx-module` and most tests is in `lua-resty-core`.

@agentzh I think @ghedo' suggestion  is almost good for me. How do you think of it ? By the way, I will continue to refactor until we agree on the API Design :D

ghedo:
> I have consider to alloc ngx_ssl_t in FFI Function on ngx_cycle->pool, I'm not sure it is the best practice to alloc memory in FFI Function on ngx_cycle->pool.

`ngx_http_lua_ssl_ctx_create_method()` can probably take a `ngx_http_request_t` as argument, like the other FFI functions do, that way you can use `r->pool`.

> Since we have provide API to feed the SSL_CTX object, so I choose to provide the superfluous method to more purely control SSL negotiation protocols not default llcf->ssl_protocols.

IMO it would still be better to have a string parameter in  `ngx_http_lua_ssl_ctx_create_method()` that uses the same format as `lua_ssl_protocols`. You wouldn't have to use `llcf->ssl_protocols` itself, just the string format. This way you could stilll use `ngx_ssl_create()`, you wouldn't expose the underlying OpenSSL API and it would be more consistent.

But really, configuring enabled protocols doesn't need to happen at `SSL_CTX` creation time. A somewhat better API IMO would separate the ctx creation (and use `llcf->ssl_protocols` by default, so the user doesn't need to think about this if they don't want to) but provide an additional function to change the enabled protocols if the user wants to do that. But again, IMO :)

> Just simply test in lua-nginx-module and most tests is in lua-resty-core.

Well, yeah, but you also need to test the functionality you are adding. Something like creating a ctx, setting a user certificate on it, and using that ctx with a socket to connect to a server.

detailyang:
@ghedo 
>ngx_http_lua_ssl_ctx_create_method() can probably take a ngx_http_request_t as argument, like the other FFI functions do, that way you can use r->pool.

Since we want to cache `ssl_ctx` in lrucache, we should make sure the lifecycle of `ssl_ctx` do not bind to the `r->pool`. I'm ready to refactor the code you are suggesting as the following:

```c
void * ngx_http_lua_ffi_ssl_ctx_init(ngx_uint_t protocols, char **err);
```
we will alloc `ngx_ssl_t` on `ngx_cycle->pool`, because we want to cache `ssl_ctx` in lrucache.

>IMO it would still be better to have a string parameter in ngx_http_lua_ssl_ctx_create_method() that uses the same format as lua_ssl_protocols. You wouldn't have to use llcf->ssl_protocols itself, just the string format. This way you could stilll use ngx_ssl_create(), you wouldn't expose the underlying OpenSSL API and it would be more consistent.
But really, configuring enabled protocols doesn't need to happen at SSL_CTX creation time. A somewhat better API IMO would separate the ctx creation (and use llcf->ssl_protocols by default, so the user doesn't need to think about this if they don't want to) but provide an additional function to change the enabled protocols if the user wants to do that. But again, IMO :)

Now that we are ready to reuse `ngx_ssl_create`, so let's move the argument `method`, just use `protocols` to as the protocols argument. Also we can compare string in `lua-resty-core` then to pass real arguments to `void * ngx_http_lua_ffi_ssl_ctx_init(ngx_uint_t protocols, char **err)`


>Well, yeah, but you also need to test the functionality you are adding. Something like creating a ctx, setting a user certificate on it, and using that ctx with a socket to connect to a server.

That's good and it's worth to do to make sure  the functionality is okay without `lua-resty-code` :D

detailyang:
@ghedo After refactor the codebase, It reuse the `ngx_ssl_create` function. But about the `ngx_ssl_t`, it's alloced on the stack rather than heap. Since there is not a good way to manage the memory (malloc?) in the FFI function. So just alloc on the stack which is a not bad way IMO. In `tcp:setsslctx`, we alloc the `ngx_ssl_t` in `r->pool` to make sure the lifecycle of `ngx_ssl_t` is the same as `request`.

@agentzh How do you think of it?

agentzh:
@detailyang We should not allocate SSL data structures in `ngx_cycle->pool` since for small memory blocks in the nginx memory pool, they can never be freed individually (they will only be freed when the whole pool is destroyed).

Also, we should not let the SSL CTX thing bound to `r->pool` either.

agentzh:
@detailyang Your current implementation seems good. I'll need to have a closer look.

@doujiang24 What do you think of it?

@ghedo Thanks for your review! Appreciated!

t/151-ssl-ctx.t
```diff
@@ -0,0 +1,384 @@
...
+our $serverKey = read_file("t/cert/ca-client-server/server.key");
+our $serverUnsecureKey = read_file("t/cert/ca-client-server/server.unsecure.key");
+our $serverCrt = read_file("t/cert/ca-client-server/server.crt");
+our $caKey = read_file("t/cert/ca-client-server/ca.key");
+our $caCrt = read_file("t/cert/ca-client-server/ca.crt");
+our $http_config = <<_EOS_;
+lua_package_path "\$prefix/html/?.lua;$TEST_NGINX_LUA_PACKAGE_PATH/?.lua;;../lua-resty-lrucache/lib/?.lua;";
+
+init_by_lua_block {
+    local ffi = require "ffi"
```
agentzh: The lua-nginx-module's test suite already depends on lua-resty-core. I think we'd better avoid duplicating such lua-resty-core code here in the test suite, which adds maintenance overhead. We can just test against the lua-resty-core Lua API directly.

src/ngx_http_lua_ssl.c
```diff
@@ -13,6 +13,11 @@
 #if (NGX_HTTP_SSL)
 
 
+#define ngx_http_lua_ssl_check_method(method, method_len, s)             \
```
agentzh: Better use Lua inline functions instead of macros for this.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +39,135 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+#ifdef LIBRESSL_VERSION_NUMBER
+
+    *err = "LibreSSL not supported";
+    return NGX_ERROR;
+
+#else
+
+#   if OPENSSL_VERSION_NUMBER < 0x1000205fL
+
+    *err = "at least OpenSSL 1.0.2e required but found " OPENSSL_VERSION_TEXT;
```
agentzh: Is this really requiring OpenSSL 1.0.2e+?

src/ngx_http_lua_ssl.c
detailyang: This macro is remaining after refactoring and should be removed.

src/ngx_http_lua_ssl.c
detailyang: The API [SSL_CTX_add1_chain_cert](https://www.openssl.org/docs/man1.1.0/ssl/SSL_CTX_add1_chain_cert.html) says 
> These functions were first added to OpenSSL 1.0.2.

So I choose to requiring 1.0.2e in order to keep consistent with other function requires

t/151-ssl-ctx.t
detailyang: Yup, these tests actually adds maintenance overhead for me, Since these tests is the same as lua-resty-core. According what you have said, Can I think that if we are adding the FFI C API, we should simply test it in `lua-nginx-module`  and specially test in `lua-resty-core` ?

src/ngx_http_lua_ssl.c
agentzh: @detailyang Okay. Good to know :)

detailyang:
@agentzh @doujiang24  @ghedo I did some refactoring again, can you guys have a look at it? 
Thanks :D

src/ngx_http_lua_socket_tcp.c
```diff
@@ -316,6 +322,19 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
...
     lua_pushvalue(L, -1);
     lua_setfield(L, -2, "__index");
+
+#if (NGX_HTTP_SSL)
+
+#ifndef NGX_LUA_NO_FFI_API
+
+    /* expose tcp object metatable to global for FFI */
+    lua_pushvalue(L, -1);
+    lua_setglobal(L, ngx_http_lua_ngx_socket_tcp_mt_key);
```
doujiang24: maybe we'd better set it into REGISTRY table? And we can get it by `debug.getregistry`.
like: https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/ctx.lua#L12

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +34,132 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+void
+ngx_http_lua_ffi_ssl_ctx_free(void *cdata)
+{
+    SSL_CTX *ssl_ctx = cdata;
+
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log,
+                   0,
+                   "lua ssl ctx free: %p:%d",
+                   ssl_ctx,
+                   ssl_ctx->references);
```
doujiang24: style: maybe bellow can be better?
```
0, "lua ssl ctx free: %p:%d",
ssl_ctx, ssl_ctx->references);
```
I mean we don't need one argument one line.

src/ngx_http_lua_socket_tcp.c
detailyang: Agree. Put it into REGISTRY table is better than global table.

src/ngx_http_lua_ssl.c
detailyang: gotacha :D

detailyang: Maybe it's better as the following ?

```c
    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log,
                   0, "lua ssl ctx free: %p:%d", ssl_ctx, ssl_ctx->references);
```

It will do not exceed 80 columns.

doujiang24: Yeah, one line is better :)

detailyang:
Hello :D
@agentzh  @doujiang24  @ghedo  How about it now about the API design and implement?

t/151-ssl-ctx.t
```diff
@@ -0,0 +1,73 @@
...
+    void *ngx_http_lua_ffi_ssl_ctx_init(unsigned int protocols, char **err);
+
+    void ngx_http_lua_ffi_ssl_ctx_free(void *cdata);
+
+    int ngx_http_lua_ffi_ssl_ctx_set_priv_key(void *cdata_ctx,
+        void *cdata_key, char **err);
+
+    int ngx_http_lua_ffi_ssl_ctx_set_cert(void *cdata_ctx,
+        void *cdata_cert, char **err);
+]]
```
agentzh: Let's simply test the lua-resty-core's Lua API wrappers here in lua-nginx-module's test suite. Such things add to maintenance overhead.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +34,129 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+    if (sk_X509_num(cert) < 1) {
+        *err = "sk_X509_num() failed";
+        ngx_ssl_error(NGX_LOG_ERR, ngx_cycle->log, 0, *err);
+        return NGX_ERROR;
+    }
+
+    x509 = sk_X509_value(cert, 0);
+    if (x509 == NULL) {
+        *err = "sk_X509_value() failed";
+        ngx_ssl_error(NGX_LOG_ERR, ngx_cycle->log, 0, *err);
```
agentzh: I think instead of logging the real error message into nginx's error log file, we should propagate it to the Lua land via the `err` output parameter (of course, the type of `err` needs adjustment). Let's avoid direct logging in these FFI C functions. We should let the Lua land caller to handle these error and inspect these error messages, which is much more flexible.

detailyang: Agreed. propagating the OpenSSL error message to Lua land is more flexible. So I'm thinking expose the error code by ` unsigned long ERR_get_error(void);`, let the Lua land caller to get the error string with error code by these functions as the following:

```c
 #include <openssl/err.h>

 char *ERR_error_string(unsigned long e, char *buf);
 void ERR_error_string_n(unsigned long e, char *buf, size_t len);

 const char *ERR_lib_error_string(unsigned long e);
 const char *ERR_func_error_string(unsigned long e);
 const char *ERR_reason_error_string(unsigned long e);
```

Also we will implement the FFI API upon these functions. @agentzh How about this?

t/151-ssl-ctx.t
detailyang: Well. The only thing I'm confused  is what's the 'simply test' ?  if we want to simply test the `ngx_http_lua_ffi_ssl_ctx_set_priv_key` and `ngx_http_lua_ffi_ssl_ctx_set_cert` , should we set the priv_key and cert then send it to server ? Or just set the priv_key and cert then testing the return value.

doujiang24: @detailyang Yeah, you got it :)

src/ngx_http_lua_ssl.c
doujiang24: @agentzh I think agentzh just means remove the `ngx_ss_error` here :)
Seems we don't need `ERR_get_error` here?

detailyang: @doujiang24  uhh, It looks like I'm worrying. But how to actively get the specific error message in Lua land with out call `ERR_get_error` ?

doujiang24: @detailyang Okay, I see, yes we need `ERR_get_error`.
But I think we should do it in C land, something like this:
```
int ffi_c_func(char *err, int *err_len) {
    /* get message by ERR_get_error or something else */
    err_len = sprintf(err, message) - err;
}
```

ElvinEfendi:
Is there any plan to implement setting `ssl_trusted_certificate` in the same manner? When you use `lua_ssl_trusted_certificate` directive, the file is being loaded for every `location` context(I'm writing up a more detailed issue to explain how this can be problematic).

Thanks for working on this!

src/ngx_http_lua_ssl.c
detailyang: @doujiang24 Maybe the `err_len` is no need. I check the API `char *ERR_error_string(unsigned long e, char *buf);` It says:
>ERR_error_string() generates a human-readable string representing the error code e, and places it at buf. buf must be at least 120 bytes long. If buf is NULL , the error string is placed in a static buffer. 

So pointing the static buffer is not bad IMO as the following:

```c
   if (SSL_CTX_use_PrivateKey(ssl_ctx, key) == 0) {
        e = ERR_get_error();
        if (e == 0) {
            *err = "SSL_CTX_use_PrivateKey() failed";

        } else {
            *err = ERR_error_string(e, NULL);
        }

        return NGX_ERROR;
    }
```

@doujiang24  @agentzh How about this?

detailyang:
@ElvinEfendi Yup, the `ssl_trusted_certificate ` is in our plan. But now we need sure the API Design is okay so that we do not break the API Design if we want to add new API in future :D

agentzh:
@detailyang I think it's cleaner to let the caller allocate the error message buffer. See existing FFI C API used by lua-resty-core for such examples (like `resty.core.shdict`).

detailyang:
@agentzh @doujiang24 

I peek the OpenSSL implement of [ERR_error_string](https://github.com/openssl/openssl/blob/master/crypto/err/err.c#L551), when buffer is NULL , it use static buffer as the return value. But In multi thread, it's wrong. So caller should allocate the error message buffer :D .

Now the new implement is as the following (POC):

```c
int
ngx_http_lua_ffi_ssl_ctx_set_priv_key(void *cdata_ctx, void *cdata_key,
    char **err_buf, size_t err_buf_len)
{
    SSL_CTX     *ssl_ctx = cdata_ctx;
    EVP_PKEY    *key = cdata_key;

    u_long       e;

    if (SSL_CTX_use_PrivateKey(ssl_ctx, key) == 0) {
        e = ERR_get_error();
        if (e == 0) {
            *err_buf = "SSL_CTX_use_PrivateKey() failed";

        } else {
            ERR_error_string_n(e, *err_buf, err_buf_len);
        }

        return NGX_ERROR;
    }

    return NGX_OK;
}
```

These are two questions to point out.

1. the type of err_buf_len is `size_t` rather than `size_t *`
When Setting the error message by `ERR_error_string_n`, cannot get the error message length. Since the `ERR_error_string_n` writes at most length characters (including the terminating 0) and truncates the string if necessary. So it guarantee the error message is '\0' terminating.

2. the `ERR_error_string_n` was added in OpenSSL 0.9.6 

Should we ask for the OpenSSL version at least 0.9.6 ? Or use `ERR_error_string` to replace when OpenSSL version is less than 0.9.6. It will need more works to implement.

agentzh:
@detailyang I think we can just drop support for openssl versions older than 0.9.6 :)

detailyang:
@agentzh  How about this API now as the following ? 

```c
int
ngx_http_lua_ffi_ssl_ctx_set_priv_key(void *cdata_ctx, void *cdata_key,
    char **err_buf, size_t err_buf_len)
{
    SSL_CTX     *ssl_ctx = cdata_ctx;
    EVP_PKEY    *key = cdata_key;

    u_long       e;

    if (SSL_CTX_use_PrivateKey(ssl_ctx, key) == 0) {
        e = ERR_get_error();
        if (e == 0) {
            *err_buf = "SSL_CTX_use_PrivateKey() failed";

        } else {
            ERR_error_string_n(e, *err_buf, err_buf_len);
        }

        return NGX_ERROR;
    }

    return NGX_OK;
}
```

It's a little tricky that the caller should pass the error buffer address、 the length of error buffer and the error buffer address will be used to point the literal string or be placed the error message from OpenSSL `ERR_error_string_n`

BTW, the tests failed because I use `lua-resty-core` rather than FFI to simplely test.

agentzh:
@detailyang No, that's wrong to overwrite the buffer pointer since the caller may not check the pointer value at all, like this:

```C
const char buf[64];
if (foo(&buf, 64) != 0) {
    printf("error: %64s\n", buf);
}
```

So we should use `ngx_copy()` to copy the literal string into the user-supplied error string buffer.

BTW, you should edit `.travis.yml` to make the lua-resty-core repo point to your own in this PR. We can edit it ourselves later when we merge your work.

detailyang:
@agentzh gotcha :D

After rethinking about a lot of API design (refer to the other lua-resty-core FFI implementations) and I implemented different version on my localhost, now the API design is as the following:

```c

static size_t
ngx_http_lua_ssl_get_error(u_long e, u_char *ssl_err_buf,
    size_t ssl_err_buf_len, u_char *default_errmsg,
    size_t default_errmsg_len)
{
    size_t len;

    if (e == 0) {
        len = ngx_min(ssl_err_buf_len, default_errmsg_len);
        ssl_err_buf = ngx_copy(ssl_err_buf, default_errmsg, len);
        return len;
    }

    ERR_error_string_n(e, (char *) ssl_err_buf, ssl_err_buf_len);

    return ngx_strlen(ssl_err_buf);
}


int
ngx_http_lua_ffi_ssl_ctx_set_priv_key(void *cdata_ctx, void *cdata_key,
    u_char *err_buf, size_t *err_buf_len)
{
    SSL_CTX     *ssl_ctx = cdata_ctx;
    EVP_PKEY    *key = cdata_key;

    char        *err;

    if (SSL_CTX_use_PrivateKey(ssl_ctx, key) == 0) {
        err = "SSL_CTX_use_PrivateKey() failed";
        goto failed;
    }

    return NGX_OK;

failed:

    *err_buf_len = ngx_http_lua_ssl_get_error(ERR_get_error(), err_buf,
                                                  *err_buf_len, (u_char *)err,
                                                  ngx_strlen(err));
    return NGX_ERROR;
}
```

You can review the latest changes on above commits when you are free :D

detailyang:
@agentzh  Is it okay ? And what something can I change for this PR

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,160 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+    }
+
+    return NGX_OK;
+#   endif  /* OPENSSL_VERSION_NUMBER < 0x1000205fL */
+#endif
+
+failed:
+
+    *err_buf_len = ngx_http_lua_ssl_get_error(ERR_get_error(), err_buf,
+                                                  *err_buf_len, (u_char *) err,
```
rainingmaster: or uniform format, should we align in front of the brackets? also in follow `ngx_http_lua_ssl_get_error`

detailyang: Good catch

detailyang:
At first I'm super sorry It's too long not to update and I'm stuck in personal work recently.Now It's time to continue completing this feature. 

@agentzh Can you help me  to continue review this PR ?

src/ngx_http_lua_socket_tcp.h
```diff
@@ -92,6 +92,7 @@ struct ngx_http_lua_socket_tcp_upstream_s {
 
 #if (NGX_HTTP_SSL)
     ngx_str_t                        ssl_name;
+    ngx_ssl_t                        *ssl;
```
agentzh: Style: alignment issue. The `ssl` identifier should be aligned, excluding the `*` prefix.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,160 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+static size_t
+ngx_http_lua_ssl_get_error(u_long e, u_char *ssl_err_buf,
+    size_t ssl_err_buf_len, u_char *default_errmsg,
+    size_t default_errmsg_len)
+{
+    size_t len;
+
+    if (e == 0) {
+        len = ngx_min(ssl_err_buf_len, default_errmsg_len);
+        ssl_err_buf = ngx_copy(ssl_err_buf, default_errmsg, len);
```
agentzh: The return value is useless. Better use `ngx_memcpy` here instead of `ngx_copy`.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,160 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+               OPENSSL_VERSION_TEXT;
+    goto failed;
+
+#   else
+
+    X509              *x509 = NULL;
+    SSL_CTX           *ssl_ctx = cdata_ctx;
+    STACK_OF(X509)    *cert = cdata_cert;
+
+#ifdef OPENSSL_IS_BORINGSSL
```
agentzh: Here it should use 2 levels of indentation after `#`.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,160 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+    }
+
+    if (SSL_CTX_use_certificate(ssl_ctx, x509) == 0) {
+        err = "SSL_CTX_use_certificate() failed";
+        goto failed;
+    }
+
+    /* read rest of the chain */
+
+    for (i = 1; i < sk_X509_num(cert); i++) {
```
agentzh: Why call `sk_X059_num` for more than once?

agentzh: I think we can simply save the result of the first `sk_X509_num` above to a local C variable.

detailyang:
Fix the issues as  the following and merge the newest openresty/master branch:

*  alignment issue
* useless return value, use `ngx_memcpy`
*  2 levels of indentation after #
* use local variable to eliminate more than function call
* remove unused module in tests

@agentzh  Thanks for reviewing :-)

detailyang:
@agentzh I did some code refactor again and not sure it's good enough. Can you review this PR again if you are free. Thanks

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,169 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+#ifndef NGX_LUA_NO_FFI_API
+
+
+void *
+ngx_http_lua_ffi_ssl_ctx_init(ngx_uint_t protocols, char **err)
+{
+    ngx_ssl_t           ssl;
+
+    ssl.log = ngx_cycle->log;
+    if (ngx_ssl_create(&ssl, protocols, NULL) != NGX_OK) {
```
agentzh: I think if we should inherit existing configurations from `ngx_http_lua_loc_conf_t`'s `ssl` field? Please see `ngx_http_lua_set_ssl` for more details. And here in the `create_ctx` API, we can simply allow the user to override the default settings. This would be better.

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,169 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
...
+        *err = "failed to create ssl ctx";
```
agentzh: Also, I think it would be better to return OpenSSL's own error messages if there's any. Otherwise fixed error messages made up in the caller does not give much helpful info.

agentzh: Well, this is not a priority though.

agentzh:
@dndx Will you please have a look at this PR as well as openresty/lua-resty-core#89 ? We also need client certificate support ourselves :) Thanks!

src/ngx_http_lua_ssl.c
detailyang: Well. inherit existing configurations from `ngx_http_lua_loc_conf_t`s ssl field will required this API need `ngx_http_request_t`. we should allow it run any phase. Especially, it can be allowed to run in `init_by_lua_*` to easily cache it. May we expose these FFI  API (like ciphers, crl, ca and so on) to Lua Land . It will be more  flexible and effective to feed arguments to SSL_CTX object.

src/ngx_http_lua_ssl.c
detailyang: Here we reuse the `ngx_ssl_create` API. However we cannot get more OpenSSL's error info from this API.  Surely, we can rewrite this API to get more detail if we want :)

detailyang:
@agentzh I'm ready to provide additional API as the following:

```c
 ngx_http_lua_ffi_ssl_ctx_add_ca_cert；

 ngx_http_lua_ffi_ssl_ctx_set_ciphers；
 
 int ngx_http_lua_ffi_ssl_ctx_set_crl；
 
ngx_http_lua_ffi_ssl_ctx_set_cert_store；
 
 ngx_http_lua_ffi_ssl_x509_store_init；
 
 ngx_http_lua_ffi_ssl_x509_store_free；
 
ngx_http_lua_ffi_ssl_x509_store_add_cert；
```

Based these API, it can provide the ability like `lua_ssl_ciphers` and `lua_ssl_crl` and  `lua_ssl_trusted_certificate`;  The `x509 store` API will be more useful to reduce loading trusted certificate frequently. Caller can load the system trusted certificate like the following:

```lua
local system_cert =  read_file("/etc/pki/tls/cert.pem")
local local_cert =  read_file("xx.pem")
local cert_store, err = ssl.create_x509_store(system_cert, local_cert)
if cert_store == nil then
    return error(err)
end

local ssl_ctx, err = ssl.create_ctx{
    priv_key = priv_key,
    cert = cert,
    cert_store = cert_store
}
```

src/ngx_http_lua_ssl.c
agentzh: @detailyang We'll add cosocket support to `init_by_lua*` soon. By then it will also have a (fake) request object, just like the timer handler. So I don't think this is a showstopper. Any other things I'm overlooking?

agentzh:
@detailyang That's awesome! We'll look into this ASAP.

kostyay-lms:
Any updates on this? I also need this functionality :)

detailyang:
@kostyay-lms Sorry, i'm stuck in my side. However luckily, @agentzh  is helping us  review this PR, it well be okay as  soon as possible :)

daurnimator:
Instead of adding the large API for managing the `SSL_CTX*` explicitly; could we just add `:setsslctx()` and rely on external libraries to implement the ssl context creation? There are already a few libraries in the lua ecosystem that support `SSL_CTX*` management (e.g. luasec, luaossl) that I feel like it would be a waste to reinvent.

agentzh:
@detailyang Please rebase to the latest master. Thank you! It seems like there are some merge conflicts.

agentzh:
@daurnimator I can see your point. But I think this relatively low level API is better to be self-contained and minimally implemented. interfacing with other Lua libraries create undesired dependencies and could also be slow.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1200,6 +1226,35 @@ ngx_http_lua_socket_conn_error_retval_handler(ngx_http_request_t *r,
...
+
+    ngx_ssl_t *ssl;
+
+    ssl = ngx_pcalloc(r->pool, sizeof(ngx_ssl_t));
+    if (ssl == NULL) {
+        *err = "no memory";
+        return NGX_ERROR;
+    }
+
+    ssl->ctx = ssl_ctx;
```
daurnimator: Should increment ref count of `SSL_CTX*`?

daurnimator:
> @daurnimator I can see your point. But I think this relatively low level API is better to be self-contained and minimally implemented.

I think those two goals are at odds: I was suggesting a minimal implementation (of just `setsslctx`) without being self-contained (i.e. leave the `SSL_CTX*` creation to something else; keep it all in e.g. a lua-resty-sslctx project)

daurnimator:
Also consider an API of `:setssl` that takes an openssl `SSL*` instead of an `SSL_CTX*`.
You can derive an `SSL*` from an `SSL_CTX*` but not the inverse; which makes `SSL*` more general.

agentzh:
@daurnimator By "minimal" I mean zero external dependencies for this thing. So as a whole, the software is minimized (including any external deps). Sorry for the confusion.

xiwenc:
@detailyang Would love to have this PR accepted. Are you still actively working on this?

pfremm:
I am also curious.  @detailyang will you be working on a merge?

src/ngx_http_lua_ssl.c
```diff
@@ -34,4 +38,504 @@ ngx_http_lua_ssl_init(ngx_log_t *log)
 }
 
 
+static size_t
+ngx_http_lua_ssl_get_error(u_long e, u_char *ssl_err,
+    size_t ssl_err_len, const char *default_err, size_t default_err_len)
```
spacewander: Just put some notes here:
OpenSSL stores errors in a queue. And it might generate multiple errors in a call.
Therefore,
1. You need to drain the error queue to get all errors.
2. Someone might only fetch the first error, and leave the others in the queue.

Current implementation of this function doesn't satisfy the deduction 1 and is vulnerable to the deduction 2.

helinbo2015:
@detailyang  @agentzh 
how is the progress of this PR? i am eager that this PR can be accepted by master branch.

moonming:
@thibaultcha do you have time to take a look this PR?  thanks.

EnricoMazzu:
Hello, is there a plan to give support of mtls on this library?

PR #1001, opened 2017-03-01
Fixed SSL version error to be less ambiguous.

### LINKED ISSUE #1000: OpenSSL version error not correct.
Nginx reports this at runtime if the OpenSSL version is too old:

`2017/02/28 20:39:54 [emerg] 25466#25466: at least OpenSSL 1.0.2e required but found OpenSSL 1.0.1t  3 May 2016`

Upgrading the package on the system has no effect as Ngnix is checking the version it was compiled against, not the version that is installed. This error should state this instead, perhaps:

`Nginx was compiled with OpenSSL x, but requires at least 1.0.2e`

gnif:
Fixes #1000


config
```diff
@@ -281,6 +281,18 @@ END
             fi
             . auto/feature
         fi
+        
+        if [ $ngx_found = no ]; then
```
agentzh: Please do not include unrelated changes. Also, it's recommended to use OpenResty directly or to specify the `LUAJIT_LIB` and `LUAJIT_INC` environments properly if building the module yourself.

PR #1009, opened 2017-03-09
feature: allow use of ngx.exit() in the context of body_filter_by_lua_*

rainingmaster:
Ported from [header_filter_by_lua](https://github.com/openresty/lua-nginx-module/commit/3f8a592a), I have try to add  a new feature that add `ngx.exit()` in the context of `body_filter_by_lua`. But like in the context of `balance_by_lua_*`, we can not set return code. Run `ngx.exit()` in `body_filter_by_lua_*` will stop immediately, but return a code before.


src/ngx_http_lua_control.c
```diff
@@ -319,11 +319,18 @@ ngx_http_lua_ngx_exit(lua_State *L)
                                | NGX_HTTP_LUA_CONTEXT_CONTENT
                                | NGX_HTTP_LUA_CONTEXT_TIMER
                                | NGX_HTTP_LUA_CONTEXT_HEADER_FILTER
+                               | NGX_HTTP_LUA_CONTEXT_BODY_FILTER
                                | NGX_HTTP_LUA_CONTEXT_BALANCER
                                | NGX_HTTP_LUA_CONTEXT_SSL_CERT
                                | NGX_HTTP_LUA_CONTEXT_SSL_SESS_STORE
                                | NGX_HTTP_LUA_CONTEXT_SSL_SESS_FETCH);
 
+    /* rc is useless in bodyfilter phase */
```
agentzh: I think the only exit code makes sense in the output body filter is `NGX_ERROR` and 444.

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -140,6 +146,19 @@ ngx_http_lua_body_filter_by_chunk(lua_State *L, ngx_http_request_t *r,
...
+    if (ctx->body_exited) {
+        /* send last content */
+        for (cl = in; cl->next; cl = cl->next) {
+            /* do nothing */
+        }
+        /* make ngx_http_write_filter_module happy */
+        cl->buf->last_buf = 1;
+
+        ctx->eof = 1;
+
```
agentzh: For `ngx.exit(ngx.ERROR)` and `ngx.exit(444)`, it just makes no sense to send out a last buf or returning `NGX_OK` to the output body filter caller.

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -218,13 +231,7 @@ ngx_http_lua_body_filter_file(ngx_http_request_t *r, ngx_chain_t *in)
     /*  make sure we have a valid code chunk */
     ngx_http_lua_assert(lua_isfunction(L, -1));
 
-    rc = ngx_http_lua_body_filter_by_chunk(L, r, in);
-
-    if (rc != NGX_OK) {
-        return NGX_ERROR;
```
agentzh: Why remove this?

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -87,11 +87,15 @@ ngx_http_lua_body_filter_by_chunk(lua_State *L, ngx_http_request_t *r,
     ngx_chain_t *in)
 {
     ngx_int_t        rc;
+    ngx_chain_t     *cl;
     u_char          *err_msg;
     size_t           len;
 #if (NGX_PCRE)
     ngx_pool_t      *old_pool;
 #endif
+    ngx_http_lua_ctx_t          *ctx;
```
agentzh: Need a blank line before this line when you fail to align the variable name being declared with the previous lines.

src/ngx_http_lua_bodyfilterby.c
rainingmaster: @agentzh Because `ngx_http_lua_body_filter_by_chunk` is return `NGX_OK` or `NGX_ERROR` only, so I think this `if` is unnecessary.
Also, this is ported from https://github.com/openresty/lua-nginx-module/commit/3f8a592a#diff-24e9ee6ce24592d8abdf568146073458L211

agentzh: @rainingmaster OK. Fair enough.

src/ngx_http_lua_control.c
rainingmaster: We come across a situation where we need to constantly compare a flag to determine whether or not to stop the output stream, so we need to read the flag in `body_filter` and stop it. At this point because the Http response header has been sent, so I think `rc` is not too much significance

agentzh:
@rainingmaster Always returning `NGX_OK` in our body filter handler on the C land is wrong. We should return `NGX_ERROR` to signal the content handler to stop continuing immediately.

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -86,12 +86,16 @@ ngx_int_t
...
+    ngx_chain_t                 *cl;
+    u_char                      *err_msg;
+    size_t                       len;
 #if (NGX_PCRE)
-    ngx_pool_t      *old_pool;
+    ngx_pool_t                  *old_pool;
 #endif
+    ngx_http_lua_ctx_t          *ctx;
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
```
rainingmaster: @agentzh fix.
I think it will add blank lines to mix into the following code, so I forced all of its variable declaration

rainingmaster:
@agentzh [run `ngx.exit()` first time in `body_filter`](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-a6f1336f42adee1a8437f5073832da13R159) will return `NGX_OK` because I think we should use follow body_filter to filter the output buffer， and [`NGX_ERROR` will be return](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-a6f1336f42adee1a8437f5073832da13R275) while `ctx.body_exited` sign to stop continuing immediately.
Do you mean we should judge the `rc` is `ngx.ERROR` or the argument is a code no smaller than 300 when `ngx.exit()` is executed in the bodyfilter phase?

agentzh:
@rainingmaster Yes, otherwise the content handler will simply generate more data and continue calling the body filter chain.

rainingmaster:
@agentzh Should we judge the `rc` is `ngx.ERROR` or the argument is a code no smaller than 300 [here](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-198bce3a7eb5e4a6fdc54aae29b46483R492)?

src/ngx_http_lua_control.c
agentzh: @rainingmaster But we have to reject meaningless status code when the user tries. Silently ignoring things can make the user scratch her head.

rainingmaster: @agentzh Can we do this: just log an error log ([like this one](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_control.c#L371)) in [here](https://github.com/rainingmaster/lua-nginx-module/blob/bodyfilter_exit/src/ngx_http_lua_control.c#L333). Because the difference between the output of NGX_OK or NGX_ERROR in `ngx_http_lua_body_filter_by_chunk` is only the output of the most have a buf, the user may be difficult to see the difference? 
Now, once the user calls the `ngx.exit()` in body fitler, it will output the last buf and exit.

agentzh:
@rainingmaster I think we should just throw out a Lua exception when a meaningless status code is provided as the argument. We should limit automatic nginx error logging in our Lua API. Catch the error on the low level, and let the upper levels handle the error. This `ngx.exit()` is a low level API, so it should not handle the error by make the decision by itself to log the error.

rainingmaster:
@agentzh  Well, I've tried to change the nginx error logging to a Lua exception. Is that right like [this](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-198bce3a7eb5e4a6fdc54aae29b46483R332)?

rainingmaster:
@agentzh  Can you help me see if there are any more questions? We plan to use massively online. Thanks：)

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -140,6 +146,19 @@ ngx_http_lua_body_filter_by_chunk(lua_State *L, ngx_http_request_t *r,
...
+        /* send last content */
+        for (cl = in; cl->next; cl = cl->next) {
+            /* do nothing */
+        }
+        /* make ngx_http_write_filter_module happy */
+        cl->buf->last_buf = 1;
+
+        ctx->eof = 1;
+
+        return NGX_OK;
```
rainingmaster: @agentzh Thanks for reply~
I intend to let it continue to follow `bodyfilter`. My idea is to exit in the bodyfiler phase, a bit of the scene is likely to be the output of the general body, so I hope that before the output of all the accumulated buffer. [`Ngx_OK` can be used here](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-a6f1336f42adee1a8437f5073832da13R159) to pass the rest of the bodyfiler phase (such as ngx_http_gzip_body_filter), output all of the remaining content, to prevent clients from accessing any content. [return `ngx.ERROR` here](https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-a6f1336f42adee1a8437f5073832da13R275) prevent again into the bodyfilter stage, behind the content is no longer the output.

agentzh: @rainingmaster Your use case looks wrong from the perspective of the HTTP protocol. You are essentially truncating the original response body but still pretend to the client to be well formed. This is just wrong. The client should be notified of this data truncation instead of being cheated. What exactly is your real world business-level use case?

agentzh: @rainingmaster Also, according to our and nginx's convention, returning (or exiting with) `NGX_OK` usually means exiting the current *handler invocation* successfully, instead of exiting the whole request successfully. The only exception is the content phase handler, which is special anyway. In this context, returning (or exiting with) `NGX_OK` from a body filter handler should not end the whole response body, but just quit the current body filter handler invocation, but it should not prevent the current body filter handler from being invoked again in the same request.

BTW, I'm having problems reading your English text in your comment above. I'm just assuming that I understand your comment and intention correctly here, with reference to your patch here.

rainingmaster: @agentzh  I agree with your opinion that send a broken response body is wrong. 
We just want to exit to bodyfilter and stop to send remaining response body while we get some problem in `bodyfilter` phase. But in this phase response header maybe has sent. How can we deal with status like this?
Sorry of my poor English, most of them is come from google's translation :(

agentzh: @rainingmaster In your case, you should just use `ngx.exit(ngx.ERROR)` to abort the response body. Data already sent out on the wire cannot be undone. The client should handle the case of a truncated response when it detects truncation. Your current way of cheating client by sending a last chunk terminator just makes client impossible to detect truncation.

rainingmaster: @agentzh Sorry, I have a special condition to forget to describe: because we get the content from the nginx's upstream module, so you can only add some Lua operation in the `bodyfilter` phase. can we exit `ngx.ERROR` at bodyfilter and let client detects truncation? How can I do? thanks a lot

agentzh: @rainingmaster Just return `NGX_ERROR` from the body filter handler on the C land. That will do the trick.

rainingmaster: @agentzh But sometimes client get nothing because we did not  sent anything while bodyfilter return `NGX_ERROR`, should the client consider this a normal response?

agentzh: @rainingmaster In that case, the client would see a premature connection abort, which is reasonable as well.

rainingmaster: @agentzh Yes, that's more reasonable in that case.
However, if we run exit in bodyfilter after sending http response header,we will still encounter the unreasonable situation mentioned above...

agentzh: @rainingmaster It's completely reasonable otherwise you have to fully buffer the whole response before sending a single byte out to the wire.

agentzh: Alas. I'm already tired of explain this back and forth, and over and over again.

rainingmaster: @agentzh Sorry to disturb you...
I have try modified to use `ngx.ERROR` in body filter now. I hope this can be able to can provide some help for a friend in need about exiting in body filter

hongliang5316:
@agentzh 
Our customers have such a distorted demand scenario:
They require real-time control of an asynchronous request, when the request comes in, we need to request the third party asynchronous interface, third party interface will tell us whether the request should be terminated immediately, but the length of time requested by the third party interface does not affect the flow of this request,If this request has not yet ended, the third party interface is required to terminate, then we must terminate this request, in order to achieve this effect, we consider the implementation of the body_filter phase.
Can you have another idea about this?Thanks for your help very much.

luksbelt:
Question: this feature will be included in some future release? thanks

agentzh:
@luksbelt Well, @thibaultcha might help take care of this.

rainingmaster:
@thibaultcha could you help me check these code, I have rebase the newest code now, thank you

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -136,7 +141,8 @@ ngx_http_lua_body_filter_by_chunk(lua_State *L, ngx_http_request_t *r,
 
     lua_settop(L, 0);
 
-    if (rc == NGX_ERROR || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
+    if (ctx->body_filter_aborted || rc == NGX_ERROR
+        || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
```
doujiang24: style: the following style can be better.
```suggestion
        if (ctx->body_filter_aborted
            || rc == NGX_ERROR
            || rc >= NGX_HTTP_SPECIAL_RESPONSE)
        {
```

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -264,6 +271,10 @@ ngx_http_lua_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         }
     }
 
+    if (ctx->body_exited) {
+        return NGX_ERROR;
```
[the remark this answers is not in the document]
doujiang24: @rainingmaster you missed this comment?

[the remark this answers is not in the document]
rainingmaster: @doujiang24 Thanks for your remind. Ffinish now. https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-a6f1336f42adee1a8437f5073832da13R262

src/ngx_http_lua_bodyfilterby.c
rainingmaster: OK, we only allow 444 here now. https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-198bce3a7eb5e4a6fdc54aae29b46483R381

src/ngx_http_lua_common.h
```diff
@@ -487,6 +487,8 @@ typedef struct ngx_http_lua_ctx_s {
 
     unsigned         exited:1;
 
+    unsigned         body_exited:1;
```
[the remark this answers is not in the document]
rainingmaster: Finish now: https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-55d5b0e2cfa68aeb2d83e4507fdf70b5R562

src/ngx_http_lua_control.c
```diff
@@ -480,6 +493,19 @@ ngx_http_lua_ffi_exit(ngx_http_request_t *r, int status, u_char *err,
         return NGX_ERROR;
     }
 
+    if (ctx->context & NGX_HTTP_LUA_CONTEXT_BODY_FILTER) {
+
+        if (status != NGX_ERROR && status < 400) {
```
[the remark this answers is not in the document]
rainingmaster: finish now: https://github.com/openresty/lua-nginx-module/pull/1009/files#diff-198bce3a7eb5e4a6fdc54aae29b46483R381

rainingmaster:
@doujiang24  Finish, could you help me check again?

rainingmaster:
@thibaultcha Hello, could you help me review this pr? So thanks

Issue #1016, opened 2017-03-21
Nginx 1.11.11 Compilation Error


charlesportwoodii:
Hi again,

The Lua module fails to compile with Nginx 1.11.11. Just wanted to get this on the radar.

### Compilation Error
```bash
modules/lua-nginx-module/src/ngx_http_lua_headers.c: In function ‘ngx_http_lua_ngx_req_raw_header’:
modules/lua-nginx-module/src/ngx_http_lua_headers.c:145:15: error: incompatible types when assigning to type ‘ngx_buf_t * {aka struct ngx_buf_s *}’ from type ‘ngx_chain_t {aka struct ngx_chain_s}’
             b = hc->busy[i];
               ^
modules/lua-nginx-module/src/ngx_http_lua_headers.c:221:15: error: incompatible types when assigning to type ‘ngx_buf_t * {aka struct ngx_buf_s *}’ from type ‘ngx_chain_t {aka struct ngx_chain_s}’
             b = hc->busy[i];
               ^
objs/Makefile:1935: recipe for target 'objs/addon/src/ngx_http_lua_headers.o' failed
make[1]: *** [objs/addon/src/ngx_http_lua_headers.o] Error 1
make[1]: Leaving directory '/tmp/nginx-1.11.11'
Makefile:8: recipe for target 'build' failed
make: *** [build] Error 2
```

### Versions
Nginx: 1.11.11
Lua Module: v0.10.7

### Related Issues:
Related: https://github.com/openresty/echo-nginx-module/issues/64

### Notes

>     *) Bugfix: a segmentation fault might occur in a worker process if
>       number of large_client_header_buffers in a virtual server was
>       different from the one in the default server.

This appears to be the offending commit: https://github.com/nginx/nginx/commit/fca26c2e53d46549320a3b23ce0398c5921348dd

agentzh:
@charlesportwoodii Patches welcome.

charlesportwoodii:
@agentzh I know. =) I was looking at creating a patch earlier today but got hung up on other things. Not sure if I'm going to be able to get to creating a patch this week or not. I'll submit a PR if I can get some time this week to take a deeper look at it.

defanator:
@charlesportwoodii, @agentzh - I've just submitted PR #1017 that fixes the build. I also did a simple check with `ngx.req.raw_header()` and it seems to work.

neilstuartcraig:
[...]

PR #1017, opened 2017-03-22
fixed build with nginx/1.11.11

defanator:
Closes openresty/lua-nginx-module#1016

See also:
http://hg.nginx.org/nginx/rev/e662cbf1b932

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -78,12 +78,12 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     ngx_buf_t                   *b, *first = NULL;
     ngx_int_t                    i, j;
 #if defined(nginx_version) && nginx_version >= 1011011
-    ngx_buf_t                 **bb;
+    ngx_buf_t                  **bb;
     ngx_chain_t                 *cl;
-    ngx_array_t                 *busy_bufs;
 #endif
     ngx_connection_t            *c;
     ngx_http_request_t          *r, *mr;
+    ngx_http_lua_ctx_t          *ctx;
     ngx_http_connection_t       *hc;
 
     n = lua_gettop(L);
@@ -109,6 +109,11 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         return luaL_error(L, "http v2 not supported yet");
     }
 #endif
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
+    if (ctx == NULL) {
+        return luaL_error(L, "no ctx found");
+    }
 
 #if 1
     dd("hc->nbusy: %d", (int) hc->nbusy);
@@ -156,21 +161,30 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     dd("size: %d", (int) size);
 
 #if defined(nginx_version) && nginx_version >= 1011011
-    busy_bufs = ngx_array_create(r->pool, hc->nbusy + 1, sizeof(ngx_buf_t *));
+    if (hc->nbusy && ctx->busy_bufs_ptrs == NULL) {
+        ctx->busy_bufs_ptrs = ngx_array_create(r->pool, hc->nbusy, sizeof(ngx_buf_t *));
+        if (ctx->busy_bufs_ptrs == NULL) {
+            return luaL_error(L, "no memory");
+        }
+
+        for (cl = hc->busy; cl; /* void */) {
+            bb = ngx_array_push(ctx->busy_bufs_ptrs);
+            if (bb == NULL) {
+                return luaL_error(L, "no memory");
+            }
+
+            *bb = cl->buf;
+            cl = cl->next;
+        }
+    }
 #endif
 
     if (hc->nbusy) {
         b = NULL;
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        for (cl = hc->busy; cl; /* void */) {
-            bb = ngx_array_push(busy_bufs);
-            *bb = cl->buf;
-            cl = cl->next;
-        }
-
-        bb = busy_bufs->elts;
-        for (i = busy_bufs->nelts; i > 0; i--) {
+        bb = ctx->busy_bufs_ptrs->elts;
+        for (i = ctx->busy_bufs_ptrs->nelts; i > 0; i--) {
             b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
@@ -252,8 +266,8 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (hc->nbusy) {
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = busy_bufs->elts;
-        for (i = busy_bufs->nelts; i > 0; i--) {
+        bb = ctx->busy_bufs_ptrs->elts;
+        for (i = ctx->busy_bufs_ptrs->nelts; i > 0; i--) {
             b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -80,10 +80,10 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 #if defined(nginx_version) && nginx_version >= 1011011
     ngx_buf_t                  **bb;
     ngx_chain_t                 *cl;
+    ngx_http_lua_srv_conf_t     *lscf;
 #endif
     ngx_connection_t            *c;
     ngx_http_request_t          *r, *mr;
-    ngx_http_lua_ctx_t          *ctx;
     ngx_http_connection_t       *hc;
 
     n = lua_gettop(L);
@@ -109,11 +109,6 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         return luaL_error(L, "http v2 not supported yet");
     }
 #endif
-
-    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
-    if (ctx == NULL) {
-        return luaL_error(L, "no ctx found");
-    }
 
 #if 1
     dd("hc->nbusy: %d", (int) hc->nbusy);
@@ -161,20 +156,12 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     dd("size: %d", (int) size);
 
 #if defined(nginx_version) && nginx_version >= 1011011
-    if (hc->nbusy && ctx->busy_bufs_ptrs == NULL) {
-        ctx->busy_bufs_ptrs = ngx_array_create(r->pool, hc->nbusy, sizeof(ngx_buf_t *));
-        if (ctx->busy_bufs_ptrs == NULL) {
-            return luaL_error(L, "no memory");
-        }
+    lscf = ngx_http_get_module_srv_conf(r, ngx_http_lua_module);
 
-        for (cl = hc->busy; cl; /* void */) {
-            bb = ngx_array_push(ctx->busy_bufs_ptrs);
-            if (bb == NULL) {
-                return luaL_error(L, "no memory");
-            }
-
-            *bb = cl->buf;
-            cl = cl->next;
+    if (hc->nbusy) {
+        bb = lscf->busy_bufs_ptrs;
+        for (cl = hc->busy; cl; cl = cl->next) {
+            *bb++ = cl->buf;
         }
     }
 #endif
@@ -183,8 +170,8 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         b = NULL;
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = ctx->busy_bufs_ptrs->elts;
-        for (i = ctx->busy_bufs_ptrs->nelts; i > 0; i--) {
+        bb = lscf->busy_bufs_ptrs;
+        for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
@@ -266,8 +253,8 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (hc->nbusy) {
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = ctx->busy_bufs_ptrs->elts;
-        for (i = ctx->busy_bufs_ptrs->nelts; i > 0; i--) {
+        bb = lscf->busy_bufs_ptrs;
+        for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -28,6 +28,12 @@ static int ngx_http_lua_ngx_req_header_set(lua_State *L);
 static int ngx_http_lua_ngx_resp_get_headers(lua_State *L);
 
 
+#if defined(nginx_version) && nginx_version >= 1011011
+static ngx_buf_t  **busy_bufs_ptrs = NULL;
+static ngx_int_t    prealloc_nbusy = 0;
+#endif
+
+
 static int
 ngx_http_lua_ngx_req_http_version(lua_State *L)
 {
@@ -80,7 +86,6 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 #if defined(nginx_version) && nginx_version >= 1011011
     ngx_buf_t                  **bb;
     ngx_chain_t                 *cl;
-    ngx_http_lua_srv_conf_t     *lscf;
 #endif
     ngx_connection_t            *c;
     ngx_http_request_t          *r, *mr;
@@ -155,22 +160,31 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     dd("size: %d", (int) size);
 
-#if defined(nginx_version) && nginx_version >= 1011011
-    lscf = ngx_http_get_module_srv_conf(r, ngx_http_lua_module);
-
     if (hc->nbusy) {
-        bb = lscf->busy_bufs_ptrs;
+#if defined(nginx_version) && nginx_version >= 1011011
+        if (hc->nbusy > prealloc_nbusy) {
+            if (busy_bufs_ptrs != NULL)
+                ngx_pfree(ngx_cycle->pool, busy_bufs_ptrs);
+
+            busy_bufs_ptrs = ngx_palloc(ngx_cycle->pool,
+                                        hc->nbusy * sizeof(ngx_buf_t *));
+
+            if (busy_bufs_ptrs == NULL) {
+                return luaL_error(L, "no memory");
+            }
+
+            prealloc_nbusy = hc->nbusy;
+        }
+
+        bb = busy_bufs_ptrs;
         for (cl = hc->busy; cl; cl = cl->next) {
             *bb++ = cl->buf;
         }
-    }
 #endif
-
-    if (hc->nbusy) {
         b = NULL;
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = lscf->busy_bufs_ptrs;
+        bb = busy_bufs_ptrs;
         for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
@@ -253,7 +267,7 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (hc->nbusy) {
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = lscf->busy_bufs_ptrs;
+        bb = busy_bufs_ptrs;
         for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -170,8 +170,8 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         }
 
         bb = busy_bufs->elts;
-        for (i = busy_bufs->nelts - 1; i > 0; i--) {
-            b = bb[i];
+        for (i = busy_bufs->nelts; i > 0; i--) {
+            b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
             b = hc->busy[i];
@@ -253,8 +253,8 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
 #if defined(nginx_version) && nginx_version >= 1011011
         bb = busy_bufs->elts;
-        for (i = busy_bufs->nelts - 1; i > 0; i--) {
-            b = bb[i];
+        for (i = busy_bufs->nelts; i > 0; i--) {
+            b = bb[i - 1];
 #else
         for (i = 0; i < hc->nbusy; i++) {
             b = hc->busy[i];
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -28,12 +28,6 @@ static int ngx_http_lua_ngx_req_header_set(lua_State *L);
 static int ngx_http_lua_ngx_resp_get_headers(lua_State *L);
 
 
-#if defined(nginx_version) && nginx_version >= 1011011
-static ngx_buf_t  **busy_bufs_ptrs = NULL;
-static ngx_int_t    prealloc_nbusy = 0;
-#endif
-
-
 static int
 ngx_http_lua_ngx_req_http_version(lua_State *L)
 {
@@ -86,6 +80,7 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 #if defined(nginx_version) && nginx_version >= 1011011
     ngx_buf_t                  **bb;
     ngx_chain_t                 *cl;
+    ngx_http_lua_main_conf_t    *lmcf;
 #endif
     ngx_connection_t            *c;
     ngx_http_request_t          *r, *mr;
@@ -102,6 +97,10 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (r == NULL) {
         return luaL_error(L, "no request object found");
     }
+
+#if defined(nginx_version) && nginx_version >= 1011011
+    lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
+#endif
 
     ngx_http_lua_check_fake_request(L, r);
 
@@ -162,21 +161,22 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     if (hc->nbusy) {
 #if defined(nginx_version) && nginx_version >= 1011011
-        if (hc->nbusy > prealloc_nbusy) {
-            if (busy_bufs_ptrs != NULL)
-                ngx_pfree(ngx_cycle->pool, busy_bufs_ptrs);
+        if (hc->nbusy > lmcf->prealloc_nbusy) {
+            if (lmcf->busy_bufs_ptrs) {
+                ngx_free(lmcf->busy_bufs_ptrs);
+            }
 
-            busy_bufs_ptrs = ngx_palloc(ngx_cycle->pool,
-                                        hc->nbusy * sizeof(ngx_buf_t *));
+            lmcf->busy_bufs_ptrs = ngx_alloc(hc->nbusy * sizeof(ngx_buf_t *),
+                                             r->connection->log);
 
-            if (busy_bufs_ptrs == NULL) {
+            if (lmcf->busy_bufs_ptrs == NULL) {
                 return luaL_error(L, "no memory");
             }
 
-            prealloc_nbusy = hc->nbusy;
+            lmcf->prealloc_nbusy = hc->nbusy;
         }
 
-        bb = busy_bufs_ptrs;
+        bb = lmcf->busy_bufs_ptrs;
         for (cl = hc->busy; cl; cl = cl->next) {
             *bb++ = cl->buf;
         }
@@ -184,7 +184,7 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         b = NULL;
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = busy_bufs_ptrs;
+        bb = lmcf->busy_bufs_ptrs;
         for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
@@ -267,7 +267,7 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (hc->nbusy) {
 
 #if defined(nginx_version) && nginx_version >= 1011011
-        bb = busy_bufs_ptrs;
+        bb = lmcf->busy_bufs_ptrs;
         for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -77,7 +77,9 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     size_t                       size;
     ngx_buf_t                   *b, *first = NULL;
     ngx_int_t                    i, j;
-    ngx_chain_t                 *cl, *ln;
+#if defined(nginx_version) && nginx_version >= 1011011
+    ngx_chain_t                 *cl;
+#endif
     ngx_connection_t            *c;
     ngx_http_request_t          *r, *mr;
     ngx_http_connection_t       *hc;
@@ -148,11 +150,15 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     if (hc->nbusy) {
         b = NULL;
-        for (cl = hc->busy; cl; /* void */) {
-            ln = cl;
-            cl = cl->next;
 
-            b = ln->buf;
+#if defined(nginx_version) && nginx_version >= 1011011
+        for (cl = hc->busy; cl; /* void */) {
+            b = cl->buf;
+            cl = cl->next;
+#else
+        for (i = 0; i < hc->nbusy; i++) {
+            b = hc->busy[i];
+#endif
 
             dd("busy buf: %d: [%.*s]", (int) i, (int) (b->pos - b->start),
                b->start);
@@ -227,11 +233,15 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     }
 
     if (hc->nbusy) {
-        for (cl = hc->busy; cl; /* void */) {
-            ln = cl;
-            cl = cl->next;
 
-            b = ln->buf;
+#if defined(nginx_version) && nginx_version >= 1011011
+        for (cl = hc->busy; cl; /* void */) {
+            b = cl->buf;
+            cl = cl->next;
+#else
+        for (i = 0; i < hc->nbusy; i++) {
+            b = hc->busy[i];
+#endif
 
             if (!found) {
                 if (b != first) {
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -78,7 +78,9 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     ngx_buf_t                   *b, *first = NULL;
     ngx_int_t                    i, j;
 #if defined(nginx_version) && nginx_version >= 1011011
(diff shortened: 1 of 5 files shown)


agentzh:
@defanator You broke builds with earlier versions of the nginx core, as evidenced in the travis ci report of this PR:

https://travis-ci.org/openresty/lua-nginx-module/builds/213702739

agentzh:
@defanator Also, I think you should enable nginx 1.11.11 in the travis ci matrix by editing the `.travis.yml` file in your branch for this PR.

defanator:
@agentzh, yep, already working on a fix, will add to the PR shortly.

defanator:
Done. I'm not sure it will pass though, as it seems like the same fix will be needed in echo module (at least).

defanator:
Ok, it did not pass due to other reasons, but I've just submitted https://github.com/openresty/echo-nginx-module/pull/65 anyway.

agentzh:
@defanator The Travis CI tests are failing for nginx 1.11.11.

agentzh:
@defanator Okay, it was due to the patches from the `openresty/openresty` repository.

msva:
@defanator can you re-make this PR by changing NgX version in `.travis.yml` to 1.11.2, so @agentzh will be able to finally merge this PR?

lucasRolff:
Any update on this? :octocat: would like to build a newer nginx version with the lua module 👍

agentzh:
@dndx Will you look at this as well? Many thanks!

dndx:
@agentzh This fix LGTM. Once `1.11.13`'s patch sets are created, hopefully the build will turn green as well.

agentzh:
@defanator Although your patch makes this module builds with nginx 1.11.11, the test file `t/104-req-raw-header.t` in this module's test suite is failing horribly:

https://gist.github.com/agentzh/d2f46d3f822bd21a6085bd83a725ee9d

Obviously you need to get the whole test suite passing before this patch can be merged :)

Thanks!

dndx:
@agentzh @defanator I was able to repro the issue locally as well (good incentive to run tests next time before I greenlight a change..)

I think the issue is caused by a logic mismatch after the change.

Before https://github.com/nginx/nginx/commit/23946be384c42478f3dbc661f60ac0903d1b97f2, [`hc->busy` used to be in the order that headers were received](https://github.com/nginx/nginx/commit/fca26c2e53d46549320a3b23ce0398c5921348dd#diff-d57c27b3aa34ba85676b639a618161c7L1494). That is, if the client sent us:

```
A: [very large content]
B: [very large content]
C: [very large content]
```

Then `hc->busy` would be in format similar to this:

```
hc->busy[0] = "A: [very large content]"
hc->busy[1] = "B: [very large content]"
hc->busy[2] = "C: [very large content]"
```

and that is what `ngx_http_lua_ngx_req_raw_header()` assumes.

However, inside https://github.com/nginx/nginx/commit/23946be384c42478f3dbc661f60ac0903d1b97f2 this is no longer the case. Since [now NGINX appends the newly created buffers to the front of the linked list](https://github.com/nginx/nginx/commit/fca26c2e53d46549320a3b23ce0398c5921348dd#diff-d57c27b3aa34ba85676b639a618161c7R1498), with your change the loop now actually parses incoming buffers in reverse order they were reveived. That is, C is being seen first, followed by B and A. This breaks the assumption `ngx_http_lua_ngx_req_raw_header()` makes and causes subsequent logic to not work. Specifically, it will only copy the buffer that contains the request line and nothing afterwards (because `ngx.req.raw_header()` wants headers in order but the `->next` pointer of the first buffer client sent us is `NULL`, as it is the last one inside the chain link).

Unfortunately the fix won't be easy. In order to preserve current semantics of `ngx.req.raw_header()`, we potentially have to create another chain link in reverse order of `hc->busy` in order to iterate it correctly, since chain link are singly linked and we can not go backwards. If we decide to not do the reversal, then `ngx.req.raw_header()` could return headers in reverse order they are received from the client, which is probably not acceptable.

The reason why it appears to work earlier is that this bug is only evident if client sent us a lot of headers and we had to actually allocate more buffers into `hc->busy`.

@defanator Can you update your patch to fix this?

dndx:
Some debug log to show this behavior:

```
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 1: [Header510] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 2: [Header485] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 3: [Header460] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 4: [Header435] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 5: [Header410] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 6: [Header385] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 7: [Header360] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 8: [Header335] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 9: [Header310] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 10: [Header285] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 11: [Header260] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 12: [Header235] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 13: [Header210] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 14: [Header185] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 15: [Header160] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 16: [Header135] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 17: [Header110] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 18: [Header83] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 19: [Header55] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 20: [Header27] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159. 
lua *** ngx_http_lua_ngx_req_raw_header: busy buf: 21: [GET /t HTTP/1.1^M           
Host] at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 159.                     
lua *** ngx_http_lua_ngx_req_raw_header: found first at 21 at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 170. 
lua *** ngx_http_lua_ngx_req_raw_header: adding size 561 at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 174. 
lua *** ngx_http_lua_ngx_req_raw_header: header size: 562 at ../ngx_lua-0.10.7/src/ngx_http_lua_headers.c line 182.
```

defanator:
@dndx, I'll try to look at this - thanks for the insights!

defanator:
@dndx, @agentzh, I tried to implement reverse handling of busy buffers in nginx >= 1.11.11, but obviously it is not enough, I'm still getting the following failures in `t/104-req-raw-header.t`:

```
Test Summary Report
-------------------
t/104-req-raw-header.t (Wstat: 10240 Tests: 204 Failed: 40)
  Failed tests:  2, 5, 8, 11, 20, 23, 26, 29, 32, 35, 38
                41, 44, 47, 50, 53, 56, 59, 80, 83, 86
                89, 92, 95, 104, 107, 110, 113, 116, 119
                146, 149, 164, 167, 182, 185, 188, 191
                194, 197
  Non-zero exit status: 40
Files=1, Tests=204,  5 wallclock secs ( 0.04 usr  0.03 sys +  0.39 cusr  0.15 csys =  0.61 CPU)
Result: FAIL
```

agentzh:
@defanator I think you just need to debug the failures for your own changes :) These tests are passing completely with older nginx cores anyway.

defanator:
@agentzh, not sure I'll have enough time for this in upcoming days, but I'll try to dig further.

defanator:
Ok, I think I finally made it:

```
$ prove t/104-req-raw-header.t
[..]
All tests successful.
Files=1, Tests=204,  6 wallclock secs ( 0.06 usr  0.02 sys +  0.38 cusr  0.15 csys =  0.61 CPU)
Result: PASS
```

src/ngx_http_lua_headers.c
```diff
@@ -77,6 +77,11 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     size_t                       size;
     ngx_buf_t                   *b, *first = NULL;
     ngx_int_t                    i, j;
+#if defined(nginx_version) && nginx_version >= 1011011
+    ngx_buf_t                 **bb;
```
agentzh: Alignment problem? We should be consistent with the nginx coding style.

src/ngx_http_lua_headers.c
```diff
@@ -145,10 +155,27 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
...
+#if defined(nginx_version) && nginx_version >= 1011011
+    busy_bufs = ngx_array_create(r->pool, hc->nbusy + 1, sizeof(ngx_buf_t *));
+#endif
+
     if (hc->nbusy) {
         b = NULL;
+
+#if defined(nginx_version) && nginx_version >= 1011011
+        for (cl = hc->busy; cl; /* void */) {
+            bb = ngx_array_push(busy_bufs);
```
agentzh: I think we should check the case that `ngx_array_push` returns NULL. Please also fix other similar places.

src/ngx_http_lua_headers.c
```diff
@@ -145,10 +155,27 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     dd("size: %d", (int) size);
 
+#if defined(nginx_version) && nginx_version >= 1011011
+    busy_bufs = ngx_array_create(r->pool, hc->nbusy + 1, sizeof(ngx_buf_t *));
```
agentzh: `ngx_array_create` may return `NULL`. We should check that here.

I'm a bit worried about allocating inside the request memory pool since it will lead to "memory leaks" when the user Lua code calls this Lua API function in a tight loop. Can we allocate it elsewhere?

agentzh: Also, I don't think we need to use `ngx_array_t` here since this array does not have to grow.

dndx: Does not appears to me that the `+ 1` here is necessary when creating the array.

defanator:
@agentzh, @dndx, thanks for your comments and suggestions. I see the concern of allocating memory for busy buffers on every call to `ngx.req.raw_header` now. I'll try to address this in a way @dndx suggested.

defanator:
With the b00da5a, memory is being allocated once on the first call to `ngx.req.raw_header`.

In regards to using `ngx_array_t` in this particular case - I do not see any significant benefits in implementing separate allocation of an array of pointers or smth - do you? From my PoV, `ngx_array_t` fits fine here.

defanator:
One thought is that this check could be eliminated then: https://github.com/openresty/lua-nginx-module/pull/1017/files#diff-8593203ffbe2a651fb1761cebff6c139R172

Not sure if it something worthy.

src/ngx_http_lua_headers.c
```diff
@@ -145,10 +160,36 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
...
+#if defined(nginx_version) && nginx_version >= 1011011
+    if (hc->nbusy && ctx->busy_bufs_ptrs == NULL) {
+        ctx->busy_bufs_ptrs = ngx_array_create(r->pool, hc->nbusy, sizeof(ngx_buf_t *));
+        if (ctx->busy_bufs_ptrs == NULL) {
+            return luaL_error(L, "no memory");
+        }
+
+        for (cl = hc->busy; cl; /* void */) {
+            bb = ngx_array_push(ctx->busy_bufs_ptrs);
+            if (bb == NULL) {
```
agentzh: Better *keep* this check since it helps our `ngx-releng` static code analyzer.

src/ngx_http_lua_headers.c
```diff
@@ -145,10 +160,36 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
...
     dd("size: %d", (int) size);
 
+#if defined(nginx_version) && nginx_version >= 1011011
+    if (hc->nbusy && ctx->busy_bufs_ptrs == NULL) {
+        ctx->busy_bufs_ptrs = ngx_array_create(r->pool, hc->nbusy, sizeof(ngx_buf_t *));
+        if (ctx->busy_bufs_ptrs == NULL) {
+            return luaL_error(L, "no memory");
+        }
+
+        for (cl = hc->busy; cl; /* void */) {
```
agentzh: Why don't you put `cl = cl->next` here? That would read better.

src/ngx_http_lua_headers.c
```diff
@@ -145,10 +160,36 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     dd("size: %d", (int) size);
 
+#if defined(nginx_version) && nginx_version >= 1011011
+    if (hc->nbusy && ctx->busy_bufs_ptrs == NULL) {
+        ctx->busy_bufs_ptrs = ngx_array_create(r->pool, hc->nbusy, sizeof(ngx_buf_t *));
```
agentzh: I still don't see why we should use `ngx_array_t` here, as mentioned in my previous round of review. This array never needs to grow IMHO. It's more efficient both CPU-wise and memory-wise to just allocate a flat `ngx_buf_t` C array here.

agentzh:
@defanator @dndx I think that temporary array should be shared across all the requests instead of letting each request allocating a new array in its request memory pool, which is wasteful. Let's hook it up to the `ngx_http_lua_main_conf_t` and share it across the request boundaries.

agentzh:
Because we have production users doing C2M with OpenResty, so every byte counts :)

defanator:
@agentzh, are you sure that sticking to the `ngx_http_lua_main_conf_t` will work in a case when there are several parallel requests running with `ngx.req.raw_header` being used inside? (I'm worrying about contexts here.)

In general, configuration pool should be used for static configuration parts, as I understand it. Everything that subject to change in runtime (i.e., busy buffers) should be allocated elsewhere.

agentzh:
@defanator `ngx.req.raw_header()` never yields *and* the actual use of that array is limited to the lifetime of the `ngx.req.raw_header()` call (since the resulting data is copied into the Lua land before it returns anyway). Therefore, it's completely safe to share the array on the global level. What's your concern then?

dndx:
@defanator @agentzh also need to pay attention that [`large_client_header_buffers`](large_client_header_buffers) is allowed to be overwritten at `server` level, so maybe we have to allocate it inside `ngx_http_lua_srv_conf_t` instead of `ngx_http_lua_main_conf_t`. Since `ngx_http_lua_srv_conf_t` is a union.

centminmod:
did this pull request fix break again with lua nginx 0.10.8 and nginx 1.11.13 ? 

```
bjs -I objs/addon/ndk -I src/stream \
        -o objs/addon/src/ngx_http_lua_capturefilter.o \
        ../lua-nginx-module-0.10.8/src/ngx_http_lua_capturefilter.c
../lua-nginx-module-0.10.8/src/ngx_http_lua_headers.c:774:13: error: implicit declaration of function 'luaL_getn' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
        n = luaL_getn(L, 3);
            ^
1 error generated.
ccache /usr/bin/clang -ferror-limit=0 -c -fPIC -I/usr/local/include/luajit-2.1  -pipe  -O -Wall -Wextra -Wpointer-arith -Wconditional-uninitialized -Wno-unused-parameter -Werror -g -m64 -mtune=native -mfpmath=sse -g -O3 -fstack-protector -fuse-ld=gold --param=ssp-buffer-size=4 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wno-sign-compare -Wno-string-plus-int -Wno-deprecated-declarations -Wno-unused-parameter -Wno-unused-const-variable -Wno-conditional-uninitialized -Wno-mismatched-tags -Wno-sometimes-uninitialized -Wno-parentheses-equality -Wno-tautological-compare -Wno-self-assign -Wno-deprecated-register -Wno-deprecated -Wno-invalid-source-encoding -Wno-pointer-sign -Wno-parentheses -Wno-enum-conversion -Wno-c++11-compat-deprecated-writable-strings -Wno-write-strings -gsplit-dwarf -DNDK_SET_VAR -DNDK_UPSTREAM_LIST -DNDK_SET_VAR -I src/core -I src/event -I src/event/modules -I src/os/unix -I ../ngx_devel_kit-0.3.0/objs -I objs/addon/ndk -I ../lua-nginx-module-0.10.8/src/api -I /usr/local/include/luajit-2.1 -I ../pcre-8.40 -I ../libressl-2.4.5/.openssl/include -I ../zlib-1.2.11 -I objs -I src/http -I src/http/modules -I src/http/v2 -I ../ngx_devel_kit-0.3.0/src -I ../ngx_devel_kit-0.3.0/src -I ../ngx_devel_kit-0.3.0/objs -I objs/addon/ndk -I src/stream \
        -o objs/addon/src/ngx_http_lua_clfactory.o \
        ../lua-nginx-module-0.10.8/src/ngx_http_lua_clfactory.c
make[1]: *** [objs/addon/src/ngx_http_lua_headers.o] Error 1
make[1]: Leaving directory `/svr-setup/nginx-1.11.13'
make: *** [build] Error 2
```

spacewander:
@centminmod 
It might refer to another issue:
https://github.com/openresty/lua-nginx-module/pull/1030

centminmod:
@spacewander indeed it does :)

defanator:
@agentzh - I see now - thanks for the clarification. As pointed by @dndx, `large_client_header_buffers` may be overridden on server level [1], so I moved pre-allocation to the `ngx_http_lua_merge_srv_conf()`.

[1] http://nginx.org/r/large_client_header_buffers

agentzh:
@defanator Putting it on the server level is not ideal since we may have a lot of `ngx_http_lua_srv_conf_t` allocated for many upstream blocks and many `server {}` blocks. I think we can simply check if the current `nbusy` number is larger than the currently allocated global array, if yes, then simply deallocate the existing one and allocate a bigger array.

Simply put, we only need to allocate a buffer of the size of the largest `nbusy` number ever configured and that virtual server is actually using `ngx.req.raw_header`.

It's more efficient than making the `ngx_http_lua_srv_conf_t` struct larger.

defanator:
@agentzh, previously you've stated that using `ngx_array_t` is not needed there since it never has to grow. Should we finally return to the idea of using `ngx_array_t` allocated somewhere in the `ngx_http_lua_main_conf_t`, given your latest comment?

(My thought here is that `ngx_array_t` already has `nelts` field suitable for comparison with `nbusy`, so there is no need to add another variable for that.)

defanator:
After some attempts to re-iterate with `ngx_array_t` in configuration contexts and pre-allocating memory during the configuration parsing stage, I realized that the better way would be to introduce a couple of variables directly in the `ngx_http_lua_headers.c` and use those in runtime.

src/ngx_http_lua_headers.c
```diff
@@ -146,9 +161,36 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     dd("size: %d", (int) size);
 
     if (hc->nbusy) {
+#if defined(nginx_version) && nginx_version >= 1011011
+        if (hc->nbusy > prealloc_nbusy) {
+            if (busy_bufs_ptrs != NULL)
+                ngx_pfree(ngx_cycle->pool, busy_bufs_ptrs);
```
agentzh: The nginx coding style requires `{}` to be used for `if` statement at all times.

src/ngx_http_lua_headers.c
```diff
@@ -146,9 +161,36 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
...
+
+            busy_bufs_ptrs = ngx_palloc(ngx_cycle->pool,
```
agentzh: It is generally a bad idea to use the nginx memory pool for this since relatively small memory blocks *never* get freed in `ngx_pfree`. We should use `ngx_alloc` and `ngx_free` here directly.

src/ngx_http_lua_headers.c
```diff
@@ -28,6 +28,12 @@ static int ngx_http_lua_ngx_req_header_set(lua_State *L);
 static int ngx_http_lua_ngx_resp_get_headers(lua_State *L);
 
 
+#if defined(nginx_version) && nginx_version >= 1011011
+static ngx_buf_t  **busy_bufs_ptrs = NULL;
+static ngx_int_t    prealloc_nbusy = 0;
```
agentzh: This is a bad idea to use global C variables since we have to be super careful for global variable handling during HUP reload and etc. Better use the `main_conf_t` instead to avoid all such headaches.

src/ngx_http_lua_module.c
```diff
@@ -1020,6 +1020,7 @@ ngx_http_lua_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
     }
 
 #endif  /* NGX_HTTP_SSL */
+
```
agentzh: Please avoid unrelated changes.

dndx:
@defanator I took a look at your latest change and it seems to be OK. Can you fix the failed tests really quick?

defanator:
@dndx, you mean travis-ci tests? In that case,

 - for the 1.9.15, I don't quite understand how failed `129-ssl-socket.t` is related to the changes proposed in this PR,

 - for the 1.11.11, I can take a look at the openresty patches and try to rebase those for the latest mainline at the moment (1.11.13), but it may take some time.

In regards to this PR - should I merge current master branch to my branch? Or "rebase" it? (I'm definitely not a git/github expert.)

msva:
By  the way, yes. Both failing tests doesn't look anyhow relevant to proposed 
changes and I'd bet at least 1.1.11-tests (.13, actually) should be fixed by 
upstream (openresty), instead.


> @dndx, you mean travis-ci tests? In that case,
> 
>  - for the 1.9.15, I don't quite understand how failed `129-ssl-socket.t` is
> related to the changes proposed in this PR,
> 
>  - for the 1.11.11, I can take a look at the openresty patches and try to
> rebase those for the latest mainline at the moment (1.11.13), but it may
> take some time.
> 
> In regards to this PR - should I merge current master branch to my branch?
> Or "rebase" it? (I'm definitely not a git/github expert.)

defanator:
Also, echo-nginx-module should also be fixed for the tests to pass: https://github.com/openresty/echo-nginx-module/pull/65

I'm going to proceed with that PR as well (based on the code in this one).

defanator:
Finally, we're going to do the next stable/mainline bump in upcoming days, so maybe it's worth to wait for 1.12.0/1.13.0 before proceeding with openresty patches.

dndx:
@defanator Also make sure that you add a cleanup handler to the cycle pool to free up the memory when cycle is destroyed. Otherwise it will be leaked after reload.

Make sure the build stays Valgrind clean after your change. You can do:

```
$ TEST_NGINX_USE_VALGRIND=1 prove -r t/
```

To confirm that.

defanator:
@dndx, thanks for pointing, made it in c053ad9.

defanator:
I've also checked with ASAN and did not found any leaks in `t/104-req-raw-header.t`.

msva:
1.12 was just released, by the way

src/ngx_http_lua_module.c
```diff
@@ -638,7 +639,8 @@ ngx_http_lua_init(ngx_conf_t *cf)
     volatile ngx_cycle_t       *saved_cycle;
     ngx_http_core_main_conf_t  *cmcf;
     ngx_http_lua_main_conf_t   *lmcf;
-#ifndef NGX_LUA_NO_FFI_API
+#if (!defined(NGX_LUA_NO_FFI_API)) \
```
agentzh: Style: extra parens around `!defined(...)`.

src/ngx_http_lua_module.c
```diff
@@ -638,7 +639,8 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
+    || (defined(nginx_version) && nginx_version >= 1011011)
```
agentzh: I don't think we need the `defined(nginx_version)` part since this module explicitly states in the documentation that at least nginx 1.6.0 is needed where this `nginx_version` macro is already introduced. Please fix other similar places as well. Thank you!

src/ngx_http_lua_headers.c
```diff
@@ -1415,4 +1467,19 @@ ngx_http_lua_ffi_get_resp_header(ngx_http_request_t *r,
...
+#if defined(nginx_version) && nginx_version >= 1011011
+void
+ngx_http_lua_ngx_raw_header_cleanup(void *data)
+{
+    ngx_http_lua_main_conf_t  *lmcf;
+
+    lmcf = (ngx_http_lua_main_conf_t *) data;
+
+    if (lmcf->busy_bufs_ptrs) {
+        ngx_free(lmcf->busy_bufs_ptrs);
```
agentzh: It's considered good practice to always clear out `lmcf->busy_bufs_ptrs` to avoid any chance of double frees.

defanator:
@agentzh, I used `if defined(nginx_version) ...` additional check because it is still being widely used across the sources:

```
$ grep -R -- "#if defined(nginx_version)" * | wc -l
      32
$ grep -R -- "#if nginx_version" * | wc -l
      19
```

I've removed it from changes made by this PR, but I'd suggest to do this for the rest of the code in sake of consistency.

Cleanup function has also been fixed to ensure that double free will never happen, thanks for pointing to this.

denji:
nginx 1.12.0+
```
src/ngx_http_lua_headers.c:151:15: error: assigning to 'ngx_buf_t *' (aka 'struct ngx_buf_s *') from incompatible type 'ngx_chain_t' (aka 'struct ngx_chain_s')
            b = hc->busy[i];
              ^ ~~~~~~~~~~~

src/ngx_http_lua_headers.c:227:15: error: assigning to 'ngx_buf_t *' (aka 'struct ngx_buf_s *') from incompatible type 'ngx_chain_t' (aka 'struct ngx_chain_s')
            b = hc->busy[i];
              ^ ~~~~~~~~~~~
```

defanator:
@denji, please make sure that you have applied changes proposed in this PR before compiling. It works fine with 1.12.0 as well.

agentzh:
@defanator Merged with the following patch. Thanks!

```diff
diff --git a/.travis.yml b/.travis.yml
index 2c563527..7534c3c6 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,7 +50,7 @@ env:
     - TEST_NGINX_SLEEP=0.006
   matrix:
     - NGINX_VERSION=1.9.15
-    - NGINX_VERSION=1.11.11
+    - NGINX_VERSION=1.11.2
 
 services:
  - memcache
diff --git a/src/ngx_http_lua_common.h b/src/ngx_http_lua_common.h
index f2cfbe8a..f37d776a 100644
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -200,8 +200,9 @@ struct ngx_http_lua_main_conf_s {
     ngx_uint_t           malloc_trim_req_count;
 
 #if nginx_version >= 1011011
-    ngx_buf_t          **busy_bufs_ptrs;
-    ngx_int_t            prealloc_nbusy;
+    /* the following 2 fields are only used by ngx.req.raw_headers() for now */
+    ngx_buf_t          **busy_buf_ptrs;
+    ngx_int_t            busy_buf_ptr_count;
 #endif
 
     unsigned             requires_header_filter:1;
diff --git a/src/ngx_http_lua_headers.c b/src/ngx_http_lua_headers.c
index 986f06eb..6700ce80 100644
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -164,22 +164,22 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
 
     if (hc->nbusy) {
 #if nginx_version >= 1011011
-        if (hc->nbusy > lmcf->prealloc_nbusy) {
-            if (lmcf->busy_bufs_ptrs) {
-                ngx_free(lmcf->busy_bufs_ptrs);
+        if (hc->nbusy > lmcf->busy_buf_ptr_count) {
+            if (lmcf->busy_buf_ptrs) {
+                ngx_free(lmcf->busy_buf_ptrs);
             }
 
-            lmcf->busy_bufs_ptrs = ngx_alloc(hc->nbusy * sizeof(ngx_buf_t *),
-                                             r->connection->log);
+            lmcf->busy_buf_ptrs = ngx_alloc(hc->nbusy * sizeof(ngx_buf_t *),
+                                            r->connection->log);
 
-            if (lmcf->busy_bufs_ptrs == NULL) {
+            if (lmcf->busy_buf_ptrs == NULL) {
                 return luaL_error(L, "no memory");
             }
 
-            lmcf->prealloc_nbusy = hc->nbusy;
+            lmcf->busy_buf_ptr_count = hc->nbusy;
         }
 
-        bb = lmcf->busy_bufs_ptrs;
+        bb = lmcf->busy_buf_ptrs;
         for (cl = hc->busy; cl; cl = cl->next) {
             *bb++ = cl->buf;
         }
@@ -187,7 +187,7 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
         b = NULL;
 
 #if nginx_version >= 1011011
-        bb = lmcf->busy_bufs_ptrs;
+        bb = lmcf->busy_buf_ptrs;
         for (i = hc->nbusy; i > 0; i--) {
             b = bb[i - 1];
 #else
@@ -270,9 +270,9 @@ ngx_http_lua_ngx_req_raw_header(lua_State *L)
     if (hc->nbusy) {
 
 #if nginx_version >= 1011011
-        bb = lmcf->busy_bufs_ptrs;
-        for (i = hc->nbusy; i > 0; i--) {
-            b = bb[i - 1];
+        bb = lmcf->busy_buf_ptrs;
+        for (i = hc->nbusy - 1; i >= 0; i--) {
+            b = bb[i];
 #else
         for (i = 0; i < hc->nbusy; i++) {
             b = hc->busy[i];
@@ -1491,9 +1491,9 @@ ngx_http_lua_ngx_raw_header_cleanup(void *data)
 
     lmcf = (ngx_http_lua_main_conf_t *) data;
 
-    if (lmcf->busy_bufs_ptrs) {
-        ngx_free(lmcf->busy_bufs_ptrs);
-        lmcf->busy_bufs_ptrs = NULL;
+    if (lmcf->busy_buf_ptrs) {
+        ngx_free(lmcf->busy_buf_ptrs);
+        lmcf->busy_buf_ptrs = NULL;
     }
 }
 #endif

```

agentzh:
Note: nginx 1.11.11+ are still not an officially supported target of this module yet.

More work is needed. For example, I'm still seeing (at least) the following test failures when using nginx 1.11.11 though they are not related to this PR:

```
t/045-ngx-var.t ..................... 1/54
#   Failed test 'TEST 8: $proxy_host & $proxy_port - raw resp headers like'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 950.
#                   'HTTP/1.1 200 OK
# Server: nginx/1.11.11 (no pool)
# Date: Thu, 13 Apr 2017 21:59:35 GMT
# Content-Type: text/plain
# Transfer-Encoding: chunked
# Connection: close
# '
#     doesn't match '(?^s:Proxy-Host: 127.0.0.1\:\d+\r
# Proxy-Port: \d+\r
# )'

#   Failed test 'TEST 8: $proxy_host & $proxy_port - raw resp headers like'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 950.
#                   'HTTP/1.1 200 OK
# Server: nginx/1.11.11 (no pool)
# Date: Thu, 13 Apr 2017 21:59:35 GMT
# Content-Type: text/plain
# Transfer-Encoding: chunked
# Connection: close
# '
#     doesn't match '(?^s:Proxy-Host: 127.0.0.1\:\d+\r
# Proxy-Port: \d+\r
# )'
# Looks like you failed 2 tests of 54.
t/045-ngx-var.t ..................... Dubious, test returned 2 (wstat 512, 0x200)
Failed 2/54 subtests
```

defanator:
@agentzh, in regards to the `t/045-ngx-var.t`, the following patch fixes that test:

```
--- src/http/modules/ngx_http_proxy_module.c.orig	2017-04-17 21:55:16.675135473 +0000
+++ src/http/modules/ngx_http_proxy_module.c	2017-04-17 21:55:44.491772327 +0000
@@ -808,10 +808,10 @@
 static ngx_http_variable_t  ngx_http_proxy_vars[] = {
 
     { ngx_string("proxy_host"), NULL, ngx_http_proxy_host_variable, 0,
-      NGX_HTTP_VAR_CHANGEABLE|NGX_HTTP_VAR_NOCACHEABLE|NGX_HTTP_VAR_NOHASH, 0 },
+      NGX_HTTP_VAR_CHANGEABLE|NGX_HTTP_VAR_NOCACHEABLE, 0 },
 
     { ngx_string("proxy_port"), NULL, ngx_http_proxy_port_variable, 0,
-      NGX_HTTP_VAR_CHANGEABLE|NGX_HTTP_VAR_NOCACHEABLE|NGX_HTTP_VAR_NOHASH, 0 },
+      NGX_HTTP_VAR_CHANGEABLE|NGX_HTTP_VAR_NOCACHEABLE, 0 },
 
     { ngx_string("proxy_add_x_forwarded_for"), NULL,
       ngx_http_proxy_add_x_forwarded_for_variable, 0, NGX_HTTP_VAR_NOHASH, 0 },
```

(tested with 1.12.0)

You have the same one for 1.11.2:
https://github.com/openresty/openresty/blob/master/patches/nginx-1.11.2-proxy_host_port_vars.patch

agentzh:
@defanator OK, so those tests also reply on OpenResty's nginx core patches. Thanks for digging it.

defanator:
@agentzh, yes - and it's also quite unhandy to require a set of another [unrelated] modules as dependencies in order to Lua test suite to run/pass. Just thoughts. :)

agentzh:
@defanator What do you mean?

agentzh:
What unrelated modules?

msva:
> What unrelated modules?

I guess, he means that it is no need to build entire openresty bundle to just 
test this module for work

defanator:
@msva, actually, there _is_ such need, and it would be much easier if every module's test suite can be run without extra dependencies.

agentzh:
Well, we have to test interoperability between various components. Being self-contained is nice, but usually not very practical. This module's test suite is actually testing all the components getting involved. For example, this test suite used to catch a memory issue in a new nginx mainline release when in the valgrind testing mode, which was kinda tricky to reproduce.. Also, it used to catch a deep memory bug in the LuaJIT core in the valgrind testing mode in a random run (which is also hard to reproduce).

agentzh:
I don't really mind the overhead of setting up the test environment for this test suite, since it just needs to be set up only once for one developer. But the bigger chance of catching as many bugs as possible in the long run is invaluable. We have to make a good balance here.

agentzh:
Well, I believe the discussion is already OT now.

agentzh:
This module sits at the very heart of OpenResty. So it's natural for this module's test suite to exercise many OpenResty components. Better catch bugs early than sorry.

PR #1034, opened 2017-04-05
fix the return value of lpush when dict overflow

hy05190134:

### DIFF
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -5384,7 +5384,7 @@ Inserts the specified (numerical or string) <code>value</code> at the head of th
 
 If <code>key</code> does not exist, it is created as an empty list before performing the push operation. When the <code>key</code> already takes a value that is not a list, it will return <code>nil</code> and <code>"value not a list"</code>.
 
-It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>nil</code> and the string "no memory".
+It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>false</code> and the string "no memory".
 
 This feature was first introduced in the <code>v0.10.6</code> release.
 
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -5384,7 +5384,7 @@ Inserts the specified (numerical or string) <code>value</code> at the head of th
 
 If <code>key</code> does not exist, it is created as an empty list before performing the push operation. When the <code>key</code> already takes a value that is not a list, it will return <code>nil</code> and <code>"value not a list"</code>.
 
-It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>false</code> and the string "no memory".
+It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>a false value</code> and the string "no memory".
 
 This feature was first introduced in the <code>v0.10.6</code> release.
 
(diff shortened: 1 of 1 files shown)


agentzh:
@hy05190134 I think it should return `nil` in this case. Will you prepare a *standalone* and *minimal* example that can demonstrate this behavior?

tokers:
@agentzh  The function `ngx_http_lua_shdict_push_helper` (ngx_http_lua_shdict.c) really push `false` and a literal `no memory` to stack, when attempting to alloc a new node from the slab failed.

hy05190134:
@agentzh  I just set shared dict for 1k, and I use lpush duatively to fill the shared dict, when it failed, it just  return false and "no memory".

agentzh:
@hy05190134 I was expecting a runnable lua/nginx.conf snippet ;)

@doujiang24 Mind to take care of this?

doujiang24:
@agentzh actually we return `false` instead of `nil` now.
And I checked the `set`, `safe_set` operations too, they also return `false`.
Maybe we should fix the document?

agentzh:
@doujiang24 OK, we just need to be consistent here.

hy05190134:
@agentzh  so not only lpush should be corrected

agentzh:
@doujiang24 I still think we should always return `nil`, `"no memory"` in such cases. This aligns up with the rest of Lua API functions.

doc/HttpLuaModule.wiki
```diff
@@ -5384,7 +5384,7 @@ Inserts the specified (numerical or string) <code>value</code> at the head of th
 
 If <code>key</code> does not exist, it is created as an empty list before performing the push operation. When the <code>key</code> already takes a value that is not a list, it will return <code>nil</code> and <code>"value not a list"</code>.
 
-It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>nil</code> and the string "no memory".
+It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>false</code> and the string "no memory".
```
agentzh: I think the document should just say `return a false value`. Both `nil` and `false` are considered a "false value" when evaluated in the boolean context. So either is fine.

The caller should always use `if not rc then` or `if rc then` to test the first return value, instead of the too specific form, `if rc ~= true` or `rc ~= nil`. The latter is wrong.

hy05190134: right, I just use if rc == nil before I find the return value is wrong, even thought, the document should be modified. I just modify the document

doc/HttpLuaModule.wiki
```diff
@@ -5384,7 +5384,7 @@ Inserts the specified (numerical or string) <code>value</code> at the head of th
 
 If <code>key</code> does not exist, it is created as an empty list before performing the push operation. When the <code>key</code> already takes a value that is not a list, it will return <code>nil</code> and <code>"value not a list"</code>.
 
-It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>nil</code> and the string "no memory".
+It never overrides the (least recently used) unexpired items in the store when running out of storage in the shared memory zone. In this case, it will immediately return <code>a false value</code> and the string "no memory".
```
agentzh: I think we should remove the `<code>` and `</code>` tags here.

agentzh:
@doujiang24 I thought a bit more about this. `lpush` returns a `length` instead of a true-false value. So it must return `nil` on errors, otherwise it just does not make sense. It's unlike `set` and et c, which return a `success` value which is a true-false value.

doujiang24:
@agentzh Yeah, agreed, we should return `nil` for `lpush` and `rpush` here.
@hy05190134 Could you do me a favor? Thanks :)

hy05190134:
@doujiang24  sure, I try to modify the code.

PR #1037, opened 2017-04-07
feature: supported collect 'error.log' without reading the nginx error log files.

membphis:

mini use case:

```
    lua_intercept_log 4; # maximum number of error log rows

    location /t {
        access_by_lua_block {
            ngx.log(ngx.ERR, "enter 1")
            ngx.log(ngx.ERR, "enter 2")
            ngx.say("hello")
        }
        log_by_lua_block {
            -- fetch error.log, we can reforward it to log server
            local t = ngx.req.get_intercept_log()
            ngx.log(ngx.ERR, "intercept log line:", t)
        }
    }
```

todo: 

* add API document.
* add a control flag: distinguish copies or intercept logs.

this PR depends on nginx core patch: https://github.com/membphis/openresty/tree/intercept_log_patch/patches, I'll submit later.


src/ngx_http_lua_directive.h
```diff
@@ -71,6 +71,8 @@ char *ngx_http_lua_rewrite_no_postpone(ngx_conf_t *cf, ngx_command_t *cmd,
 char *ngx_http_lua_conf_lua_block_parse(ngx_conf_t *cf,
     ngx_command_t *cmd);
 
+char *ngx_http_lua_intercept_log(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf);
```
agentzh: Style: need 2 blank lines here.

src/ngx_http_lua_util.h
```diff
@@ -285,10 +288,10 @@ ngx_http_lua_create_ctx(ngx_http_request_t *r)
     ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (!llcf->enable_code_cache && r->connection->fd != (ngx_socket_t) -1) {
-        lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
+    lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
+    dd("lmcf: %p", lmcf);
 
-        dd("lmcf: %p", lmcf);
+    if (!llcf->enable_code_cache && r->connection->fd != (ngx_socket_t) -1) {
```
agentzh: Why this change?

t/151-intercept-log.t
```diff
@@ -0,0 +1,195 @@
...
+=== TEST 1: sanity
+--- http_config
+    lua_intercept_log 4m;
+--- config
+    location /t {
+        access_by_lua_block {
+            ngx.log(ngx.ERR, "enter 1")
+            ngx.log(ngx.ERR, "enter 11")
+
+            local t = ngx.get_log()
```
agentzh: We should only test the FFI-based Lua API provided by `lua-resty-core`. We no longer want to add more CFunction-based Lua API to the global `ngx` table, which can never scale. Everytime we add a new CFunction entry to the `ngx.` global table, the more GC overhead we add at the same time.

src/ngx_http_lua_module.c
```diff
@@ -93,6 +93,13 @@ static ngx_command_t ngx_http_lua_cmds[] = {
       0,
       NULL },
 
+    { ngx_string("lua_intercept_log"),
```
agentzh: I'm afraid this is confusing. NGINX already has two different kinds of logs: access logs and error logs. We should be more explicit here to avoid confusions.

src/ngx_http_lua_module.c
```diff
@@ -748,6 +755,11 @@ ngx_http_lua_init(ngx_conf_t *cf)
         dd("Lua VM initialized!");
     }
 
+    if (lmcf->requires_intercept_log) {
+        cf->cycle->log_intercept_handler = (ngx_log_intercept_pt)
```
agentzh: BTW, we should protect such code with our own C macros so that this nginx module can still be compiled with the standard nginx core. And when working with the stock nginx cores, we should emit an error when the user configures this config directive that requires our nginx core patch.

src/ngx_http_lua_util.c
```diff
@@ -4093,4 +4093,33 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
...
+    }
+
+    intercept_log = &lmcf->intercept_log;
+
+    if (n > intercept_log->max_size - intercept_log->size) {
+        return NGX_DECLINED;
+    }
+
+    ngx_memcpy(intercept_log->data + intercept_log->size, buf, n);
+    intercept_log->size += n;
```
agentzh: Here we lose the valuable info about the boundaries of individual error log messages, which is not desired.

src/ngx_http_lua_common.h
```diff
@@ -147,6 +149,13 @@ typedef struct {
 } ngx_http_lua_preload_hook_t;
 
 
+typedef struct {
+    ssize_t              max_size;
+    size_t               size;
+    u_char              *data;
```
agentzh: Instead of a flat static buffer, I think we should use a ring buffer that can

1. distinguish between error log message boundaries so that we don't have to split it later ourselves (some error log messages do span multiple lines, like those with a Lua backtrace, which makes late splitting nontrivial), and
2. when the buffer is full, the oldest log messages are automatically discarded and overwritten. A ring buffer design can avoid moving existing log data inside the buffer for this feature.

src/ngx_http_lua_util.c
```diff
@@ -4093,4 +4093,33 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
...
+
+    lmcf = ngx_http_cycle_get_module_main_conf(ngx_cycle, ngx_http_lua_module);
+    if (lmcf == NULL) {
+        return NGX_DECLINED;
+    }
+
+    intercept_log = &lmcf->intercept_log;
+
+    if (n > intercept_log->max_size - intercept_log->size) {
+        return NGX_DECLINED;
```
agentzh: Instead of discarding the latest error log messages, we should discard the oldest ones in the buffer instead. New error log messages are almost always more valuable than the oldest ones. We need to make the buffer still get updated with the latest error logs even when the Lua application fails to extract the logs in time.

t/151-intercept-log.t
```diff
@@ -0,0 +1,195 @@
...
+--- request
+GET /t
+--- response_body
+log length:341
+--- grep_error_log eval
+qr/enter \d+/
+--- grep_error_log_out eval
+[
+"enter 1
+enter 11
```
agentzh: The output misses many important info like log level, timestamps, connection ID, request ID, request context info. For example, a complete error log message should look like this:

```
2017/04/11 12:29:48 [error] 51725#0: *14 connect() failed (111: Connection refused) while connecting to upstream, client: 127.0.0.1, server: localhost, request: "GET /main HTTP/1.1", subrequest: "/memc", upstream: "memcached://127.0.0.1:11211", host: "localhost"
```

t/151-intercept-log.t
agentzh: Also, I think the Lua API should return individual error log messages as separate strings instead of directly returning a big bulk that requires manual splitting.

src/ngx_http_lua_log.c
```diff
@@ -275,6 +276,9 @@ ngx_http_lua_inject_log_api(lua_State *L)
     lua_pushcfunction(L, ngx_http_lua_ngx_log);
     lua_setfield(L, -2, "log");
 
+    lua_pushcfunction(L, ngx_http_lua_ngx_get_log);
```
membphis: I'm sorry, I missed the news, I was ready to submit this evening, but I did not finish it. I will submit a new version tomorrow morning.

src/ngx_http_lua_ringbuff_log.h
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (C) OpenResty, Inc.
+ */
+
+
+#ifndef _NGX_HTTP_LUA_LOG_RINGBUFF_H_INCLUDED_
+#define _NGX_HTTP_LUA_LOG_RINGBUFF_H_INCLUDED_
+
+
+#include <nginx.h>
+#include <ngx_core.h>
```
agentzh: I think it's better to include `ngx_http_lua_common.h` instead.

src/ngx_http_lua_directive.c
```diff
@@ -27,6 +27,7 @@
 #include "ngx_http_lua_ssl_certby.h"
 #include "ngx_http_lua_lex.h"
 #include "api/ngx_http_lua_api.h"
+#include "ngx_http_lua_ringbuff_log.h"
```
agentzh: I think the file name is better to be `ngx_http_lua_log_ringbuf.h`.

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,107 @@ ngx_http_lua_inject_log_consts(lua_State *L)
     /* }}} */
 }
 
+
+#ifndef NGX_LUA_NO_FFI_API
+int
+ngx_http_lua_ffi_filter_log(int level, u_char *err, size_t *errlen)
```
agentzh: I think `ngx_http_lua_ffi_errlog_set_filter_level` is a better name.

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,107 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+    return NGX_ERROR;
+#endif
+}
+
+
+int
+ngx_http_lua_ffi_errlog_count(u_char *err, size_t *errlen)
+{
+#ifdef HAVE_INTERCEPT_ERROR_LOG_PATCH
+    ngx_http_lua_log_ringbuff_t     *log_ringbuff;
```
agentzh: We should use `ringbuf` instead of `ringbuff` in identifiers. Please fix other similar places.

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,107 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+
+    log_ringbuff = ngx_cycle->intercept_error_log_data;
+
+    if (!log_ringbuff) {
```
agentzh: We should use ` == NULL` instead for pointers' nullity test. Note that it's different from testing a boolean value.

src/ngx_http_lua_ringbuff_log.c
```diff
@@ -0,0 +1,182 @@
...
+
+#include <nginx.h>
+#include "ngx_http_lua_ringbuff_log.h"
+
+
+typedef struct {
+    int    is_data;
+    int    log_level;             /* attatch info */
+    size_t len;                   /* data length */
+}ringbuff_head;
```
agentzh: Also, I think `ngx_http_lua_log_ringbuf_header_t` is a better name. Please note that `head` and `header` are quite different things.

src/ngx_http_lua_directive.c
```diff
@@ -1700,6 +1701,77 @@ ngx_http_lua_conf_read_lua_token(ngx_conf_t *cf,
...
+{
+#ifndef HAVE_INTERCEPT_ERROR_LOG_PATCH
+    return "missing intercept error log patch";
+#else
+    ngx_str_t                     *value;
+    ssize_t                        size;
+    u_char                        *data;
+    ngx_cycle_t                   *cycle;
+    ngx_http_lua_main_conf_t      *lmcf = conf;
+    ngx_http_lua_log_ringbuff_t   *log_ringbuff;
```
agentzh: I think this variable should be renamed to just `ringbuf` instead.

src/ngx_http_lua_ringbuff_log.c
```diff
@@ -0,0 +1,182 @@
+
+/*
+ * Copyright (C) OpenResty, Inc.
```
agentzh: No company copyright notice in this open source project.

membphis:
@agentzh @dndx please take a review when you have time :)

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,113 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+                           "missing intercept error log patch in the nginx "
+                           "core")
+              - err;
+    return NGX_ERROR;
+#endif
+}
+
+
+int
+ngx_http_lua_ffi_errlog(ngx_http_lua_ffi_table_elt_t *out, size_t max,
```
agentzh: Better use verbs in function names. For example, `ngx_http_lua_ffi_get_errlog_data`.

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,113 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+                           "missing intercept error log patch in the nginx "
+                           "core")
+              - err;
+    return NGX_ERROR;
+#endif
+}
+
+
+int
+ngx_http_lua_ffi_errlog_count(u_char *err, size_t *errlen)
```
agentzh: Better use verbs in function names. For example, `ngx_http_lua_ffi_get_errlog_count`.

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,113 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+    }
+
+    for (i = 0; i < count; i++) {
+        ngx_http_lua_log_ringbuf_read(ringbuf, &log_level, &data, &len);
+        out[i].key.len = log_level;
+        out[i].value.len = len;
+        out[i].value.data = data;
+    }
+
+    return NGX_OK;
```
agentzh: I think it's better for this function to return the `count` directly so that the caller does not have to call `ngx_http_lua_ffi_get_errlog_count` first. This can be more efficient.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,185 @@
...
+        rb->head = rb->data;
+        return rb->head;
+    }
+
+    return rb->head;
+}
+
+
+static void
+ngx_http_lua_log_ringbuf_append_tail(ngx_http_lua_log_ringbuf_t *rb,
```
agentzh: The word "append" already means adding at the tail. Adding things to the front is "prepend".

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,185 @@
...
+    if (rb->tail >= rb->head) {
+        return rb->data + rb->size - rb->tail;
+    }
+
+    return rb->head - rb->tail;
+}
+
+
+ngx_int_t
+ngx_http_lua_log_ringbuf_write(ngx_http_lua_log_ringbuf_t *rb, int log_level,
```
agentzh: I suggest you add simple comment explaining each of these functions at the beginning of each function. For example, it's not immediately obvious how this function is different from your `ngx_http_lua_log_ringbuf_append_tail` function above. We need more code comments.

src/ngx_http_lua_log.c
membphis: @agentzh  do you mean we check the variable `out`? if it's a `NULL` value, we return `count` directly and do nothing?

call `ngx_http_lua_ffi_get_errlog_data` twice?

membphis:
@agentzh submitted new version, please take a review :)

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,78 @@ ngx_http_lua_inject_log_consts(lua_State *L)
     /* }}} */
 }
 
+
+#ifndef NGX_LUA_NO_FFI_API
+int
+ngx_http_lua_ffi_set_errlog_filter(int level, u_char *err,
+    size_t *errlen)
```
agentzh: Seems like we do not really need a line break here?

src/ngx_http_lua_log.c
```diff
@@ -313,4 +314,78 @@ ngx_http_lua_inject_log_consts(lua_State *L)
...
+        *errlen = ngx_snprintf(err, *errlen,
+                               "API \"get_errlog_data\" depends on directive "
+                               "\"lua_intercept_error_log\"")
+                  - err;
+        return NGX_ERROR;
+    }
+
+    count = ringbuf->count;
+    if (count == 0) {
+        return count;
```
agentzh: Seems like we can just return the message length directly here? This way we can save the `size _t *loglen` parameter. Because the message length can never exceed 4K, for example, the positive part of an `int` should be more than enough.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,206 @@
...
+
+
+#include "ngx_http_lua_common.h"
+#include "ngx_http_lua_log_ringbuf.h"
+
+
+typedef struct {
+    unsigned    is_data:1;
+    unsigned    log_level:4;
+    int         len;
```
agentzh: Beware of padding, this layout is very inefficient. Please always put larger struct fields before smaller ones. Here you are doing the opposite. Alas.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,206 @@
...
+
+ngx_int_t
+ngx_http_lua_log_ringbuf_write(ngx_http_lua_log_ringbuf_t *rb, int log_level,
+    void *buf, size_t n)
+{
+    size_t              free_spaces, head_len;
+
+    ngx_http_lua_log_ringbuf_header_t       *head;
+
+    head_len = sizeof(ngx_http_lua_log_ringbuf_header_t);
```
agentzh: We shouldn't use an automatic variable for constant values like this. It is best practice to use anonymous enum for this:

```C
enum {
    HEADER_LEN = sizeof(ngx_http_lua_log_ringbuf_header_t),
};
```

Also, we should use "header len" instead of "head len". The latter is confusing.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    ngx_http_lua_log_ringbuf_append(rb, 1, log_level, buf, n);
+
+    return NGX_OK;
+}
+
+
+/* read log from ring buffer, do reset if all of the logs were readed. */
+
```
agentzh: style: no blank line here. Please fix other similar places accordingly. Thank you.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,206 @@
...
+    if (rb->size - (rb->head - rb->data) <
+        sizeof(ngx_http_lua_log_ringbuf_header_t))
+    {
+        rb->head = rb->data;
+        return rb->head;
+    }
+
+    /* placehold data */
+    head = (ngx_http_lua_log_ringbuf_header_t *) rb->head;
+    if (!head->is_data) {
```
agentzh: The `is_data` field is a bit wasteful and confusing. How about introducing a design similar to the "sentinel" thing in nginx's `ngx_rbtree_t` and `ngx_queue_t` data structures?

membphis:
```c
typedef struct {
    unsigned    is_data:1;
    unsigned    log_level:4;
    unsigned    len:24;
} ngx_http_lua_log_ringbuf_header_t;
```

changed to this version.

the length of `ngx_http_lua_log_ringbuf_header_t` is 4bytes now, the "sentinel" of `ngx_rbtree_t` was a point and it's 8bytes.

agentzh: @membphis Please always put larger fields *before* smaller fields. I've seen this mistake several times. Be very careful about address alignment and padding.

agentzh:
@membphis Will you please rebase to the latest master?

membphis:
@agentzh rebased ^_^

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,203 @@
...
+
+
+#include "ngx_http_lua_common.h"
+#include "ngx_http_lua_log_ringbuf.h"
+
+
+typedef struct {
+    unsigned    len:24;
+    unsigned    log_level:4;
+    unsigned    is_data:1;
```
agentzh: Should we use a similar concept of `sentinel` to `ngx_queue_t` and `ngx_rbtree_t` in the nginx core? Use of the `is_data` flag in every node might not be so efficient. I posted this comment in my previous rounds of review but it seems it got lost.

membphis:
@agentzh submitted new version https://github.com/openresty/lua-nginx-module/pull/1037/commits/928ab280b7dc2c03e820f730e12e0b3465cb2f6c fixed all points.

src/ngx_http_lua_log_ringbuf.c
membphis: I can't sure the new version is right, please take a view.

src/ngx_http_lua_util.h
```diff
@@ -285,10 +288,10 @@ ngx_http_lua_create_ctx(ngx_http_request_t *r)
     ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
 
     llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
-    if (!llcf->enable_code_cache && r->connection->fd != (ngx_socket_t) -1) {
-        lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
+    lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);
```
doujiang24: I'm sorry, this is useless now. I wrote the comment long ago, and I didn't realize it just now.

src/ngx_http_lua_misc.c
```diff
@@ -58,6 +62,40 @@ ngx_http_lua_ngx_req_is_internal(lua_State *L)
 
 
 static int
+ngx_http_lua_ngx_req_get_intercept_log(lua_State *L)
```
doujiang24: I'm sorry, this is useless now. I wrote the comment long ago, and I didn't realize it just now.

src/ngx_http_lua_util.h
membphis: I think the version you viewed is old.

src/ngx_http_lua_log_ringbuf.h
```diff
@@ -0,0 +1,30 @@
...
+
+
+typedef struct {
+    ngx_uint_t filter_level;
+    char      *tail;              /*  writed point      */
+    char      *head;              /*  readed point      */
+    char      *data;              /*  buffer            */
+    size_t     size;              /*  buffer total size */
+    size_t     count;             /*  count of logs     */
+}ngx_http_lua_log_ringbuf_t;
```
membphis: good cache.

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,201 @@
...
+#include "ddebug.h"
+
+
+#include "ngx_http_lua_common.h"
+#include "ngx_http_lua_log_ringbuf.h"
+
+
+typedef struct {
+    unsigned    len:24;
+    unsigned    log_level:4;
```
agentzh: @membphis Maybe we should make it fill 32-bit instead of 28-bit here? This struct will be 32-bit due to padding requirements anyway.

membphis: fixed https://github.com/openresty/lua-nginx-module/pull/1037/commits/cc2d1970f5f38b9afa7e611e5089ced5b66387b4

src/ngx_http_lua_log_ringbuf.c
```diff
@@ -0,0 +1,201 @@
...
+} ngx_http_lua_log_ringbuf_header_t;
+
+
+enum {
+    HEADER_LEN = sizeof(ngx_http_lua_log_ringbuf_header_t)
+};
+
+
+static void * ngx_http_lua_log_ringbuf_next_header(
```
agentzh: We should remove the space after `*`. This issue was found by the `ngx-releng` tool, BTW.

src/ngx_http_lua_directive.c
```diff
@@ -1700,6 +1701,71 @@ ngx_http_lua_conf_read_lua_token(ngx_conf_t *cf,
 }
 
 
+char *
+ngx_http_lua_intercept_error_log(ngx_conf_t *cf, ngx_command_t *cmd,
```
agentzh: I think this function should go to the `ngx_http_lua_log.c` file instead. This `util.c` file is already too big. Better not make it grow.

Also, I don't think this function should even exist when `HAVE_INTERCEPT_ERROR_LOG_PATCH` is not defined.

agentzh:
@membphis Merged with the following extra patch:

```diff
diff --git a/.travis.yml b/.travis.yml
index 1720e1c4..359c8cc6 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,9 +66,9 @@ install:
   - if [ ! -f download-cache/pcre-$PCRE_VER.tar.gz ]; then wget -P download-cache http://ftp.cs.stanford.edu/pub/exim/pcre/pcre-$PCRE_VER.tar.gz; fi
   - if [ ! -f download-cache/openssl-$OPENSSL_VER.tar.gz ]; then wget -P download-cache https://www.openssl.org/source/openssl-$OPENSSL_VER.tar.gz; fi
   - git clone https://github.com/openresty/test-nginx.git
-  - git clone -b intercept_log_patch https://github.com/membphis/openresty.git ../openresty
+  - git clone https://github.com/openresty/openresty.git ../openresty
   - git clone https://github.com/openresty/no-pool-nginx.git ../no-pool-nginx
-  - git clone -b intercept_log https://github.com/membphis/openresty-devel-utils.git
+  - git clone https://github.com/openresty/openresty-devel-utils.git
   - git clone https://github.com/openresty/mockeagain.git
   - git clone https://github.com/openresty/lua-cjson.git
   - git clone https://github.com/openresty/lua-upstream-nginx-module.git ../lua-upstream-nginx-module
diff --git a/src/ngx_http_lua_directive.c b/src/ngx_http_lua_directive.c
index ef6f70dc..a92f056d 100644
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -28,6 +28,7 @@
 #include "ngx_http_lua_lex.h"
 #include "api/ngx_http_lua_api.h"
 #include "ngx_http_lua_log_ringbuf.h"
+#include "ngx_http_lua_log.h"
 
 
 typedef struct ngx_http_lua_block_parser_ctx_s
@@ -1706,7 +1707,7 @@ ngx_http_lua_intercept_error_log(ngx_conf_t *cf, ngx_command_t *cmd,
     void *conf)
 {
 #ifndef HAVE_INTERCEPT_ERROR_LOG_PATCH
-    return "missing intercept error log patch in the nginx core";
+    return "not found: missing the intercept error log patch for nginx";
 #else
     ngx_str_t                     *value;
     ssize_t                        size;
diff --git a/src/ngx_http_lua_log.c b/src/ngx_http_lua_log.c
index 4b13b8d5..9e87a590 100644
--- a/src/ngx_http_lua_log.c
+++ b/src/ngx_http_lua_log.c
@@ -315,6 +315,31 @@ ngx_http_lua_inject_log_consts(lua_State *L)
 }
 
 
+#ifdef HAVE_INTERCEPT_ERROR_LOG_PATCH
+ngx_int_t
+ngx_http_lua_intercept_log_handler(ngx_log_t *log,
+    ngx_uint_t level, void *buf, size_t n)
+{
+    ngx_http_lua_log_ringbuf_t  *ringbuf;
+
+    dd("enter");
+
+    ringbuf = (ngx_http_lua_log_ringbuf_t  *)
+                    ngx_cycle->intercept_error_log_data;
+
+    if (level > ringbuf->filter_level) {
+        return NGX_OK;
+    }
+
+    ngx_http_lua_log_ringbuf_write(ringbuf, level, buf, n);
+
+    dd("intercept log: %s\n", buf);
+
+    return NGX_OK;
+}
+#endif
+
+
 #ifndef NGX_LUA_NO_FFI_API
 int
 ngx_http_lua_ffi_set_errlog_filter(int level, u_char *err, size_t *errlen)
@@ -342,8 +367,7 @@ ngx_http_lua_ffi_set_errlog_filter(int level, u_char *err, size_t *errlen)
     return NGX_OK;
 #else
     *errlen = ngx_snprintf(err, *errlen,
-                           "missing intercept error log patch in the nginx "
-                           "core")
+                           "missing the intercept error log patch for nginx")
               - err;
     return NGX_ERROR;
 #endif
@@ -377,8 +401,7 @@ ngx_http_lua_ffi_get_errlog_data(char **log, int *loglevel, u_char *err,
     return loglen;
 #else
     *errlen = ngx_snprintf(err, *errlen,
-                           "missing intercept error log patch in the nginx "
-                           "core")
+                           "missing the intercept error log patch for nginx")
               - err;
     return NGX_ERROR;
 #endif
diff --git a/src/ngx_http_lua_log.h b/src/ngx_http_lua_log.h
index 42f18394..115679af 100644
--- a/src/ngx_http_lua_log.h
+++ b/src/ngx_http_lua_log.h
@@ -13,6 +13,10 @@
 
 
 void ngx_http_lua_inject_log_api(lua_State *L);
+#ifdef HAVE_INTERCEPT_ERROR_LOG_PATCH
+ngx_int_t ngx_http_lua_intercept_log_handler(ngx_log_t *log,
+    ngx_uint_t level, void *buf, size_t n);
+#endif
 
 
 #endif /* _NGX_HTTP_LUA_LOG_H_INCLUDED_ */
diff --git a/src/ngx_http_lua_util.c b/src/ngx_http_lua_util.c
index 8e7c63f1..ba9c9579 100644
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -4094,29 +4094,4 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
 }
 
 
-ngx_int_t
-ngx_http_lua_intercept_log_handler(ngx_log_t *log,
-    ngx_uint_t level, void *buf, size_t n)
-{
-#ifdef HAVE_INTERCEPT_ERROR_LOG_PATCH
-    ngx_http_lua_log_ringbuf_t  *ringbuf;
-
-    dd("enter");
-
-    ringbuf = (ngx_http_lua_log_ringbuf_t  *)
-                    ngx_cycle->intercept_error_log_data;
-
-    if (level > ringbuf->filter_level) {
-        return NGX_OK;
-    }
-
-    ngx_http_lua_log_ringbuf_write(ringbuf, level, buf, n);
-
-    dd("intercept log: %s\n", buf);
-#endif
-
-    return NGX_OK;
-}
-
-
 /* vi:set ft=c ts=4 sw=4 et fdm=marker: */
diff --git a/src/ngx_http_lua_util.h b/src/ngx_http_lua_util.h
index 82fa8875..a37852f2 100644
--- a/src/ngx_http_lua_util.h
+++ b/src/ngx_http_lua_util.h
@@ -249,9 +249,6 @@ ngx_http_cleanup_t *ngx_http_lua_cleanup_add(ngx_http_request_t *r,
 void ngx_http_lua_cleanup_free(ngx_http_request_t *r,
     ngx_http_cleanup_pt *cleanup);
 
-ngx_int_t ngx_http_lua_intercept_log_handler(ngx_log_t *log,
-    ngx_uint_t level, void *buf, size_t n);
-
 
 #define ngx_http_lua_check_if_abortable(L, ctx)                             \
     if ((ctx)->no_abort) {                                                  \
```

PR #782, opened 2016-05-31
Add jit_stack_size option

alubbe:
This is a quick suggestion on how a jit_stack_size option could be implemented. I lack the actual C skills to pull this off, and I would first like to discuss the direction of this approach.

Allocating and freeing jit_stacks for each regex is too costly according to http://www.pcre.org/original/doc/html/pcrejit.html#SEC9, and it seems safe to reuse the same stack if the code runs single-threaded according to that same FAQ.

My idea would thus be to have a "global" jit_stack, initially set to 32K (the current default), and to have a lua-land function akin to this

``` lua
function ngx.re.opt(option, value)
  if option == "jit_stack_size" then
    ngx_http_lua_ffi_set_jit_stack_size(value)
  end
end
```

``` c
void 
ngx_http_lua_ffi_set_jit_stack_size(int size)
{
  pcre_jit_stack_free(jit_stack);
  int min_size = MIN(32 * 1024, size);
  jit_stack = pcre_jit_stack_alloc(min_size, size);
}
```

Every time a regex is compiled in ngx_http_lua_regex.c, we just add

``` c
pcre_assign_jit_stack(sd, NULL, jit_stack);
```

What remains to be discussed is where `jit_stack` lives. It could be a mutable C global, it could be part of the Lua state or we could instantiate it on every compile, which would be more expensive, but assuming that most users combine "j" with "o", it might not make a dent after all.

Feedback is most welcome.

PR #1041, opened 2017-04-14
feature: Added ngx_http_lua_ffi_ssl_ocsp_get_nextupdate

alubbe:
This PR is a possible solution to https://github.com/openresty/lua-resty-core/issues/75 and is complemented by https://github.com/openresty/lua-resty-core/pull/99.
Looking forward to your feedback.

### DIFF
--- a/src/ngx_http_lua_ssl_ocsp.c
+++ b/src/ngx_http_lua_ssl_ocsp.c
@@ -434,7 +434,7 @@ error:
 
 
 static time_t
-ngx_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time)
+ngx_http_lua_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time)
 {
     u_char  *value;
     size_t   len;
@@ -558,7 +558,7 @@ ngx_http_lua_ffi_ssl_ocsp_get_nextupdate(const u_char *resp, size_t resp_len,
         goto error;
     }
 
-    *nextupdate = ngx_ssl_stapling_time(nupdate_ans1);
+    *nextupdate = ngx_http_lua_ssl_stapling_time(nupdate_ans1);
     if (*nextupdate == (time_t) NGX_ERROR) {
         *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
                                     "invalid nextUpdate time in "
--- a/src/ngx_http_lua_ssl_ocsp.c
+++ b/src/ngx_http_lua_ssl_ocsp.c
@@ -433,6 +433,182 @@ error:
 }
 
 
+static time_t
+ngx_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time)
+{
+    u_char  *value;
+    size_t   len;
+    time_t   time;
+    BIO     *bio;
+
+    /*
+     * OpenSSL doesn't provide a way to convert ASN1_GENERALIZEDTIME
+     * into time_t.  To do this, we use ASN1_GENERALIZEDTIME_print(),
+     * which uses the "MMM DD HH:MM:SS YYYY [GMT]" format (e.g.,
+     * "Feb  3 00:55:52 2015 GMT"), and parse the result.
+     */
+
+    bio = BIO_new(BIO_s_mem());
+    if (bio == NULL) {
+        return NGX_ERROR;
+    }
+
+    /* fake weekday prepended to match C asctime() format */
+
+    BIO_write(bio, "Tue ", sizeof("Tue ") - 1);
+    ASN1_GENERALIZEDTIME_print(bio, asn1time);
+    len = BIO_get_mem_data(bio, &value);
+
+    time = ngx_parse_http_time(value, len);
+
+    BIO_free(bio);
+
+    return time;
+}
+
+
+int
+ngx_http_lua_ffi_ssl_ocsp_get_nextupdate(const u_char *resp, size_t resp_len,
+    const char *chain_data, size_t chain_len, time_t* nextupdate,
+    u_char *errbuf, size_t *errbuf_size)
+{
+#ifndef NGX_HTTP_LUA_USE_OCSP
+
+    *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
+                                "no OCSP support") - errbuf;
+    return NGX_ERROR;
+
+#else
+
+    BIO                   *bio = NULL;
+    X509                  *cert = NULL, *issuer = NULL;
+    OCSP_CERTID           *id = NULL;
+    OCSP_RESPONSE         *ocsp = NULL;
+    OCSP_BASICRESP        *basic = NULL;
+    ASN1_GENERALIZEDTIME  *nupdate_ans1;
+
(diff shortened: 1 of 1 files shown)


alubbe:
Also, two newbie questions:
1) I copied `ngx_ssl_stapling_time` over from the nginx source code because I couldn't figure out how to reference/reuse/import it. Is there a way to do this without patching nginx?
2) Should `ngx_http_lua_ffi_ssl_ocsp_get_nextupdate` be in a header file somewhere?

agentzh:
@alubbe To answer your questions:

1. Static C functions cannot be reused outside that compilation unit.
2. No. Those C functions are not meant to be used by other C code.

agentzh:
@ghedo @lziest Do you mind reviewing this PR for me? Many thanks!

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
 }
 
 
+static time_t
+ngx_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time)
```
agentzh: You should use the `ngx_http_lua_` prefix, no matter what.

alubbe:
> Static C functions cannot be reused outside that compilation unit.

Fair enough. I will copy the contents and rename it to `ngx_http_lua_ssl_stapling_time`.

> No. Those C functions are not meant to be used by other C code.

Can we also remove https://github.com/openresty/lua-nginx-module/pull/782/files#diff-c4f8697714ffd4ae8e3d403bc9a55715R17 ?

agentzh:
@alubbe I can easily reproduced those test failures in this PR's Travis CI report.

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
...
+    }
+
+    issuer = d2i_X509_bio(bio, NULL);
+    if (issuer == NULL) {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
+                                    "d2i_X509_bio() failed") - errbuf;
+        goto error;
+    }
+
+    id = OCSP_cert_to_id(NULL, cert, issuer);
```
lziest: hmm, I am not sure what this cert_to_id do though.

alubbe: It extracts the `id` that I believe is mandatory to call `OCSP_resp_find_status`, which in turn extracts the nextUpdate value.
https://www.openssl.org/docs/man1.1.0/crypto/OCSP_cert_to_id.html
https://www.openssl.org/docs/man1.1.0/crypto/OCSP_resp_find_status.html

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
...
+
+    bio = BIO_new_mem_buf((char *) chain_data, chain_len);
+    if (bio == NULL) {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
+                                    "BIO_new_mem_buf() failed")
+                       - errbuf;
+        goto error;
+    }
+
+    cert = d2i_X509_bio(bio, NULL);
```
ghedo: We already have FFI functions to parse certificates into `STACK_OF(X509)` (from PEM, but still), so IMO this function should just take an already parsed certificate chain instead of a DER blob.

src/ngx_http_lua_ssl_ocsp.c
ghedo: Is `issuer` supposed to be the issuer of the leaf certificate, or the issuer of the OCSP response? (they don't always match).

If I understand the documentation and the OpenSSL code correctly it is the former (issuer of certificate) so this implementation is correct, but I'm not 100% sure.

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
...
+        != 1)
+    {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
+                                    "certificate status not found in the "
+                                    "OCSP response") - errbuf;
+        goto error;
+    }
+
+    if (nupdate_ans1 == NULL)
+    {
```
ghedo: Should not be on its own line.

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
...
+    id = OCSP_cert_to_id(NULL, cert, issuer);
+    if (id == NULL) {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size,
+                                    "OCSP_cert_to_id() failed") - errbuf;
+        goto error;
+    }
+
+    if (OCSP_resp_find_status(basic, id, NULL, NULL, NULL, NULL, &nupdate_ans1)
+        != 1)
+    {
```
ghedo: Same.

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -433,6 +433,182 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
...
+
+    BIO_free(bio);
+
+    return time;
+}
+
+
+int
+ngx_http_lua_ffi_ssl_ocsp_get_nextupdate(const u_char *resp, size_t resp_len,
+    const char *chain_data, size_t chain_len, time_t* nextupdate,
```
ghedo: Should be `time_t *nextupdate` not `time_t* nextupdate`.

ghedo: Also, `next_update` might be better.

src/ngx_http_lua_ssl_ocsp.c
alubbe: This function body borrows heavily from the function right above it. This particular line is the same as https://github.com/openresty/lua-nginx-module/pull/1041/files#diff-50267b7dd63c740bc5c1d29c7387e789R357

src/ngx_http_lua_ssl_ocsp.c
alubbe: I agree that `next_update` is better, since the official field name is `nextUpdate`. However, that naming decision was already made in existing code: https://github.com/openresty/lua-nginx-module/pull/1041/files#diff-50267b7dd63c740bc5c1d29c7387e789L284
If @agentzh would allow me, I could open a PR renaming all existing instances to `next_update`, but until then I'd stick to the existing convention of `nextupdate`.

zhuizhuhaomeng:
This feature has been implemented in another PR.
See lua-resty-core git commit 7a6521deb7a354fac9cb24cb477c144559ec3c06 for more detail.

PR #1050, opened 2017-04-24
add delay_list to tag delay init_by_lua or not

rainingmaster:
Some time we need to delayed execution of init_by_lua. For example, a third party module using `ngx_shm_zone_t`, and also want to be able to use this module in init_by_lua.
So I think we should provide a init_by_lua to the other third party module for callback.

### DIFF
--- a/src/api/ngx_http_lua_api.h
+++ b/src/api/ngx_http_lua_api.h
@@ -49,6 +49,11 @@ ngx_shm_zone_t *ngx_http_lua_find_zone(u_char *name_data, size_t name_len);
 ngx_shm_zone_t *ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name,
     size_t size, void *tag);
 
+ngx_int_t ngx_http_lua_delay_list_register(ngx_conf_t *cf,
+    void *tag);
+
+ngx_int_t ngx_http_lua_delay_init_phase();
+
 
 #endif /* _NGX_HTTP_LUA_API_H_INCLUDED_ */
 
--- a/src/api/ngx_http_lua_api.h
+++ b/src/api/ngx_http_lua_api.h
@@ -49,10 +49,10 @@ ngx_shm_zone_t *ngx_http_lua_find_zone(u_char *name_data, size_t name_len);
 ngx_shm_zone_t *ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name,
     size_t size, void *tag);
 
-ngx_int_t ngx_http_lua_delay_list_register(ngx_conf_t *cf,
+ngx_int_t ngx_http_lua_delay_init_declare(ngx_conf_t *cf,
     void *tag);
 
-ngx_int_t ngx_http_lua_delay_init_phase();
+ngx_int_t ngx_http_lua_delay_init_handler();
 
 
 #endif /* _NGX_HTTP_LUA_API_H_INCLUDED_ */
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -147,7 +147,7 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name, size_t size,
     zone->init = ngx_http_lua_shared_memory_init;
     zone->data = ctx;
 
-    rc = ngx_http_lua_delay_list_register(cf, tag);
+    rc = ngx_http_lua_delay_init_declare(cf, tag);
     if (rc == NGX_ERROR) {
         return NULL;
     }
@@ -203,7 +203,7 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
         saved_cycle = ngx_cycle;
         ngx_cycle = ctx->cycle;
 
-        rc = ngx_http_lua_delay_init_phase();
+        rc = ngx_http_lua_delay_init_handler();
 
         ngx_cycle = saved_cycle;
 
@@ -218,7 +218,7 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
 
 
 ngx_int_t
-ngx_http_lua_delay_list_register(ngx_conf_t *cf, void *tag)
+ngx_http_lua_delay_init_declare(ngx_conf_t *cf, void *tag)
 {
     ngx_http_lua_main_conf_t     *lmcf;
     void                        **init;
@@ -229,25 +229,25 @@ ngx_http_lua_delay_list_register(ngx_conf_t *cf, void *tag)
         return NGX_ERROR;
     }
 
-    if (lmcf->delay_list == NULL) {
-        lmcf->delay_list =
+    if (lmcf->delay_init == NULL) {
+        lmcf->delay_init =
             ngx_array_create(cf->pool, 4, sizeof(void *));
 
-        if (lmcf->delay_list == NULL) {
+        if (lmcf->delay_init == NULL) {
             return NGX_ERROR;
         }
     }
 
-    init = lmcf->delay_list->elts;
+    init = lmcf->delay_init->elts;
 
-    for (i = 0; i < lmcf->delay_list->nelts; i++) {
+    for (i = 0; i < lmcf->delay_init->nelts; i++) {
 
         if (init[i] == tag) {
             return NGX_OK;
         }
     }
 
-    init = ngx_array_push(lmcf->delay_list);
+    init = ngx_array_push(lmcf->delay_init);
     if (init == NULL) {
         return NGX_ERROR;
     }
@@ -259,7 +259,7 @@ ngx_http_lua_delay_list_register(ngx_conf_t *cf, void *tag)
 
 
 ngx_int_t
-ngx_http_lua_delay_init_phase()
+ngx_http_lua_delay_init_handler()
 {
     ngx_http_lua_main_conf_t     *lmcf;
 
@@ -272,17 +272,17 @@ ngx_http_lua_delay_init_phase()
         return NGX_OK;
     }
 
-    lmcf->delay_list_inited++;
+    lmcf->delay_init_counter++;
 
-    if (lmcf->delay_list->nelts > lmcf->delay_list_inited) {
+    if (lmcf->delay_init->nelts > lmcf->delay_init_counter) {
         return NGX_ERROR;
     }
 
-    if (lmcf->delay_list->nelts < lmcf->delay_list_inited) {
+    if (lmcf->delay_init->nelts < lmcf->delay_init_counter) {
         return NGX_OK;
     }
 
-    /* lmcf->delay_list->nelts == lmcf->delay_list_inited */
+    /* lmcf->delay_init->nelts == lmcf->delay_init_counter */
     return lmcf->init_handler(lmcf->cycle->log, lmcf, lmcf->lua);
 }
 
--- a/src/ngx_http_lua_api.c
+++ b/src/ngx_http_lua_api.c
@@ -91,6 +91,7 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name, size_t size,
     ngx_shm_zone_t              **zp;
     ngx_shm_zone_t               *zone;
     ngx_http_lua_shm_zone_ctx_t  *ctx;
+    ngx_int_t                     rc;
     ngx_int_t                     n;
 
     lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module);
@@ -146,7 +147,10 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name, size_t size,
     zone->init = ngx_http_lua_shared_memory_init;
     zone->data = ctx;
 
-    lmcf->requires_shm = 1;
+    rc = ngx_http_lua_delay_list_register(cf, tag);
+    if (rc == NGX_ERROR) {
+        return NULL;
+    }
 
     return &ctx->zone;
 }
@@ -194,13 +198,12 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
 
     lmcf->shm_zones_inited++;
 
-    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts
-        && lmcf->init_handler)
+    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts)
     {
         saved_cycle = ngx_cycle;
         ngx_cycle = ctx->cycle;
 
-        rc = lmcf->init_handler(ctx->log, lmcf, lmcf->lua);
+        rc = ngx_http_lua_delay_init_phase();
 
         ngx_cycle = saved_cycle;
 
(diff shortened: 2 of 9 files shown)


agentzh:
@rainingmaster The 3rd-party module should use ngx_lua's public C API to register their shm zones. See

https://github.com/openresty/lua-nginx-module/blob/master/src/api/ngx_http_lua_api.h

And an example module here:

https://github.com/openresty/lua-nginx-module/tree/master/t/data/fake-shm-module

rainingmaster:
@agentzh We encountered a situation, we hope that using the shm in `http lua` and `stream lua`  at the same time, so the `init_by_lua` phase of the above two modules need to be ran  after `*_init_zone`. Therefore, it is desirable to have a callback function to control the execution time of `init_by_lua`

agentzh:
Okay, I see what problem you are trying to solve.

src/api/ngx_http_lua_api.h
```diff
@@ -49,6 +49,11 @@ ngx_shm_zone_t *ngx_http_lua_find_zone(u_char *name_data, size_t name_len);
 ngx_shm_zone_t *ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name,
     size_t size, void *tag);
 
+ngx_int_t ngx_http_lua_delay_list_register(ngx_conf_t *cf,
```
agentzh: We need a better name. "delay list" is very confusing.

src/ngx_http_lua_api.c
```diff
@@ -194,13 +198,12 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
 
     lmcf->shm_zones_inited++;
 
-    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts
-        && lmcf->init_handler)
+    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts)
```
agentzh: Why did you remove the check for an init handler here?

src/ngx_http_lua_api.c
```diff
@@ -213,4 +216,75 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
+    if (lmcf->delay_list->nelts > lmcf->delay_list_inited) {
+        return NGX_ERROR;
+    }
+
+    if (lmcf->delay_list->nelts < lmcf->delay_list_inited) {
+        return NGX_OK;
+    }
+
+    /* lmcf->delay_list->nelts == lmcf->delay_list_inited */
+     return lmcf->init_handler(lmcf->cycle->log, lmcf, lmcf->lua);
```
agentzh: Indentation issue?

src/ngx_http_lua_api.c
```diff
@@ -213,4 +216,75 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
+ngx_http_lua_delay_init_phase()
+{
+    ngx_http_lua_main_conf_t     *lmcf;
+
+    lmcf = ngx_http_cycle_get_module_main_conf(ngx_cycle, ngx_http_lua_module);
+    if (lmcf == NULL) {
+        return NGX_ERROR;
+    }
+
+    if (! lmcf->init_handler) {
```
agentzh: Style: we never put a space after the prefix operator `!`. Please fix other similar places as well.

src/ngx_http_lua_api.c
rainingmaster: @agentzh because I think use like a third module here: they can not get `lmcf->init_handler`,  and we can check init handler in `ngx_http_lua_delay_init_phase` api

src/ngx_http_lua_api.c
rainingmaster: @agentzh OK, I will fix it. Is there the same problem in https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_args.c#L253 and https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_capturefilter.c#L60, do I need to fix them? Or we just put a space before parameter in the second paragraph of judgment

rainingmaster:
@agentzh I am sorry so much style issue in my code... I've try to fix them. But I am not good at name the code, can you help me to name such as `delay_list`, `delay_list_init` and `ngx_http_lua_delay_list_register`, `ngx_http_lua_delay_init_phase`.
And [here](https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-3ed27dd07994dffb9efa8c835e198a15R262), it is safe enough to use `ngx_cycle` for other modules?
Thanks

src/ngx_http_lua_api.c
agentzh: @rainingmaster Yeah, those fixes should be in a separate PR. Each PR needs to be self-contained.

src/ngx_http_lua_api.c
```diff
@@ -213,4 +216,74 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
+        return NGX_ERROR;
+    }
+
+    if (!lmcf->init_handler) {
+        return NGX_OK;
+    }
+
+    lmcf->delay_list_inited++;
+
+    if (lmcf->delay_list->nelts > lmcf->delay_list_inited) {
```
agentzh: What does this mean?

src/ngx_http_lua_api.c
```diff
@@ -213,4 +216,74 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
+        return NGX_ERROR;
+    }
+
+    if (lmcf->delay_list->nelts < lmcf->delay_list_inited) {
```
agentzh: What does this mean?

src/ngx_http_lua_api.c
```diff
@@ -213,4 +216,74 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
     return NGX_OK;
 }
 
+
+ngx_int_t
+ngx_http_lua_delay_list_register(ngx_conf_t *cf, void *tag)
```
agentzh: What exactly is being registered here? Modules requiring delayed init handler invocation?

src/ngx_http_lua_common.h
```diff
@@ -176,6 +178,9 @@ struct ngx_http_lua_main_conf_s {
 
     ngx_array_t         *preload_hooks; /* of ngx_http_lua_preload_hook_t */
 
+    ngx_array_t         *delay_list; /* of void* */
+    unsigned             delay_list_inited;
```
agentzh: I don't know what these fields' purpose is.

src/ngx_http_lua_api.c
```diff
@@ -194,13 +198,12 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
 
-    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts
-        && lmcf->init_handler)
+    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts)
     {
         saved_cycle = ngx_cycle;
         ngx_cycle = ctx->cycle;
 
-        rc = lmcf->init_handler(ctx->log, lmcf, lmcf->lua);
+        rc = ngx_http_lua_delay_init_phase();
```
agentzh: What happens when there is no `lua_shared_dict` defined in `nginx.conf`? Is it delayed or not?

rainingmaster: @agentzh It will not delayed. Because no one use `ngx_http_lua_delay_list_register` to declare need to delay `init_by_lua`, and `init_by_lua` will still run [here](https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-c91b2ec425e9be9d56fee2df662b67d2R745)

src/ngx_http_lua_api.c
rainingmaster: @agentzh Yes, I make it like `required_shm` to mark delayed init handler, and we can run it in a same module many times.

src/ngx_http_lua_api.c
rainingmaster: @agentzh Because `ngx_http_lua_delay_init_phase` will  run in third part module, I think that if there are third party modules called `ngx_http_lua_delay_init_phase` too much times, that is, the number of times the `init` more than the number of `register`, the error should be returned.

src/ngx_http_lua_api.c
rainingmaster: @agentzh `ngx_http_lua_delay_init_phase` will run in different third part module. We will compare the number of times the init more than the number of the register: if `init` more than `register`, it is an error; if `init` less than `register`, wait other module run `ngx_http_lua_delay_init_phase`; if `init` equal `register`, run `init_by_lua`.

src/ngx_http_lua_common.h
rainingmaster: @agentzh  `delay_list` is used for register all third part module which need to delay `init_by_lua`, `delay_list_inited` is used for count how many module has run `ngx_http_lua_delay_init_phase`, and used to determine whether `init_by_lua` should be executed

src/ngx_http_lua_api.c
agentzh: @rainingmaster Then we need an error log message instead of simply returning an `NGX_ERROR`. The current way of handling this makes it very hard to debug this problem.

src/ngx_http_lua_common.h
agentzh: @rainingmaster I think you should add these as code comments to improve code readability. Also, I think we definitely need better names for these fields.

I don't think we should count the number of registrations. We only need to know whether or not to delay. Your registration API does not really register any real data from the 3rd-party modules. No callbacks from the 3rd-party modules are registered here anyway.

agentzh: @rainingmaster Using an array to count registrations is also very strange to me. It just does not make any sense. If you really want to count registrations, we should just maintain a single counter instead of an array without any useful data in its elements. Am I overlooking anything here?

agentzh: Also, please use the term "delay_init_handler" in your identifier names. Simply saying "delay" is very confusing.

agentzh:
Let me repeat in case you missed it:

We need a test module under `t/` and some tests to cover these new C APIs.

src/ngx_http_lua_common.h
rainingmaster: @agentzh using an array here because that I just want to make it like `required_shm`, we can run it again and again in a same third part module, and it is ported from `ngx_shared_memory_add`.
Maybe we just use a counter here and force mandatory third part module just run once in their own init phase?

rainingmaster:
@agentzh OK, I will add the test case as soon as possible.
Because I think of a situation: if a third party module needs to ensure that their postconfiguration run before `init_by_lua` (may need to use some of the variables in `init_by_lua` which defined in the third part module), we may also use this module. But I found that Lua can not use `ngx.var.*` to get this information in `init_by_module`, how can I improve this use case. Or only a third part `ngx_shm` is more appropriate?

agentzh:
@rainingmaster You know that NGINX variables are only available for requests? In those init contexts, there are not nginx requests yet, obviously.

src/ngx_http_lua_common.h
agentzh: @rainingmaster What do you mean by "third party module just run once in their own init phase"?

rainingmaster:
@agentzh Oh， I know. In addition to using `ngx_shm`, I have other ways  to test it ?

agentzh:
@rainingmaster See the existing test modules under `t/data/`. We only need a minimal nginx C module for testing these APIs. It can be dummy and not really functional. It's all about testing and code coverage.

src/ngx_http_lua_common.h
rainingmaster: @agentzh I have use (`ngx_http_lua_delay_init_declare`)[https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-3ed27dd07994dffb9efa8c835e198a15R243] to check that every third part module has register or not. 
If we only use a number as counter instead of array `delay_list`, we have to ensure the third part module register only not more than twice.

src/ngx_http_lua_api.c
```diff
@@ -194,13 +198,12 @@ ngx_http_lua_shared_memory_init(ngx_shm_zone_t *shm_zone, void *data)
...
 
-    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts
-        && lmcf->init_handler)
+    if (lmcf->shm_zones_inited == lmcf->shm_zones->nelts)
     {
         saved_cycle = ngx_cycle;
         ngx_cycle = ctx->cycle;
 
-        rc = lmcf->init_handler(ctx->log, lmcf, lmcf->lua);
+        rc = ngx_http_lua_delay_init_handler();
```
agentzh: When there is no `lua_shared_dict` directive configured in `nginx.conf`, there the delayed execution will never run, am I correct? This looks wrong to me.

rainingmaster: @agentzh YES. If there is no `lua_shared_dict`, or we can say there is no one declare delay `init_by_lua`(I've moved all delay `init_by_lua to use ngx_http_lua_delay_init_handler to implement, include original `lua_shared_dict`). Then, the `lmcf->delay_init` is `null`, `init_by_lua` will run in ngx_lua's postconfiguration, that is [here](https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-c91b2ec425e9be9d56fee2df662b67d2R746).

rainingmaster:
@agentzh I have try to add a test case to test the `ngx_http_lua_delay_init_handler` api. 
However, the ngx_cycle is controlled by third part module [here](https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-b4a8646df8a3d1d284dc4ab1916cbfbbR214), that may cause some problem to `init_by_lua`, this is not appropriate?

src/ngx_http_lua_api.c
agentzh: @rainingmaster That does not sound right to me. The 3rd-party modules can do whatever it likes, which may have nothing to do with ngx_lua's own `lua_shared_dict` directive. Or your change here is only meant for `ngx_meta_lua_module` and `lua_shared_dict`? That is, not meant for general 3rd-party modules?

rainingmaster: @agentzh The 3rd-part won't relay on the ngx_lua's own `lua_shared_dict`, they can use `ngx_http_lua_delay_init_handler` in their own module like [this](https://github.com/openresty/lua-nginx-module/pull/1050/files#diff-b4a8646df8a3d1d284dc4ab1916cbfbbR217), and `init_by_lua` will be delayed. 
But the disadvantage is that the third party module may need to maintain a `ngx_cycle`.

agentzh: @rainingmaster It's the 3rd-party module's responsibility to call ngx_lua's `init_by_lua*` code? Hmm, this looks very hacky and fragile to me.

agentzh:
I'm afraid this PR's approach has inherent problems in its direction.

The model should be like this:

* `ngx_meta_lua_module` implements the shdict (and other things).
* `ngx_meta_lua_module` can work with `ngx_stream_lua_module` alone, or with `ngx_http_lua_module` alone, or both, or neither.
* `ngx_http_lua_module` and `ngx_stream_lua_module` both *register* into `ngx_meta_lua_module`, but not the reversed direction.
* any 3rd-party nginx C modules who want to implement custom shm-based stuff to be shared between `ngx_http_lua_module` and `ngx_stream_lua_module` should register *only* to the `ngx_meta_lua_module`.

I'm afraid this PR fails to achieve the goal in a clean enough way. We should never expose bloody implementation details like "delay init handler" to the public C API for 3rd-party modules, which is too fragile, too hacky, and breaks encapsulation. Also, it's not scalable nor flexible.

PR #1053, opened 2017-04-25
balancer_by_lua_*: set upstream protocol http/https by lua

uvuv:
depend on the modified of lua-resty-corelib/ngx/balancer.lua

It not work on proxy_pass http://...
must be proxy_pass https://...

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -459,7 +459,7 @@ ngx_http_lua_balancer_save_session(ngx_peer_connection_t *pc, void *data)
 
 int
 ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
-    const u_char *addr, size_t addr_len, int port, char **err)
+    const u_char *addr, size_t addr_len, int port, int ssl, char **err)
 {
     ngx_url_t              url;
     ngx_http_lua_ctx_t    *ctx;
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_balancer.c
```diff
@@ -480,6 +480,24 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+            return NGX_ERROR;
+        }
+#else
+        *err = "https protocol requires SSL support";
+        return NGX_ERROR;
+#endif
+    } else if (ssl == 0) {
+        u->ssl = 0;
+    } else {
+        //by default
```
agentzh: Style: C++ style line comments should NEVER be used. You cannot find a single `//` comment in the NGINX core, can you? ;)

src/ngx_http_lua_balancer.c
```diff
@@ -480,6 +480,24 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+            *err = "proxy_pass http protocol only";
+            return NGX_ERROR;
+        }
+#else
+        *err = "https protocol requires SSL support";
+        return NGX_ERROR;
+#endif
+    } else if (ssl == 0) {
+        u->ssl = 0;
+    } else {
```
agentzh: Style: need a blank line before this `} else {` line

src/ngx_http_lua_balancer.c
```diff
@@ -480,6 +480,24 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
...
+            u->ssl = 1;
+        } else {
+            *err = "proxy_pass http protocol only";
+            return NGX_ERROR;
+        }
+#else
+        *err = "https protocol requires SSL support";
+        return NGX_ERROR;
+#endif
+    } else if (ssl == 0) {
```
agentzh: Style: need a blank line before this line.

src/ngx_http_lua_balancer.c
```diff
@@ -459,7 +459,7 @@ ngx_http_lua_balancer_save_session(ngx_peer_connection_t *pc, void *data)
 
 int
 ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
-    const u_char *addr, size_t addr_len, int port, char **err)
+    const u_char *addr, size_t addr_len, int port, int ssl, char **err)
```
agentzh: The semantics of this `ssl` parameter is very weird (0: off, 1: on, other values: default). Maybe we should introduce a separate C API function for setting SSL? Also, we may also want to dynamically configure whether to validate backend's SSL certificate and other things at the same time.

PR #1056, opened 2017-04-26
feature: add ngx.req.get_uri_ext API

hongliang5316:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -270,7 +270,7 @@ It is *highly* recommended to use [OpenResty releases](http://openresty.org) whi
 Alternatively, ngx_lua can be manually compiled into Nginx:
 
 1. Install LuaJIT 2.0 or 2.1 (recommended) or Lua 5.1 (Lua 5.2 is *not* supported yet). LuaJIT can be downloaded from the [LuaJIT project website](http://luajit.org/download.html) and Lua 5.1, from the [Lua project website](http://www.lua.org/).  Some distribution package managers also distribute LuaJIT and/or Lua.
-1. Download the latest version of the ngx_devel_kit (NDK) module [HERE](https://github.com/simpl/ngx_devel_kit/tags).
+1. Download the latest version of the ngx_devel_kit (NDK) module [HERE](https://github.com/simplresty/ngx_devel_kit/tags).
 1. Download the latest version of ngx_lua [HERE](https://github.com/openresty/lua-nginx-module/tags).
 1. Download the latest version of Nginx [HERE](http://nginx.org/) (See [Nginx Compatibility](#nginx-compatibility))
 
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_exten.h
```diff
@@ -0,0 +1,20 @@
+
+/*
+ * Copyright (C) Yichun Zhang (agentzh)
+ * Copyright (C) hongliang5316
```
agentzh: To get this merged, you have to grant your copyright to the authors of this module.

src/ngx_http_lua_exten.c
```diff
@@ -0,0 +1,67 @@
...
+void
+ngx_http_lua_inject_req_exten_api(lua_State *L)
+{
+    lua_pushcfunction(L, ngx_http_lua_ngx_req_get_uri_exten);
+    lua_setfield(L, -2, "get_uri_exten");
+}
+
+
+static int
+ngx_http_lua_ngx_req_get_uri_exten(lua_State *L)
```
agentzh: We no longer add new CFunction-based Lua API to the core.

New API functions should only go the FFI route and be exposed by the lua-resty-core library as a separate Lua module (like the existing `ngx.ssl` and `ngx.semaphore` modules and etc).

hongliang5316:
@agentzh Thanks for your review, i have removed some codes but only remaining FFI function.
I make a pull request to the `lua-resty-core` repository
https://github.com/openresty/lua-resty-core/pull/105
 Please help to review and provide suggestions, Thanks.

src/ngx_http_lua_exten.c
```diff
@@ -0,0 +1,35 @@
...
+
+int
+ngx_http_lua_ffi_req_get_ext(ngx_http_request_t *r, u_char *out, size_t *len)
+{
+    if (r->connection->fd == (ngx_socket_t) -1) {
+        return NGX_HTTP_LUA_FFI_BAD_CONTEXT;
+    }
+
+    *len = ngx_min(r->exten.len, *len);
+    ngx_memcpy(out, r->exten.data, *len);
```
agentzh: It seems that we can simply return the `r->exten.data` buffer to the Lua land because the `ffi.string()` API function will performance data copying anyway. It's a waste of CPU time to do an extra data copying.

hongliang5316: @agentzh Yes, this function keeps style of `ngx_http_lua_ffi_req_get_method_name`.

src/ngx_http_lua_exten.c
```diff
@@ -0,0 +1,34 @@
...
+
+
+#include "ngx_http_lua_util.h"
+
+
+char *
+ngx_http_lua_ffi_req_get_ext(ngx_http_request_t *r, size_t *len)
+{
+    if (r->connection->fd == (ngx_socket_t) -1) {
+        return NULL;
```
agentzh: I think we still want to preserve the return value for error codes. Better return the buffer pointer through a double pointer argument instead.

hongliang5316: @agentzh  Yeah, changed, and also https://github.com/openresty/lua-resty-core/pull/105.

config
```diff
@@ -318,6 +318,7 @@ HTTP_LUA_SRCS=" \
             $ngx_addon_dir/src/ngx_http_lua_req_body.c \
             $ngx_addon_dir/src/ngx_http_lua_uri.c \
             $ngx_addon_dir/src/ngx_http_lua_args.c \
+            $ngx_addon_dir/src/ngx_http_lua_exten.c \
```
agentzh: I guess this should just go into the `ngx_http_lua_uri.c` compilation unit? No need to create a new CU here I think.

src/ngx_http_lua_exten.c
```diff
@@ -0,0 +1,35 @@
...
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+
+#include "ngx_http_lua_util.h"
+
+
+int
+ngx_http_lua_ffi_req_get_ext(ngx_http_request_t *r, char **buf, size_t *len)
```
agentzh: I think `ngx_http_lua_ffi_req_get_uri_ext` is a better name here.

Issue #534, opened 2015-07-10
Support for client TLS certificate in tcpsock:sslhandshake


smira:
There's already an option to validate server TLS certificate using `lua_ssl_trusted_certificate`. 

It would be nice if we could provide client TLS certificate, in the same way as `ngx_http_proxy` module is doing: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_certificate, so that there would be two new options:

 * `lua_ssl_certificate`
 * `lua_ssl_certificate_key`

These options would initialise nginx ssl context with client certificates, which would be used during ssl handshake.

agentzh:
@smira Yes, this has been on my TODO list. But would you mind contributing a patch for it? ;)

Thanks!

rohitjoshi:
Any patch available?

Lekensteyn:
This appears to be fixed since March 2022 via commit 2b902657f370e392bd5066d2eafed7a3429af19e (PR #1602). There is a new `tcpsock:setclientcert` API.

denissabramovs:
I don't see and can't find setclientcert anywhere...

denissabramovs:
It was not merged...

zhuizhuhaomeng:
https://github.com/openresty/lua-nginx-module#tcpsocksetclientcert
It has been merged. But do not contain in the openresty-1.21.4 Release

PR #1060, opened 2017-05-05
add lua_ssl_certificate for server to certificate client

rainingmaster:
I have add lua_ssl_certificate ported from [proxy_module](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_certificate), and try to solve problem like [here](https://github.com/openresty/lua-nginx-module/issues/534). When use ngx.tcp.socket, we can send our certificate to server to pass verification. we have use like this in our private module, is it suitable?
If it is suitable, I will add more test cases~

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -249,6 +249,9 @@ typedef struct {
     ngx_uint_t              ssl_verify_depth;
     ngx_str_t               ssl_trusted_certificate;
     ngx_str_t               ssl_crl;
+    ngx_str_t               ssl_certificate;
+    ngx_str_t               ssl_certificate_key;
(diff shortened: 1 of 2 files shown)


agentzh:
@rainingmaster Maybe we should make it explicit that we are supporting *client* certificates here? The current names are a bit confusing.

I think maybe it's better to set this in the Lua API instead since this global setting affects all the SSL cosockets, which may give surprises. Like in this PR: https://github.com/openresty/lua-nginx-module/pull/997

rainingmaster:
OH, I kwnow. Yes, that is more flexible in that way.

CodingMinds:
@rainingmaster Is there a new PR ?

agentzh:
@CodingMinds Yes. See #997.

PR #1073, opened 2017-05-17
controll enter body filter or not by lua code

rainingmaster:

We got a problem in our progress: when we just want to use `body filter` in some special cases(maybe these case not only related to the domain name, but with the details of per request, we need to use lua code to judge in the actual situation), we have to run `body filter` in every request, and make some preparations for the lua execution environment like [these](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_bodyfilterby.c#L51), which is useless in some request.

We want to have a [switch](https://github.com/openresty/lua-nginx-module/compare/master...rainingmaster:enter_body_filter?expand=1#diff-c91b2ec425e9be9d56fee2df662b67d2R404), whether or not configured `body_filter_by_lua_*`, to control whether or not to default to enter `body filter` phase, and alson we have a [command](https://github.com/rainingmaster/lua-resty-core/blob/skip_body_filter/lib/ngx/bodyfilter.lua#L21) to control whether or not to enter body filter in each request by lua. So we've built this PR, so help me and see if there's any problem? thanks~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ install:
   - git clone https://github.com/openresty/rds-json-nginx-module.git ../rds-json-nginx-module
   - git clone https://github.com/openresty/srcache-nginx-module.git ../srcache-nginx-module
   - git clone https://github.com/openresty/redis2-nginx-module.git ../redis2-nginx-module
-  - git clone https://github.com/openresty/lua-resty-core.git ../lua-resty-core
+  - git clone -b skip_body_filter https://github.com/rainingmaster/lua-resty-core.git ../lua-resty-core
   - git clone -b v2.1-agentzh https://github.com/openresty/luajit2.git
 
 before_script:
--- a/src/ngx_http_lua_bodyfilterby.c
+++ b/src/ngx_http_lua_bodyfilterby.c
@@ -264,6 +264,11 @@ ngx_http_lua_body_filter(ngx_http_request_t *r, ngx_chain_t *in)
         }
     }
 
+    if (!ctx->enter_body_filter) {
+        dd("not enter body filter handler");
+        return ngx_http_next_body_filter(r, in);
+    }
+
     if (ctx->seen_last_in_filter) {
         for (/* void */; in; in = in->next) {
             dd("mark the buf as consumed: %d", (int) ngx_buf_size(in->buf));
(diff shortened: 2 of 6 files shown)


agentzh:
@rainingmaster Why don't you just add a flag into `ngx.ctx` table and check it at the beginning of your body filter handler? Like this:

```
access_by_lua_block {
    ...
    ngx.ctx.skip_body_filter = true
}

body_filter_by_lua_block {
    if ngx.ctx.skip_body_filter then
        return
    end

    ...
}
```

I don't see why you need to patch the core for this.

rainingmaster:
@agentzh. If we just use `ngx.ctx`, we will still enter some c function like `ngx_http_lua_body_filter_by_chunk`, and there will create some table and some malloc, which maybe is useless in this request.
In below case, we had use `body_filter_by_lua_block {  }` in our conf file for test.

We use openresty as a reverse proxy server, add use lua to do some control logic. There are may data write through `body_filter`. We found about 10% time use in prepare job for enter `body_filter` in `ngx_http_lua_body_filter_by_chunk`, and we have got a flame graph like this:


Some thing in `ngx_http_lua_body_filter_by_chunk`


We want to reduce this part of the loss. I think if we can control enter `ngx_http_lua_body_filter_by_chunk` or not, maybe it is better~

rainingmaster:
@agentzh  @doujiang24 
I have try to use config like 
```
rewrite_by_lua_block {
    ngx.ctx.skip_body_filter = true
}

body_filter_by_lua_block {
    if ngx.ctx.skip_body_filter then
        return
    end
    ...
}
```
and got flame graph like this:



when we config like:
```
rewrite_by_lua_block {
    lcoal b = require("ngx.bodyfilter")
    b.skip_body_filter(true)
}

body_filter_by_lua_block {
    ...
}
```
and we got a flame graph like:


so maybe we can reduce some useless run time in `ngx_http_lua_body_filter_by_chunk`?

agentzh:
@rainingmaster I think a much more useful optimization is to recycle the global Lua table created upon every `body_filter_by_lua*` invocation. This should make that part in your CPU flame graphs go away. Will you contribute a patch for that? This will also help the case where you do need to run the Lua body filter handler.

PR #1080, opened 2017-05-23
feature: added support for the 308 status code in ngx.redirect().

msenin:


t/022-redirect.t
```diff
@@ -269,3 +269,54 @@ GET /read
...
+            ngx.redirect("http://agentzh.org/foo?a=b&c=d", 308);
+            ngx.say("hi")
+        ';
+    }
+--- request
+GET /read
+--- response_headers
+Location: http://agentzh.org/foo?a=b&c=d
+--- response_body_like: 308 Permanent Redirect
+--- error_code: 308
```
agentzh: No newlines at the end of the file?

src/ngx_http_lua_consts.c
```diff
@@ -124,6 +124,11 @@ ngx_http_lua_inject_http_consts(lua_State *L)
     lua_setfield(L, -2, "HTTP_SEE_OTHER");
 #endif
 
+#if defined(nginx_version) && nginx_version >= 1013000
+    lua_pushinteger(L, NGX_HTTP_PERMANENT_REDIRECT);
```
agentzh: We can define this constant ourselves in case this macro is missing.

t/022-redirect.t
msenin: Fixed

src/ngx_http_lua_consts.c
msenin: I've defined NGX_HTTP_PERMANENT_REDIRECT as 308 in src/ngx_http_lua_common.h

src/ngx_http_lua_consts.c
```diff
@@ -124,6 +124,9 @@ ngx_http_lua_inject_http_consts(lua_State *L)
     lua_setfield(L, -2, "HTTP_SEE_OTHER");
 #endif
 
+    lua_pushinteger(L, NGX_HTTP_PERMANENT_REDIRECT);
+    lua_setfield(L, -2, "HTTP_PERMANENT_REDIRECT");
```
agentzh: I think you should adjust the initial table size of `ngx` earlier to avoid table auto-growth (which is slow). Also, you need to update the tests covering the number of `ngx` table fields in the test suite (otherwise the test suite would be failing).

t/022-redirect.t
```diff
@@ -318,5 +316,4 @@ Location: http://agentzh.org/foo?a=b&c=d
 GET /read
 --- response_headers
 Location: http://agentzh.org/foo?a=b&c=d
---- response_body_like: 308 Permanent Redirect
```
agentzh: Instead of removing these checks, I think it's better to adjust the test plan at the beginning of the test file by adding a 3 offset.

msenin: Does it mean to skip tests for 308?

agentzh: @msenin No, I mean still check the response body for that string pattern.

msenin: But it looks like nginx < 1.13.0 does not support "308 Permanent Redirect" code.

agentzh: @msenin I see. Never mind then.

msenin:
Hi @agentzh,
Cloud you please take a look at my pull request.
Now nginx used in openresty (1.13.6) supports code 308.
I've fixed tests.

t/022-redirect.t
msenin: @agentzh Now it should work just fine.

agentzh:
@msenin Merged. Thanks!

PR #1081, opened 2017-05-24
Add append_lua_package_path & append_lua_package_cpath directives

hongliang5316:
Sometimes we need to add a new program (`server`), but we don't want to change the content of `lua_package_path` and `lua_package_cpath`. Because those may be maintained by other programs, so we need more independent directives to configure `lua_path` and `lua_cpath`.
This can be used by this way:
```nginx
http {
    append_lua_package_path "/opt/program1/?.lua;";
    append_lua_package_cpath "/opt/program1/?.so;";
    server {
        #...
    }
    append_lua_package_path "/opt/program2/?.lua;";
    append_lua_package_cpath "/opt/program2/?.so;";
    server {
        #...
    }
}
```

### DIFF
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -241,8 +241,8 @@ ngx_http_append_lua_package_cpath(ngx_conf_t *cf, ngx_command_t *cmd,
         }
 
         if (ngx_array_init(lmcf->append_lua_cpath, cf->pool, 10,
-                    sizeof(ngx_str_t))
-                != NGX_OK)
+                           sizeof(ngx_str_t))
+            != NGX_OK)
         {
             return NGX_CONF_ERROR;
         }
@@ -283,8 +283,8 @@ ngx_http_append_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
         }
 
         if (ngx_array_init(lmcf->append_lua_path, cf->pool, 10,
-                    sizeof(ngx_str_t))
-                != NGX_OK)
+                           sizeof(ngx_str_t))
+            != NGX_OK)
         {
             return NGX_CONF_ERROR;
         }
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -221,6 +221,91 @@ ngx_http_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 }
 
 
+char *
+ngx_http_append_lua_package_cpath(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf)
+{
+
+    ngx_http_lua_main_conf_t *lmcf = conf;
+    ngx_str_t                *value;
+    ngx_str_t                *cpath;
+
+    dd("enter");
+
+    value = cf->args->elts;
+
+    if (lmcf->append_lua_cpath == NULL) {
+        lmcf->append_lua_cpath = ngx_palloc(cf->pool, sizeof(ngx_array_t));
+        if (lmcf->append_lua_cpath == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (ngx_array_init(lmcf->append_lua_cpath, cf->pool, 10,
+                    sizeof(ngx_str_t))
+                != NGX_OK)
+        {
+            return NGX_CONF_ERROR;
+        }
+    }
+
+    cpath = ngx_array_push(lmcf->append_lua_cpath);
+    if (cpath == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    cpath->len = value[1].len;
+    cpath->data = value[1].data;
+
+    if (!lmcf->requires_append_lua_cpath) {
+        lmcf->requires_append_lua_cpath = 1;
+    }
+
+    return NGX_CONF_OK;
+}
+
+
+char *
+ngx_http_append_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
+{
+
+    ngx_http_lua_main_conf_t *lmcf = conf;
+    ngx_str_t                *value;
+    ngx_str_t                *path;
+
+    dd("enter");
+
+    value = cf->args->elts;
+
+    if (lmcf->append_lua_path == NULL) {
+        lmcf->append_lua_path = ngx_palloc(cf->pool, sizeof(ngx_array_t));
+        if (lmcf->append_lua_path == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (ngx_array_init(lmcf->append_lua_path, cf->pool, 10,
+                    sizeof(ngx_str_t))
+                != NGX_OK)
+        {
+            return NGX_CONF_ERROR;
+        }
+    }
+
+    path = ngx_array_push(lmcf->append_lua_path);
+    if (path == NULL) {
+        return NGX_CONF_ERROR;
+    }
+
+    path->len = value[1].len;
+    path->data = value[1].data;
+
+    if (!lmcf->requires_append_lua_path) {
+        lmcf->requires_append_lua_path = 1;
+    }
+
+    return NGX_CONF_OK;
+}
+
+
 #if defined(NDK) && NDK
 char *
 ngx_http_lua_set_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -643,7 +643,8 @@ ngx_http_lua_init(ngx_conf_t *cf)
     ngx_int_t                   rc;
     ngx_uint_t                  i;
     size_t                      len;
-    char                       *path;
+    u_char                     *path;
+    u_char                     *tmp;
     ngx_str_t                  *p;
     ngx_array_t                *arr;
     ngx_http_handler_pt        *h;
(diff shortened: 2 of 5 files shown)


src/ngx_http_lua_module.c
```diff
@@ -661,6 +679,64 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
 
+    len = lmcf->lua_cpath.len;
+
+    if (lmcf->requires_append_lua_path) {
+        for (i = 0; i < lmcf->append_lua_cpath->nelts; i++) {
+            p = (ngx_str_t *) lmcf->append_lua_cpath->elts;
+            len += p[i].len;
+        }
+
+        path = (char *) ngx_palloc(cf->pool, len);
```
agentzh: Since you'll call `ngx_memzero()` on the memory block later anyway, why not just use `ngx_pcalloc` here instead?

src/ngx_http_lua_module.c
```diff
@@ -661,6 +679,64 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
+
+        ngx_memzero(path, len);
+
+        for (i = 0; i < lmcf->append_lua_path->nelts; i++) {
+            p = (ngx_str_t *) lmcf->append_lua_path->elts;
+            path = strncat(path, (char *) p[i].data, p[i].len);
+        }
+
+        if (lmcf->lua_path.len) {
+            path = strncat(path, (char *) lmcf->lua_path.data,
```
agentzh: Can we just use `ngx_copy()` instead to do the concatenation? It's more of the nginx style.

Please fix other similar places.

src/ngx_http_lua_directive.c
```diff
@@ -221,6 +221,91 @@ ngx_http_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    value = cf->args->elts;
+
+    if (lmcf->append_lua_path == NULL) {
+        lmcf->append_lua_path = ngx_palloc(cf->pool, sizeof(ngx_array_t));
+        if (lmcf->append_lua_path == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (ngx_array_init(lmcf->append_lua_path, cf->pool, 10,
+                    sizeof(ngx_str_t))
```
agentzh: Style: `sizeof` should be aligned with `lmcf->` in the previous line.

src/ngx_http_lua_directive.c
```diff
@@ -221,6 +221,91 @@ ngx_http_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
+    value = cf->args->elts;
+
+    if (lmcf->append_lua_cpath == NULL) {
+        lmcf->append_lua_cpath = ngx_palloc(cf->pool, sizeof(ngx_array_t));
+        if (lmcf->append_lua_cpath == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        if (ngx_array_init(lmcf->append_lua_cpath, cf->pool, 10,
+                    sizeof(ngx_str_t))
```
agentzh: Style: `sizeof` should be aligned with `lmcf->` in the previous line.

agentzh:
@hongliang5316 This is an interesting feature. Will you also add some corresponding tests to the existing test suite? Many thanks!

src/ngx_http_lua_module.c
hongliang5316: @agentzh  Sorry, i don't know enough about Nginx's common functions. I'll change it.

src/ngx_http_lua_module.c
hongliang5316: Yes, i'll chage it.

hongliang5316:
@agentzh All fixed. And also add some test cases.

src/ngx_http_lua_module.c
```diff
@@ -661,6 +680,62 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
+
+        path = (u_char *) ngx_pcalloc(cf->pool, len);
+        if (path == NULL) {
+            return NGX_ERROR;
+        }
+
+        tmp = path;
+
+        for (i = 0; i < lmcf->append_lua_cpath->nelts; i++) {
+            p = (ngx_str_t *) lmcf->append_lua_cpath->elts;
```
agentzh: I think we should move this statement outside the loop since it is invariant.

src/ngx_http_lua_module.c
```diff
@@ -661,6 +680,62 @@ ngx_http_lua_init(ngx_conf_t *cf)
         lmcf->postponed_to_access_phase_end = 0;
     }
 
+    len = lmcf->lua_cpath.len;
+
+    if (lmcf->requires_append_lua_cpath) {
+        for (i = 0; i < lmcf->append_lua_cpath->nelts; i++) {
+            p = (ngx_str_t *) lmcf->append_lua_cpath->elts;
```
agentzh: I think we should move this statement outside the loop since it is invariant.

src/ngx_http_lua_module.c
```diff
@@ -627,6 +641,11 @@ ngx_http_lua_init(ngx_conf_t *cf)
 {
     int                         multi_http_blocks;
     ngx_int_t                   rc;
+    ngx_uint_t                  i;
+    size_t                      len;
+    u_char                     *path;
+    u_char                     *tmp;
+    ngx_str_t                  *p;
```
agentzh: We usually use `p` for C string pointers only. Better use a different name for it.

src/ngx_http_lua_module.c
```diff
@@ -661,6 +680,62 @@ ngx_http_lua_init(ngx_conf_t *cf)
...
 
+    len = lmcf->lua_cpath.len;
+
+    if (lmcf->requires_append_lua_cpath) {
+        for (i = 0; i < lmcf->append_lua_cpath->nelts; i++) {
+            p = (ngx_str_t *) lmcf->append_lua_cpath->elts;
+            len += p[i].len;
+        }
+
+        path = (u_char *) ngx_pcalloc(cf->pool, len);
```
agentzh: The `path` variable is usually named `p` in the world of nginx according to its usage.

src/ngx_http_lua_module.c
agentzh: We should name the `p` variable to something else, like `value` maybe?

t/004-require.t
```diff
@@ -208,3 +208,125 @@ GET /ndk
...
+
+
+
+=== TEST 16: append package cpath (with lua_package_cpath)
+--- http_config eval
+    "lua_package_cpath '$::HtmlDir/?.so;;';
+    append_lua_package_cpath '/opt/program/?.so;';"
+--- config
+    location /main {
+        content_by_lua '
```
agentzh: Newly added test cases should use the `*_by_lua_block` form instead. The old string forms are already deprecated. Please fix other similar places.

t/004-require.t
```diff
@@ -208,3 +208,125 @@ GET /ndk
 GET /ndk
 --- response_body
 %20
+
+
```
agentzh: Better add a test case for the case where `lua_code_cache off` is configured.

agentzh:
@thibaultcha @doujiang24 @bungle  What do you think of this feature?

thibaultcha:
I am not so sure about the use case behind this. Considering the possibilities provided by `export LUA_PATH` and `;;`, we can:

Extend the `LUA_PATH` via the appropriate environment variable:
```
export LUA_PATH=./resty/?.lua;
```

Override `package.path` via the appropriate ngx_lua directive (what I think this PR tries to solve):
```
lua_package_path './lib/?.lua;';
```

Or, combine the OpenResty defaults *and* the env variable with `;;` (what I think this PR overlooked?):
```
http {
    lua_package_path './lib.lua;;';

    init_by_lua_block {
        print(package.path)
    }
}

```

The latest will do what this PR does, and also include the `LUA_PATH` env variable for even more flexibility (see beginning and end of log):

```
2017/05/29 17:18:33 [notice] 3925#0: [lua] init_by_lua:2: ./lib.lua;/usr/local/Cellar/openresty/1.11.2.2/site/lualib/?.lua;/usr/local/Cellar/openresty/1.11.2.2/site/lualib/?/init.lua;/usr/local/Cellar/openresty/1.11.2.2/lualib/?.lua;/usr/local/Cellar/openresty/1.11.2.2/lualib/?/init.lua;./resty/?.lua;
```

And for the cases needing even more flexibility, we can also do this in the Lua land (which I don't think is even necessary for the use case propose here):

```lua
package.path = "/opt/mylib/" .. package.path
```
Or did I misunderstand something that this PR is trying to solve?

src/ngx_http_lua_module.c
hongliang5316: Okay, i learned.

src/ngx_http_lua_module.c
hongliang5316: Sorry, that was a bit of mistake on my part. I'll change it.

t/004-require.t
hongliang5316: Yeah. I'll add this test case.

hongliang5316:
@thibaultcha Your comments are very good. But it doesn't apply to us.
For multiple apps, which have their only `lua_package_path`,  the `lua_package_path` directive and `LUA_PATH`  environment variable can be used only once. That's mean every app will share this. For us, i think it's not independent enough. Suppose we have one app to maintain `lua_package_path` directive in nginx.conf, But when we add a new app, this must be changed (we will upgrade the app to be compatible with the new app). 
It seems like we can modify the `package.path` in the Lua land, but compared with this pr, i think it's not intuitive enough.

spacewander:
@hongliang5316 
Sorry for my rude interruption...
I am afraid that modifying the `package.path` in Nginx configuration is less intuitive than in the Lua land.
Take an example like this:
```
http {
    append_lua_package_path "/opt/program1/?.lua;";
    server {
        # server A
    }
    append_lua_package_path "/opt/program2/?.lua;";
    server {
        # server B
    }
}
```

For guys not famous with Nginx configuration, `server A` and `server B` seem to use different `lua_package_path`. However, both `server A` and `server B` share the same `lua_package_path`.

Moreover, if we add a new server, called `server C`, with its own `append_lua_package_path`:
```
append_lua_package_path "/opt/program3/?.lua;";
server {
     # server C
}
```
Note that server A,B,C share the same `lua_package_path`. So the `append_lua_package_path` invited by server C not only changes its own `lua_package_path`, but also changes other servers' `lua_package_path`.

Assumed we only allow one app to maintain `lua_package_path` directive(or `LUA_PATH` environment variable), we are able to do some conflict detection(or hacks around `package.loaded`) in the entrance.

agentzh:
@thibaultcha @spacewander Thank you for your feedback! Yeah, it seems to me now that it would be more elegant and more flexible to just manipulate the search paths on the Lua land.

agentzh:
@hongliang5316 Maybe we should just allow multiple `init_by_lua*` directive instances?

hongliang5316:
@spacewander Your analysis is very good. This can lead to a conflict among apps.
@agentzh Yeah, allowing multiple `init_by_lua*` directive instances is more versatile and flexible, i will work for this.

agentzh:
@hongliang5316 Great. Thanks! You can follow a similar style as this PR: #935.

PR #1095, opened 2017-06-20
bugfix: (FFI shdict) every non-OK result should set the error message

javierguerragiraldez:
as mentioned in [lua-resty-core issue #120](https://github.com/openresty/lua-resty-core/issues/120), make sure FFI functions set the error message when not succesful.


agentzh:
@javierguerragiraldez Mind to add some test cases? It would be great if the test suite can cover these.

PR #1097, opened 2017-06-24
balancer: support arbitrary HTTP codes when calling ngx.exit inside b…

dndx:
…alancer_by_lua.

This PR depends on https://github.com/openresty/openresty/pull/270

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -314,7 +314,8 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
 
     if (ctx->exited && ctx->exit_code != NGX_OK) {
         rc = ctx->exit_code;
-        if (rc == NGX_ERROR || rc == NGX_BUSY || rc == NGX_DECLINED) {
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_balancer.c
```diff
@@ -314,7 +314,8 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
 
     if (ctx->exited && ctx->exit_code != NGX_OK) {
         rc = ctx->exit_code;
-        if (rc == NGX_ERROR || rc == NGX_BUSY || rc == NGX_DECLINED) {
+        if (rc == NGX_ERROR || rc == NGX_BUSY || rc == NGX_DECLINED
+            || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
```
agentzh: Style: we should use the following form here instead:

```C
if (rc == A
    || rc == B
    || rc == C
    || rc >= D)
{
    ...
}
```

agentzh: Also, we need to use special macro to ensure that we still work the same way with the stock nginx cores.

agentzh:
Merged. Thanks!

PR #1106, opened 2017-07-09
feature: added a pure C API function +ngx_http_lua_ffi_raw_log

thibaultcha:
Sister PR for openresty/lua-resty-core#128

To be used by ngx.errlog module's `rawlog()` function in the lua-resty-core library.

### DIFF
--- a/src/ngx_http_lua_log.c
+++ b/src/ngx_http_lua_log.c
@@ -428,6 +428,30 @@ ngx_http_lua_ffi_errlog_get_sys_filter_level(ngx_http_request_t *r)
     return log_level;
 }
 
+
+int
+ngx_http_lua_ffi_raw_log(ngx_http_request_t *r, unsigned int level, u_char *s,
+    size_t s_len, u_char *err, size_t *errlen)
+{
+    ngx_log_t           *log;
+
+    if (level > NGX_LOG_DEBUG || level < NGX_LOG_STDERR) {
+        *errlen = ngx_snprintf(err, *errlen, "bad log level: %d", level) - err;
+        return NGX_ERROR;
+    }
+
+    if (r && r->connection && r->connection->log) {
+        log = r->connection->log;
+
+    } else {
+        log = ngx_cycle->log;
+    }
+
+    ngx_log_error(level, log, 0, "%*s", s_len, s);
+
+    return NGX_OK;
+}
+
 #endif
 
 /* vi:set ft=c ts=4 sw=4 et fdm=marker: */


src/ngx_http_lua_log.c
```diff
@@ -428,6 +428,30 @@ ngx_http_lua_ffi_errlog_get_sys_filter_level(ngx_http_request_t *r)
...
+    }
+
+    if (r && r->connection && r->connection->log) {
+        log = r->connection->log;
+
+    } else {
+        log = ngx_cycle->log;
+    }
+
+    ngx_log_error((ngx_uint_t) level, log, 0, "[lua] %*s", s_len, s);
```
agentzh: I think we should also omit the `[lua] ` prefix here. Raw logs should be truly "raw" :)

thibaultcha: Oh well, I actually didn't want it but thought it would be preferable, happy to remove it!

thibaultcha: Done!

p0pr0ck5:
@thibaultcha just curious why we'd define `ngx_http_lua_ffi_raw_log` to accept and error string pointer, when we only have known conditions in which we'd fill the buffer and return `NGX_ERROR`. Could we not just assert at this point that a return code of `NGX_ERROR` means the level was incorrect, and leave it up to the caller to throw an error? This would save having to get the resty core buffer and buffer size pointers (and avoid the need to wrap the err in `ffi.string`). I can understand both sides of the coin here, just curious about the thought process behind the implementation.

thibaultcha:
@p0pr0ck5 No particular reason, except the fact that the error is due to a C-level check and thus created in the C-level (although, in the context of the FFI, I am perfectly fine with assuming the error's meaning in Lua-land as well). Maintainability maybe?

I'd be fine moving it to the Lua-land.

src/ngx_http_lua_log.c
```diff
@@ -428,6 +428,30 @@ ngx_http_lua_ffi_errlog_get_sys_filter_level(ngx_http_request_t *r)
     return log_level;
 }
 
+
+int
+ngx_http_lua_ffi_raw_log(ngx_http_request_t *r, int level, u_char *s,
```
agentzh: Maybe it's better to use `unsigned int` here instead of `int` so that we do not have to do the explicit type cast to `ngx_uint_t` below?

thibaultcha: SGTM, changed.

agentzh:
Merged by following @p0pr0ck5's suggestion. Thanks!

Issue #1124, opened 2017-07-26
ngx.escape_uri should escape RFC 3986 section 2.2 Reserved Characters


hunterli:
$ python -c "import urllib; print urllib.quote(\"'\")"
%27

$ resty -e "ngx.say(ngx.escape_uri(\"'\"))"
'
ngx.escape_uri should escape "'" to %27 but not.

agentzh:
@hunterli Will you contribute a pull request for it? Thanks!

hunterli:
@agentzh OK, I'll try it.

agentzh:
@hunterli Great! Looking forward to it!

spacewander:
I am afraid I don't think it is a bug.

According to the document, `ngx.escape_uri` is used to:
> Escape str as a URI component.

Although its name is `escape_uri`.

Escape uri is different from escaping URI component.
The second one doesn't need to escape `'`.
Refer to `RFC 3986 section 2.2`:
> A subset of the reserved characters (gen-delims) is used as
   delimiters of the generic URI components described in Section 3

Note that the RFC only requires the **gen-delims** part of reserved characters. **Not all the reserved characters**.

What `urllib.quote` does is more like escaping uri instead of escaping URI component.
According to its document, `urllib.quote` is used to:
> By default, this function is intended for quoting the path section of the URL.

Don't mix up these two differnt things.

hunterli:
@spacewander 
FYI: [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent](url) said:
To be more stringent in adhering to RFC 3986 (which reserves !, ', (, ), and *), even though these characters have no formalized URI delimiting uses, the following can be safely used

spacewander:
@hunterli 
Could you point out where RFC 3986 says that `!, ', (, ), and *` are reserved for URI component?

spacewander:
One place I found including `!, ', (, ), and *` is the `D.2.  Modifications`, under `Appendix D.  Changes from RFC 2396 `:
> The mark characters that are typically unsafe to decode,
   including the exclamation mark ("!"), asterisk ("*"), single-quote
   ("'"), and open and close parentheses ("(" and ")"), have been moved
   to the reserved set in order to clarify the distinction between
   reserved and unreserved and, hopefully, to answer the most common
   question of scheme designers.

However, it just says `'` is reserved compared with `RFC 2396`, doesn't say it is reserved for URI component.

[...]

PR #1128, opened 2017-07-31
Revert "bugfix: ngx.escape_uri: we did not escape URI reserved chars. #1124"

spacewander:
Commit f170505186ff61af36b3e126772b671793af9428 breaks the compatibility with RFC 3986. Here is two reasons:
1. Quote from RFC 3986 Section 2.2:
> A subset of the reserved characters (gen-delims) is used as
delimiters of the generic URI components described in Section 3

Note that RFC 3986 says 'a subset of the reserved characters
(gen-delims)', not all the reserved characters. The characters escaped
in that commit are 'sub-delims'. They are not required to be escaped
according to Section 2.2.

2. Refer to RFC 3986 "Appendix A.  Collected ABNF for URI", sub-delims
could be used as part of query and other components. This use case shows
that sub-delims are valid in some component of URI. Therefore, it would
be better if we don't escape them for URI component.


agentzh:
@spacewander Is our behavior consistent with web browser's JavaScript encodeURIComponent()? This is the only thing I care.

spacewander:
@agentzh 
encodeURIComponent doesn't escape `!~*'()`.

```javascript
> encodeURIComponent("!~*'()")
'!~*\'()'
```

agentzh:
@spacewander Then it's good! I'll merge this.

agentzh:
@spacewander Merged. Thanks!

Issue #1131, opened 2017-08-02
ngx.flush throws runtime error


tokers:
Hi!

I found `ngx.flush` throws a Lua runtime error after the `ngx.print` was called by accident.

you can reproduce the problem by the following nginx configuration.

```nginx
server {
    listen 8081;
    server_name _;
    location / {
        content_by_lua_block {
            ngx.say("Hello, World!")
            ngx.flush()
        }
    }
}
```

```bash
curl -v http://127.0.0.1:8081/ -H 'If-Unmodified-Since: Mon, 17 Apr 2006 03:10:35 GMT'
```

Then you will see something like this in the corresponding `error.log`:

```log
2017/08/02 16:37:50 [error] 21399#21399: *3 lua entry thread aborted: runtime error: content_by_lua(nginx.conf:42):3: no request ctx found
stack traceback:
coroutine 0:
        [C]: in function 'flush'
        content_by_lua(nginx.conf:42):3: in function <content_by_lua(nginx.conf:42):1>, client: 127.0.0.1, server: _, request: "GET / HTTP/1.1", host: "127.0.0.1:8081"
```

After reading the related codes, i find the problem is caused by `ngx_http_not_modified_header_filter` at the earliest, which jumps to `ngx_http_filter_finalize_request` when `ngx_http_test_if_unmodified` returns 0, and `ngx_http_filter_finalize_request` cleans ctx of all the http modules, sets `r->filter_finalize ` to 1.

Then we calling `ngx.flush`, we will get the provious runtime error because the ctx of ngx_lua is cleaned.

Maybe we should check the `r->filter_finalize` in `ngx_http_lua_ngx_flush` like `ngx_http_lua_handle_exit` and others?

> openresty version in my local machine:

```
nginx version: openresty/1.11.2.4
built by gcc 7.1.1 20170526 (Red Hat 7.1.1-2) (GCC)
built with OpenSSL 1.0.2k  26 Jan 2017
TLS SNI support enabled
[...]

PR #1132, opened 2017-08-03
bugfix: ngx.flush will throw error after calling output API functions

tokers:
When calling output API functions like ngx.print with
If-Unmodified-Since(or If-Match) in the request headers,
the ngx_http_not_modified_filter_module will "test the If-Unmodified-Since",
if failed, ngx_http_filter_finalize_request will be called and the contexts of
all http modules will be cleared, after that, error "no request found" will
be throwed if we calling ngx.flush.

Please see the github issue
https://github.com/openresty/lua-nginx-module/issues/1131
for the details.

### DIFF
--- a/src/ngx_http_lua_output.c
+++ b/src/ngx_http_lua_output.c
@@ -49,6 +49,7 @@ ngx_http_lua_ngx_echo(lua_State *L, unsigned newline)
     int                          i;
     int                          nargs;
     int                          type;
+    int                          uri_changes;
     const char                  *msg;
 
     r = ngx_http_lua_get_req(L);
@@ -146,8 +147,23 @@ ngx_http_lua_ngx_echo(lua_State *L, unsigned newline)
         size += sizeof("\n") - 1;
     }
 
+    uri_changes = r->uri_changes;
+
     if (size == 0) {
         rc = ngx_http_lua_send_header_if_needed(r, ctx);
+
+        if (ctx != ngx_http_get_module_ctx(r, ngx_http_lua_module)
+            && r->filter_finalize)
+        {
+            if (uri_changes > r->uri_changes) {
+                ctx->filter_finalize = 1;
+                return lua_yield(L, 0);
+
+            } else {
+                ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
+            }
+        }
+
         if (rc == NGX_ERROR || rc > NGX_OK) {
             lua_pushnil(L);
             lua_pushliteral(L, "nginx output filter error");
@@ -231,6 +247,18 @@ ngx_http_lua_ngx_echo(lua_State *L, unsigned newline)
                    newline ? "lua say response" : "lua print response");
 
     rc = ngx_http_lua_send_chain_link(r, ctx, cl);
+
+    if (ctx != ngx_http_get_module_ctx(r, ngx_http_lua_module)
+        && r->filter_finalize)
+    {
+        if (uri_changes > r->uri_changes) {
+            ctx->filter_finalize = 1;
+            return lua_yield(L, 0);
+
+        } else {
+            ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
+        }
+    }
 
     if (rc == NGX_ERROR || rc >= NGX_HTTP_SPECIAL_RESPONSE) {
         lua_pushnil(L);
@@ -683,6 +711,7 @@ ngx_http_lua_inject_output_api(lua_State *L)
 static int
 ngx_http_lua_ngx_send_headers(lua_State *L)
 {
+    int                      uri_changes;
     ngx_int_t                rc;
     ngx_http_request_t      *r;
     ngx_http_lua_ctx_t      *ctx;
@@ -705,7 +734,22 @@ ngx_http_lua_ngx_send_headers(lua_State *L)
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
                        "lua send headers");
 
+        uri_changes = r->uri_changes;
+
         rc = ngx_http_lua_send_header_if_needed(r, ctx);
+
+        if (ctx != ngx_http_get_module_ctx(r, ngx_http_lua_module)
+            && r->filter_finalize)
+        {
+            if (uri_changes > r->uri_changes) {
+                ctx->filter_finalize = 1;
+                return lua_yield(L, 0);
+
+            } else {
+                ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
+            }
+        }
+
         if (rc == NGX_ERROR || rc > NGX_OK) {
             lua_pushnil(L);
             lua_pushliteral(L, "nginx output filter error");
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_util.c
```diff
@@ -421,6 +421,11 @@ ngx_http_lua_send_header_if_needed(ngx_http_request_t *r,
             dd("sending headers");
             rc = ngx_http_send_header(r);
             ctx->header_sent = 1;
+
+            if (r->filter_finalize) {
+                ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
```
agentzh: This looks wrong to me. The header filter may initiate an internal redirect to another location. It looks very wrong to pollute the new location with the old location's lua module ctx data (and it may rewrite the new location's existing ctx data).

t/014-bugs.t
```diff
@@ -1018,3 +1018,49 @@ write timer set: 1
...
+--- config
+    location = /t1 {
+        content_by_lua_block {
+            ngx.say("Hello world")
+            ngx.flush()
+        }
+    }
+
+    location = /t2 {
+        lua_http10_buffering on;
```
agentzh: Why this? This makes any difference?

src/ngx_http_lua_util.c
agentzh: It's also wrong to just keep running the original location's lua code while the newly entered location's lua code may have already been running.

agentzh:
@tokers We should terminate the current request immediately just like the existing code handling `r->filter_finalize`. You can see that we always call `ngx_http_lua_request_cleanup()` in the existing code. And for this API, we should yield the current Lua light thread immediately so upper layer code can terminate it.

agentzh:
@tokers Just like the existing Lua API functions `ngx.exec()` and `ngx.exit()`. We should not proceed with the current Lua thread execution any more. It's just wrong.

tokers:
@agentzh Sorry for my fault, i will fix it, thanks your patient guidance.

tokers:
@agentzh In this case, i think not only the `ngx.flush`, after calling `ngx.say`, `ngx.print` or `ngx.send_header` with the error_page configuration like:

```nginx
location / {
    error_page 412 /t2;
    content_by_lua_block {
        ngx.say("location /")
        ngx.log(ngx.ERR, "after ngx.say")
    }
}

location /t2 {
    content_by_lua_block {
        ngx.log(ngx.ERR, "location /t2")
    }
}
``` 

And i see the `after saying` in the error.log.

So maybe we should terminate in the internal of `ngx.say`, `ngx.print` and `ngx.send_headers`?

agentzh:
@tokers Yeah, we can abort the current main light thread inside `ngx.send_headers` and etc using the same `lua_yield()` approach as `ngx.exit()` and `ngx.exec()`, I think? Alas, the filter finalization thing is a hacky feature inside the nginx.

tokers:
@agentzh I think yield the current main Light thread can't solve this problem. :(

When someone has the following config:

```nginx
location /t1 {
    error_page 412 /t2;
    content_by_lua_block {
        ngx.say("location /")
    }
}

location /t2 {
    content_by_lua_block {
        ngx.say("location /t2")
    }
}
```

And sent request just like before.
```bash
curl -v http://127.0.0.1:7106/t1 -H 'If-Unmodified-Since: Mon, 17 Apr 2006 03:10:35 GMT'
```

The header filter will be called in `ngx.say`(`/t1`), and `filter_finalize` will be triggered, but the logic will jump to the location `/t2` because the `error_page`, and the `ngx.say`(`/t2`) will be called, in this time,  if we yield the current Lua thread, we will yield the one running new location(`/t2`) 's code. This is not expected.

So is there any other way can we fix it or we may need to terminate the request?

agentzh:
@tokers Yes, I think terminate the request handler in `location /t1` in a similar way to `ngx.exec()` and `ngx.exit()`. It has nothing to do with `location /t2`.

tokers:
@agentzh I will try to fix it following this way.

tokers:
@agentzh Just like you say, i want to yield the current Light thread inside the output API functions in a similar way to `ngx.exit()` and `ngx.exec()`, there some same cleanup operations  like `ngx_http_lua_request_cleanup`, `ngx_http_lua_cleanup_pending_operation` and etc in `ngx_http_lua_handle_xxx`, shall i need to reuse these codes by creating a new function?

tokers:
@agentzh Sorry for my unskilled, i organized the logic for this pull request.

> sense: filter finalization.  

> case 1: no internal redirects happen

We still in the original location, in this case, i think we can call the `ngx_http_set_ctx()` for resuming the ngx_lua's ctx when filter finalization detected in these output API functions.

> case 2: interval redirects happen because the error_page or others

Now we can't call the `ngx_http_set_ctx()` for this will pollute the new location, instead, we need to terminate the request handler in old location, specifically, when filter finalization happens, we yield the current running Light thread, and handle the filter finalization by the higher code(do some operations like `ngx_http_lua_request_cleanup`).  
There is another special situation, these output API functions can be called in the new location as well, new location's code may run before the output API functions called in the old location return(because of the  behaivor of `ngx_http_internel_redirect()`), in this case, we will detect the filter finalization in the new location's code firstly, we also need to distinguish it.

agentzh:
@tokers BTW, the Travis CI checks on our branch is failing: https://travis-ci.org/openresty/lua-nginx-module/builds/261524307 Your branch does not even compile with nginx 1.11.2.

tokers:
@agentzh Now the tests passed in Travis CI.

agentzh:
@tokers Thanks for your contribution!

@doujiang24 @dndx Will you please review this PR for me? Many thanks!

src/ngx_http_lua_output.c
```diff
@@ -146,8 +147,23 @@ ngx_http_lua_ngx_echo(lua_State *L, unsigned newline)
...
+
+        if (ctx != ngx_http_get_module_ctx(r, ngx_http_lua_module)
+            && r->filter_finalize)
+        {
+            if (uri_changes > r->uri_changes) {
+                ctx->filter_finalize = 1;
+                return lua_yield(L, 0);
+
+            } else {
+                ngx_http_set_ctx(r, ctx, ngx_http_lua_module);
```
agentzh: Overriding the new ctx looks too dangerous to me. When can this actually happen? Will you add a corresponding test case to actually cover this code path?

tokers: When `r->filter_finalize` == 1, all the http modules' ctx was cleared(see `ngx_http_filter_finalize_request`), so this `ngx_http_set_ctx(r, ctx, ngx_http_lua_module);` is used to resume the ctx of ngx_lua.

I will add a corresponding test case for this.

zhuizhuhaomeng:
This PR need to reopen.

dragonorloong:
> This PR need to reopen.

Why hasn't this PR been merged? Is the code implementation unreasonable?

zhuizhuhaomeng:
@dragonorloong  can you reproduce this issue?
Would you please rebase this PR to the latest master ?

dragonorloong:
@zhuizhuhaomeng This issue can be reproduced
Is this PR not merged because of a conflict?

zhuizhuhaomeng:
The repository that submitted this pull request has been deleted.
So, I cannot reopen this pull request.
Would you please submit another pull request to resolve this issue?

zhuizhuhaomeng:
The repository that submitted this pull request has been deleted.
So, I cannot reopen this pull request.
Please submit another pull request to resolve this issue.

PR #1142, opened 2017-08-19
bugfix: fix a coredump when use ngx.socket.tcp:connect multi time

rainingmaster:
When using ngx.socket.tcp, and we want to reconnect in Lua, we maybe code like this:
```
local sock = ngx.socket.tcp()

for i = 1, 3 do -- retry
    local ok, err = sock:connect("www.google.com", 80)
    if not ok then
        ngx.log(ngx.ERR, "failed to connect: ", err)
    end
end
```

But if in a bad network environment, if **resolver send query failed**(In the test case, I use iptables to implement it like [this](https://github.com/rainingmaster/lua-nginx-module/commit/eca0936293d387b2962c7e082650392d2499c211#diff-354f30a63fb0907d4ad57269548329e3R93), it takes some problem: if we use `connect` multiple times and call `ngx_http_lua_socket_tcp_connect` many times, In first times, the `coctx->cleanup` will set as [`ngx_http_lua_tcp_resolve_cleanup`](https://github.com/rainingmaster/lua-nginx-module/blob/resolver-reject/src/ngx_http_lua_socket_tcp.c#L735). In the second times,we will destroy `rctx` in [here](https://github.com/rainingmaster/lua-nginx-module/blob/resolver-reject/src/ngx_http_lua_socket_tcp.c#L734) by `ngx_resolve_name_done`. Therefore, I think cleanup should be set as `NULL` [at the first exit](https://github.com/rainingmaster/lua-nginx-module/blob/resolver-reject/src/ngx_http_lua_socket_tcp.c#L746)

Here is the old and core test cases: https://www.travis-ci.org/rainingmaster/lua-nginx-module/builds/266258969

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,8 @@ before_script:
   - mysql -uroot -e 'create database ngx_test; grant all on ngx_test.* to "ngx_test"@"%" identified by "ngx_test"; flush privileges;'
 
 script:
+  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT || exit 0
+  - sudo iptables -L || exit 0
   - cd luajit2/
   - make -j$JOBS CCDEBUG=-g Q= PREFIX=$LUAJIT_PREFIX CC=$CC XCFLAGS='-DLUA_USE_APICHECK -DLUA_USE_ASSERT -msse4.2' > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make install PREFIX=$LUAJIT_PREFIX > build.log 2>&1 || (cat build.log && exit 1)
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,8 @@ before_script:
   - mysql -uroot -e 'create database ngx_test; grant all on ngx_test.* to "ngx_test"@"%" identified by "ngx_test"; flush privileges;'
 
 script:
-  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT || exit 0
-  - sudo iptables -L || exit 0
+  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT
+  - sudo iptables -L
   - cd luajit2/
   - make -j$JOBS CCDEBUG=-g Q= PREFIX=$LUAJIT_PREFIX CC=$CC XCFLAGS='-DLUA_USE_APICHECK -DLUA_USE_ASSERT -msse4.2' > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make install PREFIX=$LUAJIT_PREFIX > build.log 2>&1 || (cat build.log && exit 1)
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * 190 + 6;
+plan tests => repeat_each() * 193;
 
 our $HtmlDir = html_dir;
 
@@ -3693,10 +3693,16 @@ close: 1 nil
 
 
 
-=== TEST 61: resolver send query failed
+=== TEST 61: resolver send query failing immediately in connect()
+this case did not clear coctx->cleanup properly and would lead to memory invalid accesses.
+
+this test case requires the following iptables rule to work properly:
+
+sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT
 --- config
     location /t {
-    	resolver 127.0.0.1:10086 ipv6=off;
+        resolver 127.0.0.1:10086 ipv6=off;
+        resolver_timeout 10ms;
 
         content_by_lua_block {
             local sock = ngx.socket.tcp()
@@ -3718,5 +3724,5 @@ failed to connect: www.google.com could not be resolved
 failed to connect: www.google.com could not be resolved
 failed to connect: www.google.com could not be resolved
 hello!
---- error_log
-failed (1: Operation not permitted) while resolving
+--- error_log eval
+qr{\[alert\] .*? send\(\) failed \(\d+: Operation not permitted\) while resolving}
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * 190;
+plan tests => repeat_each() * 190 + 6;
 
 our $HtmlDir = html_dir;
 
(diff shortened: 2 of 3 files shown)


agentzh:
@rainingmaster Thanks for the report and patch!

@dndx Will you please review this PR? Many thanks!

dndx:
@agentzh seems to be a reasonable fix as [`ngx_resolve_name` will free `rctx` automatically when it fails](http://lxr.nginx.org/source/src/core/ngx_resolver.c#0486). In that case, keeping the `coctx->cleanup` set will cause next time `connect()` is called, the newly allocated `rctx` being prematurely freed.

t/058-tcp-socket.t
```diff
@@ -3690,3 +3690,33 @@ received: OK
 close: 1 nil
 --- no_error_log
 [error]
+
+
+
+=== TEST 61: resolver send query failed
```
agentzh: @rainingmaster This test case always fails on my side due to the timeout error:

```
t/058-tcp-socket.t .. # I found ONLY: maybe you're debugging?
t/058-tcp-socket.t .. 1/386
#   Failed test 'ERROR: client socket timed out - TEST 61: resolver send query failed
# '
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1929.

#   Failed test 'TEST 61: resolver send query failed - status code ok'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 936.
#          got: ''
#     expected: '200'

#   Failed test 'TEST 61: resolver send query failed - response_body - response is expected (repeated req 0, req 0)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1496.
# @@ -1,4 +0,0 @@
# -failed to connect: www.google.com could not be resolved
# -failed to connect: www.google.com could not be resolved
# -failed to connect: www.google.com could not be resolved
# -hello!

#   Failed test 'TEST 61: resolver send query failed - pattern "failed (1: Operation not permitted) while resolving" should match a line in error.log (req 0)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1147.
t/058-tcp-socket.t .. 5/386
#   Failed test 'ERROR: client socket timed out - TEST 61: resolver send query failed
# '
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1929.

#   Failed test 'TEST 61: resolver send query failed - status code ok'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 936.
#          got: ''
#     expected: '200'

#   Failed test 'TEST 61: resolver send query failed - response_body - response is expected (repeated req 1, req 0)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1496.
# @@ -1,4 +0,0 @@
# -failed to connect: www.google.com could not be resolved
# -failed to connect: www.google.com could not be resolved
# -failed to connect: www.google.com could not be resolved
# -hello!

#   Failed test 'TEST 61: resolver send query failed - pattern "failed (1: Operation not permitted) while resolving" should match a line in error.log (req 1)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1147.
t/058-tcp-socket.t .. Failed 386/386 subtests
```

agentzh: I think the actual behavior relies on the actual system's network configurations. Can we make this test case more portable please?

t/058-tcp-socket.t
```diff
@@ -3690,3 +3690,33 @@ received: OK
...
+--- config
+    location /t {
+    	resolver 127.0.0.1:10086 ipv6=off;
```
agentzh: Use of tabs here?

.travis.yml
```diff
@@ -90,6 +90,8 @@ before_script:
   - mysql -uroot -e 'create database ngx_test; grant all on ngx_test.* to "ngx_test"@"%" identified by "ngx_test"; flush privileges;'
 
 script:
+  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT || exit 0
```
agentzh: We'd better avoid requiring the developers to mess up with the local firewall settings just to run your test case. But it looks tricky to make that code path get exercised without yielding.

t/058-tcp-socket.t
agentzh: Okay, never mind, it requires special local iptables rule to work. Alas.

.travis.yml
agentzh: Why `|| exit 0` here? Shouldn't we require this to succeed? Otherwise that test case must time out?

agentzh:
@rainingmaster Hmm, this code path is very tricky to trigger in general. Just wondering how did you find this :)

I'd propose this extra patch against your branch:

```diff
diff --git a/.travis.yml b/.travis.yml
index 1046e4d3..49eeec90 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,8 @@ before_script:
   - mysql -uroot -e 'create database ngx_test; grant all on ngx_test.* to "ngx_test"@"%" identified by "ngx_test"; flush privileges;'

 script:
-  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT || exit 0
-  - sudo iptables -L || exit 0
+  - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT
+  - sudo iptables -L
   - cd luajit2/
   - make -j$JOBS CCDEBUG=-g Q= PREFIX=$LUAJIT_PREFIX CC=$CC XCFLAGS='-DLUA_USE_APICHECK -DLUA_USE_ASSERT -msse4.2' > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make install PREFIX=$LUAJIT_PREFIX > build.log 2>&1 || (cat build.log && exit 1)
diff --git a/t/058-tcp-socket.t b/t/058-tcp-socket.t
index 01e0b840..5dc1695e 100644
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -3693,10 +3693,17 @@ close: 1 nil



-=== TEST 61: resolver send query failed
+=== TEST 61: resolver send query failing immediately in connect()
+this case did not clear coctx->cleanup properly and would lead to memory invalid accesses.
+
+this test case requires the following iptables rule to work properly:
+
+sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT
+
 --- config
     location /t {
-       resolver 127.0.0.1:10086 ipv6=off;
+        resolver 127.0.0.1:10086 ipv6=off;
+        resolver_timeout 10ms;

         content_by_lua_block {
             local sock = ngx.socket.tcp()
@@ -3718,5 +3725,5 @@ failed to connect: www.google.com could not be resolved
 failed to connect: www.google.com could not be resolved
 failed to connect: www.google.com could not be resolved
 hello!
---- error_log
-failed (1: Operation not permitted) while resolving
+--- error_log eval
+qr{\[alert\] .*? send\(\) failed \(\d+: Operation not permitted\) while resolving}
```

t/058-tcp-socket.t
```diff
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * 190;
+plan tests => repeat_each() * 190 + 6;
```
agentzh: This test plan is wrong since the `repeat_each` value may get changed in special test modes like the "check leak" test mode, in which case the test plan is just wrong. We should make the test plan irrelevant to the actual values of `repeat_each`, always. The fix is

```diff
diff --git a/t/058-tcp-socket.t b/t/058-tcp-socket.t
index 5dc1695e..0743dd73 100644
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;

 repeat_each(2);

-plan tests => repeat_each() * 190 + 6;
+plan tests => repeat_each() * 193;

 our $HtmlDir = html_dir;
```

rainingmaster:
@agentzh Sorry I left some test code in prove commit, I've modified all the test cases as you requested.
Besides, the problem was discovered by [@xiaocang](https://github.com/xiaocang), and I just made a few minor changes. ;)

agentzh:
@rainingmaster Already merged. Thank you both!

PR #1145, opened 2017-08-23
ngx ssl add client addr function

smartwjw:

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -880,16 +880,17 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
 #endif
 }
 
+
 int
 ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
     size_t *addrlen, int *addrtype, char **err)
 {
 #if (NGX_HAVE_UNIX_DOMAIN)
-    struct sockaddr_un *saun;
+    struct sockaddr_un  *saun;
 #endif
-    ngx_ssl_conn_t *ssl_conn;
-    ngx_connection_t *c;
-    struct sockaddr_in *sin;
+    ngx_ssl_conn_t      *ssl_conn;
+    ngx_connection_t    *c;
+    struct sockaddr_in  *sin;
 #if (NGX_HAVE_INET6)
     struct sockaddr_in6 *sin6;
 #endif
@@ -919,14 +920,14 @@ ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
         break;
 #endif
 
-# if(NGX_HAVE_UNIX_DOMAIN)
+# if (NGX_HAVE_UNIX_DOMAIN)
     case AF_UNIX:
         saun = (struct sockaddr_un *)c->sockaddr;
-        /* on Linux sockaddr might not include sun_path at all */ 
-        if(c->socklen <= (socklen_t) offsetof(struct sockaddr_un, sun_path))
-        {
+        /* on Linux sockaddr might not include sun_path at all */
+        if (c->socklen <= (socklen_t) offsetof(struct sockaddr_un, sun_path)) {
             *addr = "";
             *addrlen = 0;
+
         } else {
             *addr = saun->sun_path;
             *addrlen = ngx_strlen(saun->sun_path);
@@ -946,6 +947,7 @@ ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
 
     return NGX_OK;
 }
+
 
 int
 ngx_http_lua_ffi_cert_pem_to_der(const u_char *pem, size_t pem_len, u_char *der,
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -899,7 +899,7 @@ ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
     }
 
     addr_text = &r->connection->addr_text;
-    *addr = (char *)addr_text->data;
+    *addr = (char *) addr_text->data;
     *addrlen = addr_text->len;
 
     return NGX_OK;
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -881,11 +881,18 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
 }
 
 int
-ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
-    size_t *addrlen, char **err)
+ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
+    size_t *addrlen, int *addrtype, char **err)
 {
+#if (NGX_HAVE_UNIX_DOMAIN)
+    struct sockaddr_un *saun;
+#endif
     ngx_ssl_conn_t *ssl_conn;
-    ngx_str_t      *addr_text;
+    ngx_connection_t *c;
+    struct sockaddr_in *sin;
+#if (NGX_HAVE_INET6)
+    struct sockaddr_in6 *sin6;
+#endif
 
     if (r->connection == NULL || r->connection->ssl == NULL) {
         *err = "bad request";
@@ -898,15 +905,44 @@ ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
         return NGX_ERROR;
     }
 
-#if (NGX_HAVE_UNIX_DOMAIN)
-    *addr = "";
-    *addrlen = 0;
-    return NGX_OK;
+    c = ngx_ssl_get_connection(ssl_conn);
+
+    switch (c->sockaddr->sa_family) {
+
+#if (NGX_HAVE_INET6)
+    case AF_INET6:
+        sin6 = (struct sockaddr_in6 *) c->sockaddr;
+        *addrlen = 16;
+        *addr = (char *) &sin6->sin6_addr.s6_addr;
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_INET6;
+
+        break;
 #endif
 
-    addr_text = &r->connection->addr_text;
-    *addr = (char *) addr_text->data;
-    *addrlen = addr_text->len;
+# if(NGX_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+        saun = (struct sockaddr_un *)c->sockaddr;
+        /* on Linux sockaddr might not include sun_path at all */ 
+        if(c->socklen <= (socklen_t) offsetof(struct sockaddr_un, sun_path))
+        {
+            *addr = "";
+            *addrlen = 0;
+        } else {
+            *addr = saun->sun_path;
+            *addrlen = ngx_strlen(saun->sun_path);
+        }
+
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_UNIX;
+        break;
+#endif
+
+    default: /* AF_INET */
+        sin = (struct sockaddr_in *) c->sockaddr;
+        *addr = (char *) &sin->sin_addr.s_addr;
+        *addrlen = 4;
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_INET;
+        break;
+    }
 
     return NGX_OK;
 }
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -898,6 +898,12 @@ ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
         return NGX_ERROR;
     }
 
+#if (NGX_HAVE_UNIX_DOMAIN)
+    *addr = "";
+    *addrlen = 0;
+    return NGX_OK;
+#endif
+
     addr_text = &r->connection->addr_text;
     *addr = (char *) addr_text->data;
     *addrlen = addr_text->len;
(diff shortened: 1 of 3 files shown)


src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,30 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+    }
+
+    ssl_conn = r->connection->ssl->connection;
+    if (ssl_conn == NULL) {
+        *err = "bad ssl conn";
+        return NGX_ERROR;
+    }
+
+    addr_text = &r->connection->addr_text;
+    *addr = (char *)addr_text->data;
```
agentzh: Style: need a space after `(char *)`.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,30 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
 #endif
 }
 
+int
+ngx_http_lua_ffi_ssl_client_addr(ngx_http_request_t *r, char **addr,
```
agentzh: OK, you are returning a raw client address here, so you should rename this C function to something like `ngx_http_lua_ffi_ssl_raw_client_addr`, just like the existing `ngx_http_lua_ffi_ssl_raw_server_addr` C function. We should be consistent with ourselves here.

smartwjw: I first name it `raw_client_addr`, I am returning ip address in plain text, like `ngx.var.remote_addr`, not `ngx.var.binary_remote_addr`, so I name it `client_addr`, am I correct?

src/ngx_http_lua_ssl_certby.c
smartwjw: I'll fix it soon.

src/ngx_http_lua_ssl_certby.c
agentzh: @smartwjw I see. Need some test cases for various different cases like unix domain socket clients.

smartwjw: @agentzh I've added some test cases for `client_addr`.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,198 @@ failed to do SSL handshake: handshake failed
...
+close: 1 nil
+
+--- error_log
+client ip: 127.0.0.1
+
+--- no_error_log
+[error]
+[alert]
+
+
```
agentzh: Please use the [reindex](https://github.com/openresty/openresty-devel-utils/blob/master/reindex) script to re-format this `.t` file. Thank you!

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,198 @@ failed to do SSL handshake: handshake failed
...
+received: Server: nginx
+received: Content-Type: text/plain
+received: Content-Length: 4
+received: Connection: close
+received: 
+received: foo
+close: 1 nil
+
+--- error_log
+client ip:
```
agentzh: We should support unix socket clients, no?

smartwjw: Yes,  but unix socket address is useless in this scenery, so I simply return empty string when using unix socket, is that OK?

t/139-ssl-cert-by.t
smartwjw: done

t/139-ssl-cert-by.t
agentzh: @smartwjw No, it's not OK. This API should be compatible with UNIX domain socket. And a single nginx `server {}` may serve different kinds of incoming SSL traffic at the same time. Being able to inspect unix domain sockets is also very useful. And we should be consistent with the existing `raw_server_addr` API.

agentzh: @smartwjw For such a low level API, we should not only target your own single use case. That's just not right :) Thanks for your understanding!

src/ngx_http_lua_ssl_certby.c
agentzh: I think we should do raw_client_addr here, which is more useful. Parsing the binary form is much more efficient than parsing the textual form.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,36 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
 #endif
 }
```
agentzh: Style: nginx requires 2 blank lines to separate adjacent C function definitions.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,36 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+    *addrlen = 0;
+    return NGX_OK;
+#endif
+
+    addr_text = &r->connection->addr_text;
+    *addr = (char *) addr_text->data;
+    *addrlen = addr_text->len;
+
+    return NGX_OK;
+}
```
agentzh: Ditto.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,199 @@ failed to do SSL handshake: handshake failed
...
+received: Server: nginx
+received: Content-Type: text/plain
+received: Content-Length: 4
+received: Connection: close
+received: 
+received: foo
+close: 1 nil
+
+--- error_log
+client addr:
```
agentzh: Can we get the address of the client unix domain socket at all?

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,72 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+    size_t *addrlen, int *addrtype, char **err)
+{
+#if (NGX_HAVE_UNIX_DOMAIN)
+    struct sockaddr_un *saun;
+#endif
+    ngx_ssl_conn_t *ssl_conn;
+    ngx_connection_t *c;
+    struct sockaddr_in *sin;
+#if (NGX_HAVE_INET6)
+    struct sockaddr_in6 *sin6;
```
agentzh: Style: all the local variable identifiers should be aligned up vertically. Please see other C functions for examples.

src/ngx_http_lua_ssl_certby.c
agentzh: @smartwjw Seen this comment?

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,72 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+
+        break;
+#endif
+
+# if(NGX_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+        saun = (struct sockaddr_un *)c->sockaddr;
+        /* on Linux sockaddr might not include sun_path at all */ 
+        if(c->socklen <= (socklen_t) offsetof(struct sockaddr_un, sun_path))
+        {
```
agentzh: Style: `{` should be in the previous line unless the `if` condition is too long and would exceed 80 columns.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,72 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_INET6;
+
+        break;
+#endif
+
+# if(NGX_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+        saun = (struct sockaddr_un *)c->sockaddr;
+        /* on Linux sockaddr might not include sun_path at all */ 
+        if(c->socklen <= (socklen_t) offsetof(struct sockaddr_un, sun_path))
```
agentzh: Style: need a space after the keyword `if`. Please see the rest of the code base (and also the official nginx core) for examples.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,72 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+        {
+            *addr = "";
+            *addrlen = 0;
+        } else {
```
agentzh: Style: need a blank line before this `} else {` line. Please see the rest of the code base (and also the official nginx core) for examples.

t/139-ssl-cert-by.t
agentzh: Do we support unix domain socket now? Hello?

agentzh:
@smartwjw BTW, using the `ngx-releng` script found the following coding style issues in your code:

screen shot 2017-10-06 at 1 04 21 pm

src/ngx_http_lua_ssl_certby.c
```diff
@@ -880,6 +880,72 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+        *addr = (char *) &sin6->sin6_addr.s6_addr;
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_INET6;
+
+        break;
+#endif
+
+# if(NGX_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+        saun = (struct sockaddr_un *)c->sockaddr;
+        /* on Linux sockaddr might not include sun_path at all */
```
agentzh: Line trailing space on this line. Also detected by the `ngx-releng` tool.

agentzh:
Also, I noticed that your branch fails the travis ci test. Will you make that pass? BTW, you can point lua-resty-core to your own branch in `.travis.yml` temporarily in this branch.

t/139-ssl-cert-by.t
smartwjw: The program itself support unix domain now, but I tested on my mac and a CentOS vm, it all returns empty string, so the test code don't need to change.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -881,6 +881,74 @@ ngx_http_lua_ffi_ssl_server_name(ngx_http_request_t *r, char **name,
...
+        *addrlen = 16;
+        *addr = (char *) &sin6->sin6_addr.s6_addr;
+        *addrtype = NGX_HTTP_LUA_ADDR_TYPE_INET6;
+
+        break;
+#endif
+
+# if (NGX_HAVE_UNIX_DOMAIN)
+    case AF_UNIX:
+        saun = (struct sockaddr_un *)c->sockaddr;
```
agentzh: Style: needs a space after type casting expressions like `(struct sockaddr_un *)`.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+
+    server {
+        listen 127.0.0.1:12345 ssl;
+        server_name   test.com;
+
+        ssl_certificate_by_lua_block { 
+            local ssl = require "ngx.ssl"
+            local byte = string.byte
+            local addr, addrtype, err = ssl.raw_client_addr()
+            local ip = string.format("%d.%d.%d.%d", byte(addr, 1), byte(addr, 2),
```
spacewander: @smartwjw 
Better to localize `string.format` just like the `string.byte`?

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+    lua_package_path "../lua-resty-core/lib/?.lua;;";
+
+    server {
+        listen 127.0.0.1:12345 ssl;
+        server_name   test.com;
+
+        ssl_certificate_by_lua_block { 
+            local ssl = require "ngx.ssl"
+            local byte = string.byte
+            local addr, addrtype, err = ssl.raw_client_addr()
```
spacewander: Better to check if `addrtype` is equal to `inet`.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+--- http_config
+    lua_package_path "../lua-resty-core/lib/?.lua;;";
+
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        server_name   test.com;
+
+        ssl_certificate_by_lua_block { 
+            local ssl = require "ngx.ssl"
+            local addr, addrtyp, err = ssl.raw_client_addr()
```
spacewander: Better to check if `addrtyp` is equal to `unix`.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+received: 
+received: foo
+close: 1 nil
+
+--- error_log
+client socket file: 
+
+--- no_error_log
+[error]
+[alert]
```
spacewander: Is it possible to add a test case for IPv6, like the [TEST 54: IPv6](https://github.com/openresty/lua-nginx-module/blob/master/t/058-tcp-socket.t#L3231)?

Simply check if the client ip and addrtype are expected is enough.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+
+=== TEST 22: get raw_client_addr - IPv4
+--- http_config
+    lua_package_path "../lua-resty-core/lib/?.lua;;";
+
+    server {
+        listen 127.0.0.1:12345 ssl;
+        server_name   test.com;
+
+        ssl_certificate_by_lua_block {
```
agentzh: Should remove the line-trailing spaces on this line.

t/139-ssl-cert-by.t
```diff
@@ -1866,3 +1866,206 @@ failed to do SSL handshake: handshake failed
...
+
+=== TEST 23: get raw_client_addr - unix domain socket
+--- http_config
+    lua_package_path "../lua-resty-core/lib/?.lua;;";
+
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        server_name   test.com;
+
+        ssl_certificate_by_lua_block {
```
agentzh: Ditto.

agentzh:
Merged with minor fixes. Thanks!

PR #1146, opened 2017-08-25
Add receiveatmost.

robertabcd:
receiveatmost is just like receive, but does not wait until the receive
buffer is fully filled, or the timeout passes. It receive at least 1
byte and at most N bytes, specified by the argument.

This is useful as there's currently no nonblocking reads. Users can
process input as soon as data arrives.

Note: I'm not very familiar with nginx development. Inputs are appreciated.


agentzh:
@robertabcd I think we already have a similar pending PR:

https://github.com/openresty/lua-nginx-module/pull/861

We meant to merge it once it passes all of our review.

robertabcd:
Oh, didn't notice that. That one certainly looks better. Feel free to drop this one, as I noticed some tests are failing. It would take time for me to fix that too.

PR #1147, opened 2017-08-30
add a function for get socket fd

mlboy:

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -19,6 +19,7 @@
 
 
 static int ngx_http_lua_socket_tcp(lua_State *L);
+static int ngx_http_lua_socket_tcp_getsockfd(lua_State *L);
 static int ngx_http_lua_socket_tcp_connect(lua_State *L);
 #if (NGX_HTTP_SSL)
 static int ngx_http_lua_socket_tcp_sslhandshake(lua_State *L);
@@ -264,6 +265,9 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
 
     lua_pushcfunction(L, ngx_http_lua_socket_tcp_settimeouts);
     lua_setfield(L, -2, "settimeouts"); /* ngx socket mt */
+
+    lua_pushcfunction(L, ngx_http_lua_socket_tcp_getsockfd);
+    lua_setfield(L, -2, "getsockfd"); /* ngx socket mt */
 
     lua_pushvalue(L, -1);
     lua_setfield(L, -2, "__index");
@@ -782,6 +786,29 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
     return lua_yield(L, 0);
 }
 
+static int
+ngx_http_lua_socket_tcp_getsockfd(lua_State *L)
+{
+    ngx_connection_t                    *c;
+    ngx_http_lua_socket_tcp_upstream_t    *u;
+
+    luaL_checktype(L, 1, LUA_TTABLE);
+
+    lua_rawgeti(L, 1, SOCKET_CTX_INDEX);
+    u = lua_touserdata(L, -1);
+    lua_pop(L, 1);
+
+    if (u == NULL || u->peer.connection == NULL) {
+        lua_pushnil(L);
+        lua_pushliteral(L, "closed");
+        return 2;
+    }
+
+    c = u->peer.connection;
+    lua_pushinteger(L,(int) c->fd);
+
+    return 1;
+}
 
 static void
 ngx_http_lua_socket_empty_resolve_handler(ngx_resolver_ctx_t *ctx)
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -19,7 +19,7 @@
 
 
 static int ngx_http_lua_socket_tcp(lua_State *L);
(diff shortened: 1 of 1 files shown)


agentzh:
@mlboy Will you please explain your use cases for this API?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -782,6 +786,29 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
     return lua_yield(L, 0);
 }
 
+static int
+ngx_http_lua_socket_tcp_getsockfd(lua_State *L)
+{
+    ngx_connection_t                    *c;
+    ngx_http_lua_socket_tcp_upstream_t    *u;
```
agentzh: Style: variable identifiers must be vertically aligned up. See other similar places for examples.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -782,6 +786,29 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
...
+    lua_pop(L, 1);
+
+    if (u == NULL || u->peer.connection == NULL) {
+        lua_pushnil(L);
+        lua_pushliteral(L, "closed");
+        return 2;
+    }
+
+    c = u->peer.connection;
+    lua_pushinteger(L,(int) c->fd);
```
agentzh: Style: need a space after the comma character.

agentzh: Also, need to test if the fd is -1 before returning it.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -782,6 +786,29 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
     return lua_yield(L, 0);
 }
```
agentzh: Style: we use 2 blank lines to separate successive C function definitions. See other similar places for examples.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -782,6 +786,29 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
...
+        lua_pushnil(L);
+        lua_pushliteral(L, "closed");
+        return 2;
+    }
+
+    c = u->peer.connection;
+    lua_pushinteger(L,(int) c->fd);
+
+    return 1;
+}
```

src/ngx_http_lua_socket_tcp.c
```diff
@@ -265,6 +266,9 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
     lua_pushcfunction(L, ngx_http_lua_socket_tcp_settimeouts);
     lua_setfield(L, -2, "settimeouts"); /* ngx socket mt */
 
+    lua_pushcfunction(L, ngx_http_lua_socket_tcp_getsockfd);
+    lua_setfield(L, -2, "getsockfd"); /* ngx socket mt */
```
agentzh: The `sock` word in the method name is redundant. It cannot really be something else. It can only be called on cosocket objects anyway. Or are you following LuaSocket's Lua API here? (I couldn't find such an API in LuaSocket).

mlboy:
step 2  i will add a method like sock:send_fd(fd,...)
and then can send a socket to fd which is you want 
case  :one for one chatroom

agentzh:
@mlboy Sending fds across process boundary can be tricky in that you would also have to pass all the associated state over and rebuild all the data structures in the other process. All this work may incur significant overhead and is very tricky to get exactly right given the complexity of the problem.

For your use case, I think it's better to just pass "control flow" and "data" across the process boundary instead of passing fds and cosockets. We already have production users who use ngx.semaphore + lua_shared_dict + ngx.sleep to do such kind of things, achieving 1.5 million ~ 2 million connections in a single box. We could make this approach even more efficient when we add listen() and accept() methods to the cosocket objects so that we can sync data and control flow on the nginx worker process level more efficiently than the lua_shared_dict + ngx.sleep approach (the intra-worker level request dispatch is already very efficient with ngx.semaphore).

zhuizhuhaomeng:
If anyone needs this API, should reimplement it via ffi.

gustavosbarreto:
I’d like to reimplement this via FFI, but I’m not sure about the best approach within the lua-nginx-module. Are there any existing examples or recommendations on how to properly work using FFI in this context?

If you have any guidance or references, I’d really appreciate it.

zhuizhuhaomeng:
here is an example: ngx_http_lua_ffi_socket_tcp_sslhandshake

linsite:
> I’d like to reimplement this via FFI, but I’m not sure about the best approach within the lua-nginx-module. Are there any existing examples or recommendations on how to properly work using FFI in this context?
> 
> I need `getfd` because I’m working on a solution where Nginx needs to pass an open socket descriptor to another process using SCM_RIGHTS. In this case, simply creating a new connection is not sufficient. I need to transfer an already open file descriptor from Nginx to another process.
> 
> If you have any guidance or references, I’d really appreciate it.

Hi, are you still working on this? I share the same need.

gustavosbarreto:
> Hi, are you still working on this? I share the same need.

@linsite https://github.com/gustavosbarreto/websocket-fd-handoff

linsite:
> > Hi, are you still working on this? I share the same need.
> 
> @linsite https://github.com/gustavosbarreto/websocket-fd-handoff

Thanks a lot.

zhuizhuhaomeng:
Have added the getfd() function.
This api will be available in the next release.

PR #1149, opened 2017-08-30
Add FFI methods for taking stats to ngx.shared.DICT

hnakamur:
This is a pull request for adding `get_stats` method to `ngx.shared.DICT`.

I'd like to know memory usage of a `ngx.shared.DICT`.
I noticed there is the systemtap solution https://github.com/openresty/openresty-systemtap-toolkit/blob/master/ngx-shm
However, I'd like to know memory usage on other OSes (such as FreeBSD) than Linux,
so I need a Lua API.

`get_stats` returns the information of the `stats` field in `ngx_slab_pool_t`
https://github.com/nginx/nginx/blob/418124e21983aa53db4d99b7fd2eb0eee7ab4be0/src/core/ngx_slab.h#L25-L31
The return value is an array table of record tables which has the following keys:

* size
* total
* used
* reqs
* fails

The last four keys corresponds to the fields in `ngx_slab_stat_t`.
https://github.com/nginx/nginx/blob/418124e21983aa53db4d99b7fd2eb0eee7ab4be0/src/core/ngx_slab.h#L25-L31

The `size` key are a calculated page size for each slot.
The size of the first slot is `pool->min_size` and the size of following slot is twice of the previous slot.
https://github.com/nginx/nginx/blob/418124e21983aa53db4d99b7fd2eb0eee7ab4be0/src/core/ngx_slab.c#L208-L216

An example config:

```
...
http {
    lua_shared_dict dogs 10m;

    server {
        ...
        location /set {
            content_by_lua_block {
                local dogs = ngx.shared.dogs
                for i = 1, 2000 do
                    dogs:set(string.format("key%d", i), string.rep("v", i))
                end
                ngx.say("STORED")
            }
        }
        location /stats {
            content_by_lua_block {
                local dogs = ngx.shared.dogs
                local stats = dogs:get_stats()
                for slot, s in ipairs(stats) do
                    ngx.say(string.format("slot=%d, size=%d, total=%d, used=%d, reqs=%d, fails=
%d", slot, s.size, s.total, s.used, s.reqs, s.fails))
                end
            }
        }
    }
}
```

An example output:

```
# curl localhost/set
STORED
# curl localhost/stats
slot=1, size=8, total=0, used=0, reqs=0, fails=0
slot=2, size=16, total=0, used=0, reqs=0, fails=0
slot=3, size=32, total=127, used=1, reqs=1, fails=0
slot=4, size=64, total=0, used=0, reqs=0, fails=0
slot=5, size=128, total=64, used=56, reqs=56, fails=0
slot=6, size=256, total=128, used=127, reqs=127, fails=0
slot=7, size=512, total=256, used=256, reqs=256, fails=0
slot=8, size=1024, total=512, used=512, reqs=512, fails=0
slot=9, size=2048, total=1024, used=1023, reqs=1023, fails=0
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6674,7 +6674,7 @@ This feature was first introduced in the `v0.7.3` release.
 [Back to TOC](#nginx-api-for-lua)
 
 ngx.shared.DICT.capacity
----------------------
+------------------------
 **syntax:** *capacity_bytes = ngx.shared.DICT:capacity()*
 
 **context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
@@ -6683,7 +6683,6 @@ ngx.shared.DICT.capacity
 
 Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
 the [lua_shared_dict](#lua_shared_dict) directive.
-
 
 Example:
 
@@ -6698,6 +6697,8 @@ Example:
 This feature was first introduced in the `v0.10.11` release.
 
 **Note:** This method requires the `resty.core.shdict` or `resty.core` modules from the [lua-resty-core](https://github.com/openresty/lua-resty-core) library.
+
+This feature requires at least nginx core version `0.7.3`.
 
 See also [ngx.shared.DICT](#ngxshareddict).
 
--- a/README.markdown
+++ b/README.markdown
@@ -3196,7 +3196,8 @@ Nginx API for Lua
 * [ngx.shared.DICT.flush_all](#ngxshareddictflush_all)
 * [ngx.shared.DICT.flush_expired](#ngxshareddictflush_expired)
 * [ngx.shared.DICT.get_keys](#ngxshareddictget_keys)
-* [ngx.shared.DICT.stats](#ngxshareddictstats)
+* [ngx.shared.DICT.capacity](#ngxshareddictcapacity)
+* [ngx.shared.DICT.free_space](#ngxshareddictfree_space)
 * [ngx.socket.udp](#ngxsocketudp)
 * [udpsock:setpeername](#udpsocksetpeername)
 * [udpsock:send](#udpsocksend)
@@ -6206,7 +6207,8 @@ The resulting object `dict` has the following methods:
 * [flush_all](#ngxshareddictflush_all)
 * [flush_expired](#ngxshareddictflush_expired)
 * [get_keys](#ngxshareddictget_keys)
-* [stats](#ngxshareddictstats)
+* [capacity](#ngxshareddictcapacity)
+* [free_space](#ngxshareddictfree_space)
 
 All these methods are *atomic* operations, that is, safe from concurrent accesses from multiple nginx worker processes for the same `lua_shared_dict` zone.
 
@@ -6671,21 +6673,51 @@ This feature was first introduced in the `v0.7.3` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
-ngx.shared.DICT.stats
+ngx.shared.DICT.capacity
 ---------------------
-**syntax:** *free_page_bytes = ngx.shared.DICT:stats()*
+**syntax:** *capacity_bytes = ngx.shared.DICT:capacity()*
 
 **context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
 
 **requires:** `resty.core.shdict` or `resty.core`
 
-Retrieving the free pages size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
+the [lua_shared_dict](#lua_shared_dict) directive.
+
+
+Example:
+
+```lua
+
+ require "resty.core"
+
+ local cats = ngx.shared.cats
+ local capacity_bytes = cats:capacity()
+```
+
+This feature was first introduced in the `v0.10.11` release.
+
+**Note:** This method requires the `resty.core.shdict` or `resty.core` modules from the [lua-resty-core](https://github.com/openresty/lua-resty-core) library.
+
+See also [ngx.shared.DICT](#ngxshareddict).
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.free_space
+--------------------------
+**syntax:** *free_page_bytes = ngx.shared.DICT:free_space()*
+
+**context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
+
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the free page size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
 
 **Note:** The memory for ngx.shared.DICT is allocated via the nginx slab allocator which has each slot for
 data size ranges like ~8, 9~16, 17~32, ..., 1025~2048, 2048~ bytes. And pages are assigned to a slot if there
 is no room in already assigned pages for the slot.
 
-So even if the return value of the `stats` method is zero, there may be room in already assigned pages, so
+So even if the return value of the `free_space` method is zero, there may be room in already assigned pages, so
 you may successfully set a new key value pair to the shared dict without getting `true` for `forcieble` or
 non nil `err` from the `ngx.shared.DICT.set`.
 
@@ -6699,8 +6731,8 @@ Example:
 
  require "resty.core"
 
- local dogs = ngx.shared.cats
- local free_page_bytes = dogs:stats()
+ local cats = ngx.shared.cats
+ local free_page_bytes = cats:free_space()
 ```
 
 This feature was first introduced in the `v0.10.11` release.
--- a/README.markdown
+++ b/README.markdown
@@ -3196,6 +3196,7 @@ Nginx API for Lua
 * [ngx.shared.DICT.flush_all](#ngxshareddictflush_all)
 * [ngx.shared.DICT.flush_expired](#ngxshareddictflush_expired)
 * [ngx.shared.DICT.get_keys](#ngxshareddictget_keys)
+* [ngx.shared.DICT.stats](#ngxshareddictstats)
 * [ngx.socket.udp](#ngxsocketudp)
 * [udpsock:setpeername](#udpsocksetpeername)
 * [udpsock:send](#udpsocksend)
@@ -6205,6 +6206,7 @@ The resulting object `dict` has the following methods:
 * [flush_all](#ngxshareddictflush_all)
 * [flush_expired](#ngxshareddictflush_expired)
 * [get_keys](#ngxshareddictget_keys)
+* [stats](#ngxshareddictstats)
 
 All these methods are *atomic* operations, that is, safe from concurrent accesses from multiple nginx worker processes for the same `lua_shared_dict` zone.
 
@@ -6666,6 +6668,34 @@ By default, only the first 1024 keys (if any) are returned. When the `<max_count
 **CAUTION** Avoid calling this method on dictionaries with a very large number of keys as it may lock the dictionary for significant amount of time and block Nginx worker processes trying to access the dictionary.
 
 This feature was first introduced in the `v0.7.3` release.
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.shared.DICT.stats
+---------------------
+**syntax:** *used, total = ngx.shared.DICT:stats()*
+
+**context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
+
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the total used counts and total size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+Example:
+
+```lua
+
+ require "resty.core"
+
+ local cats = ngx.shared.cats
+ local used, total = cats:stats()
+```
+
+This feature was first introduced in the `v0.10.11` release.
+
+**Note:** This method requires the `resty.core.shdict` or `resty.core` modules from the [lua-resty-core](https://github.com/openresty/lua-resty-core) library.
+
+See also [ngx.shared.DICT](#ngxshareddict).
 
 [Back to TOC](#nginx-api-for-lua)
 
--- a/README.markdown
+++ b/README.markdown
@@ -6719,7 +6719,7 @@ data size ranges like ~8, 9~16, 17~32, ..., 1025~2048, 2048~ bytes. And pages ar
 is no room in already assigned pages for the slot.
 
 So even if the return value of the `free_space` method is zero, there may be room in already assigned pages, so
-you may successfully set a new key value pair to the shared dict without getting `true` for `forcieble` or
+you may successfully set a new key value pair to the shared dict without getting `true` for `forcible` or
 non nil `err` from the `ngx.shared.DICT.set`.
 
 On the other hand, if already assigned pages for a slot are full and a new key value pair is added to the
--- a/README.markdown
+++ b/README.markdown
@@ -6681,7 +6681,7 @@ ngx.shared.DICT.capacity
 
 **requires:** `resty.core.shdict` or `resty.core`
 
-Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
+Retrieves the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
 the [lua_shared_dict](#lua_shared_dict) directive.
 
 Example:
@@ -6712,7 +6712,7 @@ ngx.shared.DICT.free_space
 
 **requires:** `resty.core.shdict` or `resty.core`
 
-Retrieving the free page size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+Retrieves the free page size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
 
 **Note:** The memory for ngx.shared.DICT is allocated via the nginx slab allocator which has each slot for
 data size ranges like ~8, 9~16, 17~32, ..., 1025~2048, 2048~ bytes. And pages are assigned to a slot if there
--- a/README.markdown
+++ b/README.markdown
@@ -6707,6 +6707,8 @@ This feature was first introduced in the `v0.10.11` release.
 
 **Note:** This method requires the `resty.core.shdict` or `resty.core` modules from the [lua-resty-core](https://github.com/openresty/lua-resty-core) library.
 
+This feature requires at least nginx core version `1.11.7`.
+
 See also [ngx.shared.DICT](#ngxshareddict).
 
 [Back to TOC](#nginx-api-for-lua)
--- a/README.markdown
+++ b/README.markdown
@@ -6673,13 +6673,25 @@ This feature was first introduced in the `v0.7.3` release.
 
 ngx.shared.DICT.stats
 ---------------------
-**syntax:** *used, total = ngx.shared.DICT:stats()*
+**syntax:** *free_page_bytes = ngx.shared.DICT:stats()*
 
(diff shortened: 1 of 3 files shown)


thibaultcha:
Nice job! I was about to propose such an API as well in the upcoming days (although more based on the resty.core FFI API instead of ngx_lua CFunctions). This is definitely something we have pending use-case for as well!

src/ngx_http_lua_shdict.c
```diff
@@ -2846,4 +2850,72 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    if (zone == NULL) {
+        return luaL_error(L, "bad user data for the ngx_shm_zone_t pointer");
+    }
+
+    ctx = zone->data;
+
+    ngx_shmtx_lock(&ctx->shpool->mutex);
+
+    n_slots = ngx_pagesize_shift - ctx->shpool->min_shift;
+    lua_createtable(L, n_slots, 0);
```
agentzh: We cannot really hold the lock while calling these `lua_XXX` C API functions since the latter may introduce GC steps that may call `__gc` meta methods which in turn call into shdict methods, thus leading to deadlocks (we're about to retire all the existing CFunction-based Lua API functions in ngx_lua module to address similar issues).

agentzh: Also, it's better to allow the caller to supply this Lua table so that we do not have to create a brand new Lua table upon every invocation of this API function, saving dynamic allocation and GC overhead.

agentzh: We should implement this as an pure C function instead and make a FFI Lua wrapper around it.

agentzh:
@hnakamur Will you look into the issues mentioned above? Thanks!

I really appreciate this new API function, very useful indeed.

Let's do it in the FFI way instead. We're removing those old `lua_CFunction`-based Lua APIs now, dropping support for the standard Lua interpreter altogether.

hnakamur:
@agentzh Thanks for your review and comments.
I'm starting to change this to the FFI way.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,38 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    stat = ctx->shpool->stats;
+    slot_size = ctx->shpool->min_size;
+    for (i = 0; i < slot_count; i++) {
+        sum_used += stat->used;
+        sum_size += stat->used * slot_size;
+
+        stat++;
+        slot_size <<= 1;
+    }
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
```
agentzh: Style: need a blank line after the for statement block.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,38 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    size_t                       sum_used = 0;
+    size_t                       sum_size = 0;
+
+    ctx = zone->data;
+
+    ngx_shmtx_lock(&ctx->shpool->mutex);
+    slot_count = ngx_pagesize_shift - ctx->shpool->min_shift;
+    stat = ctx->shpool->stats;
+    slot_size = ctx->shpool->min_size;
+    for (i = 0; i < slot_count; i++) {
```
agentzh: Style: need a blank line before the for statement block.

hnakamur: done

src/ngx_http_lua_shdict.c
hnakamur: done

hnakamur:
BTW Travis builds for nginx 1.11.2 fail with the error `unknown type name ‘ngx_slab_stat_t’`.

```
gcc -c -I/opt/luajit21/include/luajit-2.1  -pipe  -O -W -Wall -Wpointer-arith -Wno-unused-parameter -Werror -g -I/opt/pcre/include -I/opt/ssl/include -DNDK_SET_VAR -DNDK_UPSTREAM_LIST -DNDK_SET_VAR  -I src/core -I src/event -I src/event/modules -I src/os/unix -I /home/travis/build/openresty/ndk-nginx-module/objs -I objs/addon/ndk -I /home/travis/build/openresty/lua-nginx-module/src/api -I /opt/drizzle/include/libdrizzle-1.0 -I objs -I src/http -I src/http/modules -I src/http/v2 -I /home/travis/build/openresty/ndk-nginx-module/src -I /home/travis/build/openresty/ndk-nginx-module/src -I /home/travis/build/openresty/ndk-nginx-module/objs -I objs/addon/ndk -I /opt/luajit21/include/luajit-2.1 \
		-o objs/addon/src/ngx_http_lua_api.o \
		/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_api.c
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_shdict.c: In function ‘ngx_http_lua_ffi_shdict_get_stats’:
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_shdict.c:2992:5: error: unknown type name ‘ngx_slab_stat_t’
     ngx_slab_stat_t             *stat;
     ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_shdict.c:3003:23: error: ‘ngx_slab_pool_t’ has no member named ‘stats’
     stat = ctx->shpool->stats;
                       ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_shdict.c:3007:25: error: request for member ‘used’ in something not a structure or union
         sum_used += stat->used;
                         ^
/home/travis/build/openresty/lua-nginx-module/src/ngx_http_lua_shdict.c:3008:25: error: request for member ‘used’ in something not a structure or union
         sum_size += stat->used * slot_size;
                         ^
make[1]: *** [objs/addon/src/ngx_http_lua_shdict.o] Error 1
```

Is it OK?

agentzh:
@hnakamur No, it's not OK, we should keep compatibility with older nginx core versions. If this feature needs more recent nginx cores, we should just omit this feature with older nginx cores with C macro protection.

hnakamur:
Alright, I've added a condition to compile ngx_http_lua_ffi_shdict_get_stats only when nginx version is greater than or equal to 1.11.7.
https://github.com/nginx/nginx/commit/9ccf719be6b922715fc6bdfabf24742eeec46ba4

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,41 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
 
     return NGX_OK;
 }
+
```
agentzh: Style: we need 2 blank lines to separate C function definitions.

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,41 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
...
+    ngx_slab_stat_t             *stat;
+    size_t                       slot_count;
+    size_t                       slot_size;
+    size_t                       i;
+    size_t                       sum_used = 0;
+    size_t                       sum_size = 0;
+
+    ctx = zone->data;
+
+    ngx_shmtx_lock(&ctx->shpool->mutex);
```
agentzh: Better add a blank line after `ngx_shmtx_lock()` to make this more outstanding.

src/ngx_http_lua_shdict.c
hnakamur: done

src/ngx_http_lua_shdict.c
hnakamur: done

hnakamur:
I added document for ngx.shared.DICT.stats in response to https://github.com/openresty/lua-resty-core/pull/141#discussion_r140663493

README.markdown
```diff
@@ -6669,6 +6671,46 @@ This feature was first introduced in the `v0.7.3` release.
...
+ require "resty.core"
+
+ local dogs = ngx.shared.cats
+ local free_page_bytes = dogs:stats()
+```
+
+This feature was first introduced in the `v0.10.11` release.
+
+**Note:** This method requires the `resty.core.shdict` or `resty.core` modules from the [lua-resty-core](https://github.com/openresty/lua-resty-core) library.
+
```
agentzh: Should we document the minimum nginx core version required here?

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,24 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
...
+
+    *free_page_bytes = ctx->shpool->pfree * ngx_pagesize;
+
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
+}
+#endif /* nginx_version >= 1011007 */
 #endif /* NGX_LUA_NO_FFI_API */
```
agentzh: It would be nicer to add a blank line before these 2 `#endif` lines.

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,24 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
 
     return NGX_OK;
 }
+
+
+#if nginx_version >= 1011007
```
agentzh: Style: better add 4 spaces between `#` and `if` to indicate the nested level. Let's also do this for the corresponding `#endif` below?

README.markdown
```diff
@@ -6669,6 +6671,46 @@ This feature was first introduced in the `v0.7.3` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.shared.DICT.stats
+---------------------
+**syntax:** *free_page_bytes = ngx.shared.DICT:stats()*
```
agentzh: Hmm, I thought you would return the total size here as well. Otherwise the method name `stats` is just too general for this.

src/ngx_http_lua_shdict.c
hnakamur: done

src/ngx_http_lua_shdict.c
hnakamur: done

README.markdown
hnakamur: done

README.markdown
hnakamur: I think the total size is fun to watch but it's not useful for alerting the memory exhaustion of a shared dictionary.
So I'd like to avoid wasting to calculate unused metric here.

If the total size really needed, I'd like a separate method or make it optional like
`ngx.shared.DICT:stats(want_total_size)`

```
free_page_bytes = ngx.shared.DICT:stats()
free_page_bytes, total_size = ngx.shared.DICT:stats(true)
```

How about that?

agentzh: @hnakamur The boolean argument is not readable at all. How about 2 separate methods, `free_space()` and `capacity()`?

hnakamur: I agree on 2 separate methods and the name `free_space`, but I think `capacity` is misleading because it reminds the limit of the size of the shared dict, not the current used bytes.
How about `usage()` or `used_bytes()` ?

hnakamur: I've read code again to know how to implement the function for the total size and I realized I can only get an approximate value.

For a key value pair which belongs to a slot for size less than or equal to 2048 bytes, the total size used can be retrieved from `total` field of `ngx_slab_stat_t` struct. https://github.com/nginx/nginx/blob/release-1.13.5/src/core/ngx_slab.h#L26

However, for key value pair with management data whose size is greater than 2048 bytes, pages are assigned at https://github.com/nginx/nginx/blob/release-1.13.5/src/core/ngx_slab.c#L191-L206
and the request size is not recorded in `total` field of `ngx_slab_stat_t` struct. 
Instead, the key length and value length are recorded in `key_len` and `value_len` fields of `ngx_http_lua_shdict_node_t` struct.
https://github.com/openresty/lua-nginx-module/blob/97fbeb0bef1aa85e758210d58063376de8eaed31/src/ngx_http_lua_shdict.c#L1211-L1227

To read the `key_len` and `value_len` fields of `ngx_http_lua_shdict_node_t`, I think we need to iterate through all key value pairs in the shared dict, so it becomes a costly operation.
We would need the same caution as in `get_keys` method.
https://github.com/openresty/lua-nginx-module#ngxshareddictget_keys

But I'll avoid to use the `usage` method if it may lock the shared dictionary for significant amount of time and block Nginx worker processes trying to access the dictionary.

What do you think?

agentzh: @hnakamur capacity indeed means the total size specified in the `lua_shared_dict` config directive.

hnakamur: @agentzh Oh I see. I've added `capacity` method and updated tests and docs.
Could you review it?
Thanks!

README.markdown
```diff
@@ -6669,6 +6673,78 @@ This feature was first introduced in the `v0.7.3` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.shared.DICT.capacity
+---------------------
```
agentzh: Better align `---` with the text in the previous line.

README.markdown
```diff
@@ -6669,6 +6673,78 @@ This feature was first introduced in the `v0.7.3` release.
...
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
+the [lua_shared_dict](#lua_shared_dict) directive.
+
+
+Example:
+
+```lua
+
```
agentzh: Extra blank line here?

Please fix other similar places.

README.markdown
```diff
@@ -6669,6 +6673,78 @@ This feature was first introduced in the `v0.7.3` release.
...
+**syntax:** *capacity_bytes = ngx.shared.DICT:capacity()*
+
+**context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
+
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
+the [lua_shared_dict](#lua_shared_dict) directive.
+
+
```
agentzh: Extra blank line here?

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,34 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
...
+void
+ngx_http_lua_ffi_shdict_capacity(ngx_shm_zone_t *zone,
+    size_t *capacity_bytes)
+{
+    *capacity_bytes = zone->shm.size;
+}
+
+
+#    if nginx_version >= 1011007
+void
```
agentzh: Can we return the size directly without using pointers? This would make the FFI side code simpler.

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,34 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
 
     return NGX_OK;
 }
+
+
+void
+ngx_http_lua_ffi_shdict_capacity(ngx_shm_zone_t *zone,
+    size_t *capacity_bytes)
```
agentzh: Better return the `size_t` value directly without using an extra output argument through pointers.

README.markdown
hnakamur: done

README.markdown
hnakamur: Removed the extra blank line.

README.markdown
hnakamur: I did put this extra blank line here similar to other places like `ngx.shaed.DICT.ttl`
https://raw.githubusercontent.com/openresty/lua-nginx-module/97fbeb0bef1aa85e758210d58063376de8eaed31/README.markdown

src/ngx_http_lua_shdict.c
hnakamur: done

src/ngx_http_lua_shdict.c
hnakamur: done

README.markdown
thibaultcha: This line is inserted by the `wiki2markdown.pl` script. Until recently (and probably still the case), the README is not to be edited directly. See the note at the beginning:
```
<!---
-->
```

agentzh: @thibaultcha I see. Never mind then.

hnakamur: @thibaultcha Thanks for your comment. I didn't notice that comment. I'll update doc/HttpLuaModule.wiki and regenerate README.markdown.

hnakamur: I've updated doc/HttpLuaModule.wiki, regenerated README.markdown and confirmed there is no difference between the generated version and the README.markdown in the previous commit.

README.markdown
```diff
@@ -6669,6 +6673,79 @@ This feature was first introduced in the `v0.7.3` release.
...
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the free page size in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict).
+
+**Note:** The memory for ngx.shared.DICT is allocated via the nginx slab allocator which has each slot for
+data size ranges like ~8, 9~16, 17~32, ..., 1025~2048, 2048~ bytes. And pages are assigned to a slot if there
+is no room in already assigned pages for the slot.
+
+So even if the return value of the `free_space` method is zero, there may be room in already assigned pages, so
+you may successfully set a new key value pair to the shared dict without getting `true` for `forcieble` or
```
thibaultcha: typo: forcible

README.markdown
```diff
@@ -6669,6 +6673,79 @@ This feature was first introduced in the `v0.7.3` release.
...
 
+ngx.shared.DICT.capacity
+------------------------
+**syntax:** *capacity_bytes = ngx.shared.DICT:capacity()*
+
+**context:** *init_by_lua&#42;, set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;*
+
+**requires:** `resty.core.shdict` or `resty.core`
+
+Retrieving the capacity in bytes for the shm-based dictionary [ngx.shared.DICT](#ngxshareddict) declared with
```
thibaultcha: Just a note: all other methods in this document are not documented with a present continuous form. Most (if not all) methods actually use the third-person form:
```
Retrieves the capacity in bytes [...]
```

hnakamur: Thanks for a note. I followed `Reterieving` in [ngx.shared.DICT.get](https://github.com/openresty/lua-nginx-module#ngxshareddictget).
Should I change all `Retrieving` with `Retrieves` in all of ngx.shared.DICT methods?

thibaultcha: Oh, hmm. That is unfortunate (better being consistent if possible imho), but no, I don't think we should be updating any other method's description here, it isn't the scope of this PR. If present continuous form is judged acceptable, then please ignore my comment!

hnakamur: Alright, I've changed `Retrieving` to `Retrieves` in doc of `capacity` and `free_space` in `ngx.shared.DICT`.

doc/HttpLuaModule.wiki
```diff
@@ -5598,6 +5600,71 @@ By default, only the first 1024 keys (if any) are returned. When the <code><max_
...
+
+'''requires:''' <code>resty.core.shdict</code> or <code>resty.core</code>
+
+Retrieves the capacity in bytes for the shm-based dictionary [[#ngx.shared.DICT|ngx.shared.DICT]] declared with
+the [[#lua_shared_dict|lua_shared_dict]] directive.
+
+Example:
+
+<geshi lang="lua">
+    require "resty.core"
```
agentzh: Better use `resty.core.shdict` here instead of `resty.core`.

doc/HttpLuaModule.wiki
```diff
@@ -5598,6 +5600,71 @@ By default, only the first 1024 keys (if any) are returned. When the <code><max_
...
+non nil <code>err</code> from the <code>ngx.shared.DICT.set</code>.
+
+On the other hand, if already assigned pages for a slot are full and a new key value pair is added to the
+slot and there is no free page, you may get <code>true</code> for <code>forcible</code> or non nil <code>err</code> from the
+<code>ngx.shared.DICT.set</code> method.
+
+Example:
+
+<geshi lang="lua">
+    require "resty.core"
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
```diff
@@ -2983,6 +2983,35 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
...
+    return zone->shm.size;
+}
+
+
+#    if nginx_version >= 1011007
+size_t
+ngx_http_lua_ffi_shdict_free_space(ngx_shm_zone_t *zone)
+{
+    ngx_http_lua_shdict_ctx_t   *ctx;
+    size_t                       free_page_bytes;
```
agentzh: For such a short-lived local auto variable, we'd better use a short name like `bytes`?

agentzh:
Merged with the following extra patch:

```diff
diff --git a/doc/HttpLuaModule.wiki b/doc/HttpLuaModule.wiki
index ef59e648..41273652 100644
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -5613,7 +5613,7 @@ the [[#lua_shared_dict|lua_shared_dict]] directive.
 Example:

 <geshi lang="lua">
-    require "resty.core"
+    require "resty.core.shdict"

     local cats = ngx.shared.cats
     local capacity_bytes = cats:capacity()
@@ -5651,7 +5651,7 @@ slot and there is no free page, you may get <code>true</code> for <code>forcible
 Example:

 <geshi lang="lua">
-    require "resty.core"
+    require "resty.core.shdict"

     local cats = ngx.shared.cats
     local free_page_bytes = cats:free_space()
diff --git a/src/ngx_http_lua_shdict.c b/src/ngx_http_lua_shdict.c
index 91901147..5b48eb4b 100644
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -2996,18 +2996,16 @@ ngx_http_lua_ffi_shdict_capacity(ngx_shm_zone_t *zone)
 size_t
 ngx_http_lua_ffi_shdict_free_space(ngx_shm_zone_t *zone)
 {
+    size_t                       bytes;
     ngx_http_lua_shdict_ctx_t   *ctx;
-    size_t                       free_page_bytes;

     ctx = zone->data;

     ngx_shmtx_lock(&ctx->shpool->mutex);
-
-    free_page_bytes = ctx->shpool->pfree * ngx_pagesize;
-
+    bytes = ctx->shpool->pfree * ngx_pagesize;
     ngx_shmtx_unlock(&ctx->shpool->mutex);

-    return free_page_bytes;
+    return bytes;
 }
 #    endif /* nginx_version >= 1011007 */
```

hnakamur:
Thanks for your review and merging!

PR #1150, opened 2017-08-31
feature: added pure C functions for shdict:ttl() and shdict:expire() APIs

thibaultcha:

Sister PR for openresty/lua-resty-core#140

### DIFF
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -2898,6 +2898,7 @@ ngx_http_lua_ffi_shdict_get_ttl(ngx_shm_zone_t *zone, u_char *key,
 {
     uint32_t                     hash;
     uint64_t                     now;
+    uint64_t                     expires;
     ngx_int_t                    rc;
     ngx_time_t                  *tp;
     ngx_http_lua_shdict_ctx_t   *ctx;
@@ -2914,22 +2915,26 @@ ngx_http_lua_ffi_shdict_get_ttl(ngx_shm_zone_t *zone, u_char *key,
 
     rc = ngx_http_lua_shdict_peek(zone, hash, key, key_len, &sd);
 
-    ngx_shmtx_unlock(&ctx->shpool->mutex);
-
     if (rc == NGX_DECLINED) {
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+
         return NGX_DECLINED;
     }
 
     /* rc == NGX_OK */
 
-    if (sd->expires == 0) {
+    expires = sd->expires;
+
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+    if (expires == 0) {
         return 0;
     }
 
     tp = ngx_timeofday();
     now = (uint64_t) tp->sec * 1000 + tp->msec;
 
-    return sd->expires - now;
+    return expires - now;
 }
 
 
@@ -2954,9 +2959,9 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
 
     rc = ngx_http_lua_shdict_peek(zone, hash, key, key_len, &sd);
 
-    ngx_shmtx_unlock(&ctx->shpool->mutex);
-
     if (rc == NGX_DECLINED) {
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+
         return NGX_DECLINED;
     }
 
@@ -2970,6 +2975,8 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
     } else {
         sd->expires = 0;
     }
+
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
 
     return NGX_OK;
 }
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -2944,12 +2944,16 @@ ngx_http_lua_ffi_shdict_set_expire(ngx_shm_zone_t *zone, u_char *key,
 {
     uint32_t                     hash;
     ngx_int_t                    rc;
-    ngx_time_t                  *tp;
+    ngx_time_t                  *tp = NULL;
     ngx_http_lua_shdict_ctx_t   *ctx;
     ngx_http_lua_shdict_node_t  *sd;
 
     if (zone == NULL) {
         return NGX_ERROR;
+    }
+
+    if (exptime > 0) {
+        tp = ngx_timeofday();
     }
 
     ctx = zone->data;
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,124 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
 
     return NGX_OK;
 }
+
+
+static ngx_int_t
+ngx_http_lua_shdict_peek(ngx_shm_zone_t *shm_zone, ngx_uint_t hash,
```
thibaultcha: The idea with this `peek` function is to not consider the node "touched" and avoid the node being put at the queue's head. Open to feedback.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,124 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    ngx_int_t                    rc;
+    ngx_time_t                  *tp;
+    ngx_http_lua_shdict_node_t  *sd;
+
+    if (zone == NULL) {
+        return NGX_ERROR;
+    }
+
+    hash = ngx_crc32_short(key, key_len);
+
```
agentzh: We need to lock here otherwise the rbtree may be balancing (initiated by another worker process) when we traverse the tree here, leading to race conditions and invalid memory accesses.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,124 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
thibaultcha: Indeed; fixed!

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,136 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+    if (rc == NGX_DECLINED) {
+        return NGX_DECLINED;
+    }
+
+    /* rc == NGX_OK */
+
+    if (sd->expires == 0) {
```
agentzh: We should not use the shm memory after the lock is released. This is unsafe.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,136 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+        return 0;
+    }
+
+    tp = ngx_timeofday();
+    now = (uint64_t) tp->sec * 1000 + tp->msec;
+
+    return sd->expires - now;
```
agentzh: Ditto. This is should also be in the critical region.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,136 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+
+    /* rc == NGX_OK */
+
+    if (exptime > 0) {
+        tp = ngx_timeofday();
+        sd->expires = (uint64_t) tp->sec * 1000 + tp->msec
+                      + (uint64_t) exptime;
+
+    } else {
+        sd->expires = 0;
```
agentzh: Ditto.

src/ngx_http_lua_shdict.c
thibaultcha: Oh, sorry. That makes perfect sense. Sorry about the back and forth here; I'm still new to the shm part of ngx_lua!

thibaultcha:
@agentzh I pushed a new commit (hoping it'd help reviewing) for the proper unlocking pattern! Let me know.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,144 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    rc = ngx_http_lua_shdict_peek(zone, hash, key, key_len, &sd);
+
+    if (rc == NGX_DECLINED) {
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+        return NGX_DECLINED;
+    }
+
+    /* rc == NGX_OK */
+
```
agentzh: Why don't you just save the value of `sd->expires` in a local C variable so that we can release the lock as soon as possible?

thibaultcha: The thought did cross my mind... But I didn't think it would have made a big difference. I was wrong!

thibaultcha: Pushed a new version of this second commit which does exactly that.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,143 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    if (exptime > 0) {
+        tp = ngx_timeofday();
```
agentzh: Can we move this before the `ngx_shmtx_lock` call? It's a bit wasteful for the other branch but this way we make the critical section a little bit smaller (maybe saving one CPU cycle or so).

thibaultcha: Crossed my mind as well, but then I saw other locked branches in this module calling `ngx_timeofday()` as well. Happy to move both calls in these functions - on it.

agentzh: @thibaultcha Great. Thanks!

thibaultcha: @agentzh Done

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,146 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    ngx_int_t                    rc;
+    ngx_time_t                  *tp;
+    ngx_http_lua_shdict_ctx_t   *ctx;
+    ngx_http_lua_shdict_node_t  *sd;
+
+    if (zone == NULL) {
+        return NGX_ERROR;
+    }
+
+    tp = ngx_timeofday();
```
agentzh: This can be moved after the `if (expires == 0)` branch since it's not in the critical region anyway.

src/ngx_http_lua_shdict.c
```diff
@@ -2843,6 +2843,146 @@ ngx_http_lua_ffi_shdict_flush_all(ngx_shm_zone_t *zone)
...
+    /* rc == NGX_OK */
+
+    expires = sd->expires;
+
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+    if (expires == 0) {
+        return 0;
+    }
+    now = (uint64_t) tp->sec * 1000 + tp->msec;
```
agentzh: Style: need a blank line after the if statement block.

src/ngx_http_lua_shdict.c
thibaultcha: Uh, yeah... Sorry.

src/ngx_http_lua_shdict.c
thibaultcha: fat-fingered vim, sorry

PR #1151, opened 2017-09-11
bugfix: ARM64: ensure that only the low 47-bit values of the global v…

agentzh:
…ariables' addresses are used as light userdata; otherwise it would lead to the "bad light userdata pointer" luajit error when luajit uses the GC64 mode.


agentzh:
Seems like the problem is much bigger than that. All the static data section has exceeded 47-bit on many ARM64 systems.

Issue #1066, opened 2017-05-11
set_by_lua_block breaks config dump (-T switch)


lonerr:
Environment and build options:
```
# uname -a
FreeBSD armor.4rt.ru 11.0-RELEASE-p8 FreeBSD 11.0-RELEASE-p8 #0: Wed Feb 22 06:12:04 UTC 2017     root@amd64-builder.daemonology.net:/usr/obj/usr/src/sys/GENERIC  amd64

# nginx -V
nginx version: nginx/1.11.10
built with OpenSSL 1.0.2k  26 Jan 2017
TLS SNI support enabled
[...]

PR #1156, opened 2017-09-19
bugfix: set_by_lua_block breaks config dump (-T switch)

lonerr:
My colleague kindly provided this fix for issue #1066.


agentzh:
@lonerr It affects all those `*_by_lua_block` directives, not just `set_by_lua_block`. Thanks for the patch! Already merged.

PR #1160, opened 2017-09-23
feature: for ssl.set_http_version API

vislee:

### DIFF
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -975,11 +975,15 @@ ngx_http_lua_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
 
 #   ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
         SSL_CTX_set_alpn_select_cb(sscf->ssl.ctx, ngx_http_lua_ssl_alpn_select, NULL);
+        ngx_log_error(NGX_LOG_DEBUG, cf->log, 0,
+                      "LUA set alpn callback");
 #   endif
 
 #   ifdef TLSEXT_TYPE_next_proto_neg
         SSL_CTX_set_next_protos_advertised_cb(sscf->ssl.ctx,
                                               ngx_http_lua_ssl_npn_advertised, NULL);
+        ngx_log_error(NGX_LOG_DEBUG, cf->log, 0,
+                      "LUA set npn callback");
 #   endif
 
 #endif
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -972,6 +972,16 @@ ngx_http_lua_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
 
 #   endif
 
+
+#   ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+        SSL_CTX_set_alpn_select_cb(sscf->ssl.ctx, ngx_http_lua_ssl_alpn_select, NULL);
+#   endif
+
+#   ifdef TLSEXT_TYPE_next_proto_neg
+        SSL_CTX_set_next_protos_advertised_cb(sscf->ssl.ctx,
+                                              ngx_http_lua_ssl_npn_advertised, NULL);
+#   endif
+
 #endif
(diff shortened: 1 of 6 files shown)


src/ngx_http_lua_ssl_certby.c
```diff
@@ -524,6 +531,123 @@ ngx_http_lua_ssl_cert_by_chunk(lua_State *L, ngx_http_request_t *r)
...
+    }
+    }
+#endif
+
+    *out = (unsigned char *) NGX_HTTP_NPN_ADVERTISE;
+    *outlen = sizeof(NGX_HTTP_NPN_ADVERTISE) - 1;
+
+    #if (NGX_DEBUG)
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, c->log, 0,
+                   "LUA SSL NPN selected: %*s", (size_t) *outlen, *out);
```
agentzh: Lua is not an acronym: https://www.lua.org/about.html#name

agentzh:
@vislee What problem does this PR try to solve? I see no documentation nor test cases in this PR. Will you elaborate?

src/ngx_http_lua_ssl.h
```diff
@@ -26,6 +26,7 @@ typedef struct {
 
     int                      exit_code;  /* exit code for openssl's
                                             set_cert_cb callback */
+    int                      http_version;
```
agentzh: You just want to allow 1 and 2 here? Better be explicit. We do have protocol versions like 0.9, 1.0, and 1.1 also in the nginx core. If you just want to signify whether it is HTTP/2, then we should just make this a boolean field with a name like `is_http2`. This way, we can also make this field just take 1 bit.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1236,6 +1360,40 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
 }
 
 
+int
+ngx_http_lua_ffi_set_httpv(ngx_http_request_t *r,
```
agentzh: The `httpv` abbreviation looks weird. How about the name `ngx_http_lua_ffi_use_http2` to make our intention clearer here?

agentzh:
@vislee I just had a closer look and it seems that you are adding a Lua API for dynamically toggling the use of the HTTP/2 protocol in the ALPN phase on the connection level?

src/ngx_http_lua_ssl_session_storeby.c
```diff
@@ -252,6 +252,11 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
     cctx->session_id.data = sess->session_id;
     cctx->session_id.len = sess->session_id_length;
     cctx->done = 0;
+#if (NGX_HTTP_V2)
```
lziest: I am not quite sure why storing session will touch http2, alpn and session resumption are orthogonal.

agentzh: Yeah, this is definitely wrong. session storing happens *after* a full SSL handshake and thus also ALPN selection and ssl set_cert callback.

src/ngx_http_lua_module.c
```diff
@@ -972,6 +972,16 @@ ngx_http_lua_merge_srv_conf(ngx_conf_t *cf, void *parent, void *child)
 
 #   endif
 
+
+#   ifdef TLSEXT_TYPE_application_layer_protocol_negotiation
+        SSL_CTX_set_alpn_select_cb(sscf->ssl.ctx, ngx_http_lua_ssl_alpn_select, NULL);
```
agentzh: Hmm, maybe we should do `alpn_by_lua_block` and `alpn_by_lua_file` directives by allowing running Lua directly in the APLN and NPN hooks? This way we can ensure our application level protocol selection works on all the code paths (ssl session id and tls session ticket) in case of SSL session resumption.

vislee: Sorry for my fault. I don't have the test session. I'm starting to change to it. You can close related PR.

PR #1162, opened 2017-09-25
bugfix: the connection might be closed when some handlers using fake …

guanglinlv:
…connection are involved (like ssl_certficate_by_lua*, ssl_session_fetch_by_lua*, ssl_session_store_by_lua*).

fix the issue reported at [ngx_http_lua_ssl_cert_handler crash](https://groups.google.com/forum/#!topic/openresty/eLd8Q7qe_jk).

it's a little difficult to add a test case to cover this. any idea?

@agentzh 

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -186,7 +186,6 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 {
     lua_State                       *L;
     ngx_int_t                        rc;
-    ngx_uint_t                       reusable;
     ngx_connection_t                *c, *fc;
     ngx_http_request_t              *r = NULL;
     ngx_pool_cleanup_t              *cln;
@@ -198,7 +197,7 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 
     c = ngx_ssl_get_connection(ssl_conn);
 
-    dd("c = %p", c);
+    dd("c = %p, reusable = %d", c, (int) c->reusable);
 
     cctx = ngx_http_lua_ssl_get_ctx(c->ssl->connection);
 
@@ -221,19 +220,11 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 
     dd("first time");
 
-    hc = c->data;
-    reusable = c->reusable;
+    ngx_reusable_connection(c, 0);
 
-    if (reusable) {
-        ngx_reusable_connection(c, 0);
-    }
+    hc = c->data;
 
     fc = ngx_http_lua_create_fake_connection(NULL);
-
-    if (reusable) {
-        ngx_reusable_connection(c, 1);
-    }
-
     if (fc == NULL) {
         goto failed;
     }
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -186,6 +186,7 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 {
     lua_State                       *L;
     ngx_int_t                        rc;
+    ngx_uint_t                       reusable;
     ngx_connection_t                *c, *fc;
     ngx_http_request_t              *r = NULL;
     ngx_pool_cleanup_t              *cln;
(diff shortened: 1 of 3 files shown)


agentzh:
@guanglinlv How about the following patch instead?

```diff
diff --git a/src/ngx_http_lua_ssl_certby.c b/src/ngx_http_lua_ssl_certby.c
index c3591d1e..242724b8 100644
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -197,7 +197,7 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)

     c = ngx_ssl_get_connection(ssl_conn);

-    dd("c = %p", c);
+    dd("c = %p, reusable = %d", c, (int) c->reusable);

     cctx = ngx_http_lua_ssl_get_ctx(c->ssl->connection);

@@ -220,6 +220,8 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)

     dd("first time");

+    ngx_reusable_connection(c, 0);
+
     hc = c->data;

     fc = ngx_http_lua_create_fake_connection(NULL);
diff --git a/src/ngx_http_lua_ssl_session_fetchby.c b/src/ngx_http_lua_ssl_session_fetchby.c
index 556b7320..a5445626 100644
--- a/src/ngx_http_lua_ssl_session_fetchby.c
+++ b/src/ngx_http_lua_ssl_session_fetchby.c
@@ -191,7 +191,7 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn, u_char *id,

     c = ngx_ssl_get_connection(ssl_conn);

-    dd("c = %p", c);
+    dd("c = %p, reusable = %d", c, (int) c->reusable);

     cctx = ngx_http_lua_ssl_get_ctx(c->ssl->connection);

@@ -224,6 +224,8 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn, u_char *id,

     dd("first time");

+    ngx_reusable_connection(c, 0);
+
     hc = c->data;

     fc = ngx_http_lua_create_fake_connection(NULL);
diff --git a/src/ngx_http_lua_ssl_session_storeby.c b/src/ngx_http_lua_ssl_session_storeby.c
index bae8273d..60ef06b7 100644
--- a/src/ngx_http_lua_ssl_session_storeby.c
+++ b/src/ngx_http_lua_ssl_session_storeby.c
@@ -183,7 +183,7 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,

     c = ngx_ssl_get_connection(ssl_conn);

-    dd("c = %p", c);
+    dd("c = %p, reusable = %d", c, (int) c->reusable);

     cctx = ngx_http_lua_ssl_get_ctx(c->ssl->connection);
```

agentzh:
@guanglinlv I don't think we need to do this in the `ssl_session_store_by_lua*` context since it is invoked after the SSL handshake is completed, which means that the nginx core has already marked the connection as not reusable at that point.

agentzh:
@guanglinlv Thank you very much for the report and patch!

guanglinlv:
@agentzh I think it should be better to let the connection reusable after creating fake connection.  I'm not sure about `ssl_session_store_by_lua*` context, because i have not use it yet.

agentzh:
@guanglinlv Marking it as non-reusable after creating the fake connection does not look quite right to me since the fake connection *may* drain and hijack the real connection for the same context, which will be very bad.

guanglinlv:
@agentzh You are right, we just need marking the connection non-reusable before creating the fake connection. I will commit it again following your suggestion.

agentzh:
@guanglinlv You tried my patch above and it works for you, right?

agentzh:
@guanglinlv It's indeed tricky to write test cases for this.

guanglinlv:
@agentzh I had test it about 3 hours with the maximum concurrency connections that the nginx can handled. the problem is not repeat. I will keep a long time pressure to check it.  BTW, any idea about test case?

agentzh:
@guanglinlv Thanks for the test feedback! Regarding to the regression tests, I ended up inspecting the nginx debug logs in the corresponding test cases. See commit 97fbeb0bef for details.

guanglinlv:
@agentzh Nginx with this patch is running more than ten hours, the problem never occurs once again.

agentzh:
@guanglinlv Cool!

PR #1165, opened 2017-09-29
always nul-terminate error buffer

kahing:
errbuf is allocated using base.get_string_buf which reuses buffers,
not nul-terminating error buffer means we could be leaking random
garbage in error messages

### DIFF
--- a/src/ngx_http_lua_variable.c
+++ b/src/ngx_http_lua_variable.c
@@ -364,7 +364,7 @@ ngx_http_lua_ffi_var_get(ngx_http_request_t *r, u_char *name_data,
 int
 ngx_http_lua_ffi_var_set(ngx_http_request_t *r, u_char *name_data,
     size_t name_len, u_char *lowcase_buf, u_char *value, size_t value_len,
-    u_char *errbuf, size_t errlen)
+    u_char *errbuf, size_t *errlen)
 {
     u_char                      *p;
     ngx_uint_t                   hash;
@@ -373,12 +373,14 @@ ngx_http_lua_ffi_var_set(ngx_http_request_t *r, u_char *name_data,
     ngx_http_core_main_conf_t   *cmcf;
 
     if (r == NULL) {
-        ngx_snprintf(errbuf, errlen, "no request object found");
+        *errlen = ngx_snprintf(errbuf, *errlen, "no request object found")
+            - errbuf;
         return NGX_ERROR;
     }
 
     if ((r)->connection->fd == (ngx_socket_t) -1) {
-        ngx_snprintf(errbuf, errlen, "API disabled in the current context");
+        *errlen = ngx_snprintf(errbuf, *errlen,
+                               "API disabled in the current context") - errbuf;
         return NGX_ERROR;
     }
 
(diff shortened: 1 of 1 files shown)


agentzh:
@kahing I think we'd better output the actual error message length instead of using NULL-terminated strings here. Will you change the C API accordingly please?

BTW, it is "NULL-terminated", not "nul-terminate". The latter is not proper English.

kahing:
NUL is the ascii character and the PR was actively '\0' terminating the string so IMO it's an acceptable usage. Anyhow, I've changed it to return the error length and consume it from lua side.

agentzh:
@kahing Well, see https://en.wikipedia.org/wiki/Null-terminated_string It's a term.

agentzh:
@kahing In general, we should avoid using C strings. The existing error string handling in the variable API indeed deserves fixing.

kahing:
happy to do further amends to the PR if you want to give me some pointers.

agentzh:
@kahing You can see `lib/resty/core/exit.lua` for a good example. Thanks!

kahing:
I can't find an `exit.lua` in this repo nor in openresty. Also, this is pretty low priority internally so it may take me a while.

agentzh:
@kahing It is here:

https://github.com/openresty/lua-resty-core/blob/master/lib/resty/core/exit.lua

kahing:
ah I see the confusion. I made the change there already but forgot to push a PR to that repo, will do that now

kahing:
updates?

src/ngx_http_lua_variable.c
```diff
@@ -395,8 +398,9 @@ ngx_http_lua_ffi_var_set(ngx_http_request_t *r, u_char *name_data,
     if (v) {
         if (!(v->flags & NGX_HTTP_VAR_CHANGEABLE)) {
             dd("variable not changeable");
-            ngx_snprintf(errbuf, errlen, "variable \"%*s\" not changeable",
-                         name_len, lowcase_buf);
+            *errlen = ngx_snprintf(errbuf, *errlen,
+                         "variable \"%*s\" not changeable",
```
agentzh: Style: the continued lines should be aligned up vertically with the first argument of the function call in the previous line. See the rest of the code base for such examples. Please also fix other similar places. Thank you!

kahing:
amended commit for spacing

src/ngx_http_lua_variable.c
```diff
@@ -364,21 +364,24 @@ ngx_http_lua_ffi_var_get(ngx_http_request_t *r, u_char *name_data,
...
-    u_char *errbuf, size_t errlen)
+    u_char *errbuf, size_t *errlen)
 {
     u_char                      *p;
     ngx_uint_t                   hash;
     ngx_http_variable_t         *v;
     ngx_http_variable_value_t   *vv;
     ngx_http_core_main_conf_t   *cmcf;
 
+    /* we don't return actual errlen so need to NUL-terminate it */
```
agentzh: I'm afraid this comment is now obsolete since we do return the actual errmsg length.

kahing: removed

kahing:
spacing fixed

agentzh:
@kahing Committed as commit 81e140c59 with some minor coding style fixes. Thanks for your contribution.

PR #1178, opened 2017-10-30
feature: Set TCP congestion control for each request connection

blacktear23:
This feature can let user set each request connection's TCP congestion control. In some use case choose a good TCP congestion can make response faster than kernel's default TCP congestion.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3138,7 +3138,6 @@ Nginx API for Lua
 * [ngx.req.append_body](#ngxreqappend_body)
 * [ngx.req.finish_body](#ngxreqfinish_body)
 * [ngx.req.socket](#ngxreqsocket)
-* [ngx.req.set_cong](#ngxreqset_cong)
 * [ngx.exec](#ngxexec)
 * [ngx.redirect](#ngxredirect)
 * [ngx.send_headers](#ngxsend_headers)
@@ -5008,22 +5007,6 @@ When the `raw` argument is `true`, it is required that no pending data from any 
 You can use the "raw request socket" returned by `ngx.req.socket(true)` to implement fancy protocols like [WebSocket](http://en.wikipedia.org/wiki/WebSocket), or just emit your own raw HTTP response header or body data. You can refer to the [lua-resty-websocket library](https://github.com/openresty/lua-resty-websocket) for a real world example.
 
 This function was first introduced in the `v0.5.0rc1` release.
-
-[Back to TOC](#nginx-api-for-lua)
-
-ngx.req.set\_cong
------------------
-
-**syntax:** ngx.req.set_cong(cong_name)*
-
-**context:** *access_by_lua&#42;, content_by_lua&#42;*
-
-Set TCP congestion control algorithm for each request connection. It only works on Linux platform and if the target TCP congestion does not exists, it will generate a warning log. Below is Linux kernel supported TCP congestion control algorithm.
-
-* bic kernel version between 2.6.8 and 2.6.18
-* cubic kernel version >= 2.6.19
-* reno
-* bbr kernel version >= 4.9
 
 [Back to TOC](#nginx-api-for-lua)
 
--- a/README.markdown
+++ b/README.markdown
@@ -5018,7 +5018,12 @@ ngx.req.set\_cong
 
 **context:** *access_by_lua&#42;, content_by_lua&#42;*
 
-Set TCP congestion control for each request connection. It only works on Linux platform and if target TCP congestion is not exists, it will generate a warning log.
+Set TCP congestion control algorithm for each request connection. It only works on Linux platform and if the target TCP congestion does not exists, it will generate a warning log. Below is Linux kernel supported TCP congestion control algorithm.
+
+* bic kernel version between 2.6.8 and 2.6.18
+* cubic kernel version >= 2.6.19
+* reno
+* bbr kernel version >= 4.9
 
 [Back to TOC](#nginx-api-for-lua)
 
--- a/README.markdown
+++ b/README.markdown
@@ -3138,6 +3138,7 @@ Nginx API for Lua
 * [ngx.req.append_body](#ngxreqappend_body)
 * [ngx.req.finish_body](#ngxreqfinish_body)
 * [ngx.req.socket](#ngxreqsocket)
+* [ngx.req.set_cong](#ngxreqset_cong)
 * [ngx.exec](#ngxexec)
 * [ngx.redirect](#ngxredirect)
 * [ngx.send_headers](#ngxsend_headers)
@@ -5007,6 +5008,17 @@ When the `raw` argument is `true`, it is required that no pending data from any 
 You can use the "raw request socket" returned by `ngx.req.socket(true)` to implement fancy protocols like [WebSocket](http://en.wikipedia.org/wiki/WebSocket), or just emit your own raw HTTP response header or body data. You can refer to the [lua-resty-websocket library](https://github.com/openresty/lua-resty-websocket) for a real world example.
 
 This function was first introduced in the `v0.5.0rc1` release.
+
+[Back to TOC](#nginx-api-for-lua)
+
+ngx.req.set\_cong
+-----------------
+
+**syntax:** ngx.req.set_cong(cong_name)*
+
+**context:** *access_by_lua&#42;, content_by_lua&#42;*
+
+Set TCP congestion control for each request connection. It only works on Linux platform and if target TCP congestion is not exists, it will generate a warning log.
 
 [Back to TOC](#nginx-api-for-lua)
 
--- a/src/ngx_http_lua_tcpcong.c
+++ b/src/ngx_http_lua_tcpcong.c
@@ -12,7 +12,7 @@
 #include "ngx_http_lua_api.h"
 
 
-static int ngx_socket_set_tcp_congestion(ngx_socket_t s,
+int ngx_socket_set_tcp_congestion(ngx_socket_t s,
     const char *cong_name, size_t cong_name_len);
 static int ngx_http_lua_ngx_set_tcp_congestion(lua_State *L);
 
@@ -20,7 +20,7 @@ static int ngx_http_lua_ngx_set_tcp_congestion(lua_State *L);
 #if (NGX_LINUX)
 
 
-static int
+int
 ngx_socket_set_tcp_congestion(ngx_socket_t s,
     const char *cong_name, size_t cong_name_len)
 {
@@ -32,7 +32,7 @@ ngx_socket_set_tcp_congestion(ngx_socket_t s,
 #else
 
 
-static int
+int
 ngx_socket_set_tcp_congestion(ngx_socket_t s,
     const char *cong_name, size_t cong_name_len)
 {
--- a/src/ngx_http_lua_tcpcong.c
+++ b/src/ngx_http_lua_tcpcong.c
@@ -4,9 +4,6 @@
 #include "ddebug.h"
 
 
-#define EOSNOTSUPPORT -23
-
-
 #include "ngx_http_lua_tcpcong.h"
 #include "ngx_http_lua_util.h"
 #include "ngx_http_lua_api.h"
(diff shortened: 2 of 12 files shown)


src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+
+
+#if (NGX_LINUX)
+
+int
+ngx_socket_set_tcp_congestion(ngx_socket_t s, const char* cong_name) {
+    u_char optval[16];
+    ngx_cpystrn(optval, (u_char *)cong_name, 16);
+    return setsockopt(s, IPPROTO_TCP, TCP_CONGESTION,
+                      (void *)optval, strlen((char *)optval));
```
agentzh: We should avoid using C strings in the first place. `strlen()` is costly.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#else
+
+int
+ngx_socket_set_tcp_congestion(ngx_socket_t s, const char* cong_name) {
```
agentzh: Style: `{` should be in the separate line according to the nginx coding style. BTW, you can always find such things via the `ngx-releng` tool here:

https://github.com/openresty/openresty-devel-utils/blob/master/ngx-releng

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+    nargs = lua_gettop(L);
+    if (nargs != 1) {
+        return luaL_error(L, "attempt to pass %d arguments, bug accepted 1",
+                          nargs);
+    }
+    type = lua_type(L, 1);
+    if (type != LUA_TSTRING) {
+        return luaL_error(L, "require string parameter");
+    }
+    p = lua_tolstring(L, 1, &len);
```
agentzh: Style: always need a blank line after `if` statement blocks. Please fix other similar places as well.

src/ngx_http_lua_tcpcong.h
```diff
@@ -0,0 +1,22 @@
+
+/*
+ * Copyright (C) Rain Li (blacktear23)
```
agentzh: You need to grant the copyright to the current copyright holders of this repo for your code to be merged.

src/ngx_http_lua_tcpcong.c
agentzh: Also, we should use a space *before* instead of after `*` in the pointer type signature. See other similar places for examples.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+    ret = ngx_socket_set_tcp_congestion(r->connection->fd, p);
```
agentzh: It's a shame that we throw away the readily available string length info and require this C function to do the string length calculation again using `strlen()`.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+                      "set tcp congestion %s failed: %d", p, ret);
+    }
+    return 1;
+}
+
+void
+ngx_http_lua_inject_req_tcp_congestion_api(lua_State *L)
+{
+    lua_pushcfunction(L, ngx_http_lua_ngx_set_tcp_congestion);
+    lua_setfield(L, -2, "set_cong");
```
agentzh: I suggest use the current `setoption()` method instead of inventing a new method, as in

```lua
sock:setoption("congestion", "reno")
```

README.markdown
```diff
@@ -5010,6 +5011,17 @@ This function was first introduced in the `v0.5.0rc1` release.
...
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.req.set\_cong
+-----------------
+
+**syntax:** ngx.req.set_cong(cong_name)*
+
+**context:** *access_by_lua&#42;, content_by_lua&#42;*
+
+Set TCP congestion control for each request connection. It only works on Linux platform and if target TCP congestion is not exists, it will generate a warning log.
```
agentzh: we should use the term "congestion control algorithm" instead of "congestion control". The latter is confusing.

Also "is not exists" is not correct English. We say "does not exit".

And we need the word `the` before `target`.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+int ngx_socket_set_tcp_congestion(ngx_socket_t s, const char* cong_name);
+static int ngx_http_lua_ngx_set_tcp_congestion(lua_State *L);
+
+
+#if (NGX_LINUX)
+
+int
+ngx_socket_set_tcp_congestion(ngx_socket_t s, const char* cong_name) {
+    u_char optval[16];
+    ngx_cpystrn(optval, (u_char *)cong_name, 16);
```
agentzh: We need a space after the type casting expression `(u_char *)`.

Also, i don't think we need to copy a new buffer here in the first place.

t/062-count.t
```diff
@@ -172,7 +172,7 @@ n = 24
 --- request
 GET /test
 --- response_body
-n = 24
+n = 25
```
agentzh: We definitely need more test cases to actually cover the new code paths you added in this patch.

README.markdown
agentzh: Also, we should list some typical (if not all) algorithm names allowed by modern Linux kernels.

README.markdown
```diff
@@ -5010,6 +5011,17 @@ This function was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.req.set\_cong
```
agentzh: I think we should make it a method of the cosocket object so that we can set socket-level congestion control algorithms for the upstream cosockets as well. This would be much more useful and general.

README.markdown
```diff
@@ -3138,6 +3138,7 @@ Nginx API for Lua
 * [ngx.req.append_body](#ngxreqappend_body)
 * [ngx.req.finish_body](#ngxreqfinish_body)
 * [ngx.req.socket](#ngxreqsocket)
+* [ngx.req.set_cong](#ngxreqset_cong)
```
agentzh: BTW, this file is automatically generated from `doc/HttpLuaModule.wiki`, to quote the comment at the beginning of this file:

```
cjson.safe: {"foo":123}
<!---
-->
```

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+#endif
+
+/**
+ * Set TCP Congestion
+ * */
+static int
+ngx_http_lua_ngx_set_tcp_congestion(lua_State *L)
+{
+    ngx_http_request_t  *r;
+    int                 nargs;
```
agentzh: Alignment issue. The identifiers (excluding `*`) should be aligned up vertically. See the rest of the code base for such examples. Please fix other similar places as well.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+    ngx_cpystrn(optval, (u_char *)cong_name, 16);
+    return setsockopt(s, IPPROTO_TCP, TCP_CONGESTION,
+                      (void *)optval, strlen((char *)optval));
+}
+
+#else
+
+int
+ngx_socket_set_tcp_congestion(ngx_socket_t s, const char* cong_name) {
+    return 0;
```
agentzh: We should throw out an error here instead of silently ignoring the lack of OS support.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+    }
+    p = lua_tolstring(L, 1, &len);
+    ret = ngx_socket_set_tcp_congestion(r->connection->fd, p);
+    if (ret < 0) {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
+                      "set tcp congestion %s failed: %d", p, ret);
+    }
+    return 1;
+}
+
```
agentzh: Style: need 2 blank lines to separate successive C function definitions.

src/ngx_http_lua_tcpcong.c
```diff
@@ -0,0 +1,81 @@
...
+    }
+    type = lua_type(L, 1);
+    if (type != LUA_TSTRING) {
+        return luaL_error(L, "require string parameter");
+    }
+    p = lua_tolstring(L, 1, &len);
+    ret = ngx_socket_set_tcp_congestion(r->connection->fd, p);
+    if (ret < 0) {
+        ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
+                      "set tcp congestion %s failed: %d", p, ret);
```
agentzh: We should return the error string to the Lua caller for the Lua caller to handle any errors. We should not log an error message directly here.

src/ngx_http_lua_tcpcong.c
blacktear23: I will try to implements this but will take more time.

README.markdown
blacktear23: @agentzh is there some document about cosocket:setoption and cosocket? Or I should read the code?

agentzh: @blacktear23 See https://github.com/openresty/lua-nginx-module#tcpsocksetoption It's in the official documentation.

blacktear23: @agentzh After I do some research on let tcpsock:setoption can set congestion control I found I cannot change request's congestion control by below:
```nginx
access_by_lua '
    local sock, err = ngx.req.socket()
    if sock then
        sock.setoption("congestion", "reno")
    else
        if err then
            ngx.log(ngx.ERR, "error: ", err)
        end
    end
'
```

And error.log will found something like `content_by_lua(nginx.conf:42):9: error: no body,`

agentzh: @blacktear23 I think the error message is self-explanatory. You need a proper request body to use normal req cosockets. Otherwise, you may want to use "raw" req cosockets instead. Please read the official documentation for more details.

blacktear23: @agentzh I do read the document and I found if I set raw to true, I will get error message as below: `content_by_lua(nginx.conf:42):6: attempt to call field 'setoption' (a nil value)`

blacktear23: I found the problem.

src/ngx_http_lua_tcpcong.c
blacktear23: After I test my setoption implementation I found a problem: If use `ngx.req.socket(true)`, user should handle HTTP response by himself. If use `ngx.req.socket(false)` GET request will get error: `no body`. So if user just want to change request connection's TCP congestion `ngx.req.set_cong` will be the easy way to do it.

agentzh: @blacktear23 The "no body" error message is self-explanatory. The current request must have a request body to use this `ngx.req.socket(false)` API. Please read the official document more carefully.

agentzh: @blacktear23 No, `ngx.req.set_cong()` is a horrible API name. We should use the `setoption()` API instead. We definitely won't want to add a new Lua method for every new option we add, which will lead to a very bloated API and more memory and GC overhead.

blacktear23: @agentzh what about add an API：ngx.req.setoption ？

agentzh: @blacktear23 Why? It should be an instance method instead of a static method. This is especially true for upstream cosockets created by `ngx.socket.tcp()`, for example.

blacktear23: @agentzh One scenario is to use Lua change request connection's TCP congestion to improve TCP performance for file download, and then fall through Nginx basic logic. Such as the config below:

```
server {
    listen 80;
    location / {
        root /var/www/html;
        access_by_lua '
            ngx.req.setoption("congestion", "bbr")
        ';
    }
}
```

As you mentioned implement setoption for cosocket, after I get the socket by `ngx.req.socket` I cannot fall through Nginx basic logic. 

Is there I have some miss understanding on OpenResty?

agentzh: @blacktear23 That's a fair point. Then I suggest we add a static method to a new `ngx.req` Lua module in the lua-resty-core library:

https://github.com/openresty/lua-resty-core/

Just like its existing `ngx.ssl` and `ngx.re` Lua modules :)

We can just add a pure C API in this lua-nginx-module repo and invoke the C API through LuaJIT FFI from within the `ngx.req` module in lua-resty-core.

PR #1188, opened 2017-11-15
fix the incorrect received response bytes using openresty 1.13.6.1

gnought:
The received response bytes is added by 4, as the server name value is changed from "nginx" to "openresty". Should the test be sticked to nginx official version or openresty ?

### DIFF
--- a/t/023-rewrite/socket-keepalive.t
+++ b/t/023-rewrite/socket-keepalive.t
@@ -241,7 +241,7 @@ GET /t
 --- response_body
 connected: 1
 request sent: 61
-received response of 156 bytes
+received response of 160 bytes
 done
 --- no_error_log
 [error]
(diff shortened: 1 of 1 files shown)


spacewander:
AFAIK, the length of response is four bytes more with OpenResty, is because OpenResty uses `Server: openresty/...` by default, but Nginx uses `Server: nginx/...`.

I think the test suite of `lua-nginx-module` should be not related with OpenResty, since people may use lua-nginx-module as a separate Nginx C module.

agentzh:
Yeah, right, the ngx_lua test suite should not directly be run with OpenResty. We should always use the `util/build.sh` script to build the nginx properly. See the Travis CI config file for details:

https://github.com/openresty/lua-nginx-module/blob/master/.travis.yml

PR #1191, opened 2017-11-17
ssl: support enabling TLSv1.3 via lua_ssl_protocols

ghedo:
This simply mirrors commit https://github.com/nginx/nginx/commit/9a37eb3a62130473596e0e4c2e388d80bdb14956#diff-0584d16332cf0d6dd9adb990a3c76a0c in NGINX.

### DIFF
--- a/t/155-tlsv1_3.t
+++ b/t/155-tlsv1_3.t
@@ -1,0 +1,109 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+
+use Test::Nginx::Socket::Lua;
+
+repeat_each(3);
+
+# All these tests need to have new openssl
+my $NginxBinary = $ENV{'TEST_NGINX_BINARY'} || 'nginx';
+my $openssl_version = eval { `$NginxBinary -V 2>&1` };
+
+if ($openssl_version =~ m/built with OpenSSL (0|1\.0|1\.1\.0.*)/) {
+    plan(skip_all => "too old OpenSSL, need 1.1.1, was $1");
+} else {
+    plan tests => repeat_each() * (blocks() * 5);
+}
+
+$ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
+$ENV{TEST_NGINX_MEMCACHED_PORT} ||= 11211;
+
+sub read_file {
+    my $infile = shift;
+    open my $in, $infile
+        or die "cannot open $infile for reading: $!";
+    my $cert = do { local $/; <$in> };
+    close $in;
+    $cert;
+}
+
+our $TestCertificate = read_file("t/cert/test.crt");
+our $TestCertificateKey = read_file("t/cert/test.key");
+
+#log_level 'warn';
+log_level 'debug';
+
+no_long_string();
+#no_diff();
+
+run_tests();
+
+__DATA__
+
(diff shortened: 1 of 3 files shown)


agentzh:
@ghedo Is it possible to add a test case to the test suite so that we can have this change properly covered?

agentzh:
@ghedo Thanks for your contribution as always!

ghedo:
The current TLSv1.3 situation is such that it would require special NGINX+OpenResty builds (e.g. with BoringSSL, or with OpenSSL 1.1 which is not supported) so I don't think it's currently possible to add tests for this.

agentzh:
@ghedo I see. We need to polish up and merge one of the OpenSSL 1.1 PRs first.

agentzh:
@ghedo OpenSSL 1.1.0 support has landed in the latest master branch. Just FYI :) Thanks!

ghedo:
I added a commit to test this feature, but apparently OpenSSL hasn't released TLSv1.3 in a stable version yet, so the build fails.

ghedo:
And even when OpenSSL will release it, the test will only work in OpenSSL 1.1.0, so the 1.0.2 builds will fail anyway.

agentzh:
@ghedo We can try checking the OpenSSL version number in the `nginx -V` output and skip the whole `.t` test file conditionally (assuming your TLS 1.3 tests are in a separate `.t` file)?

We are already doing something similar in the existing test suite:

https://github.com/openresty/lua-nginx-module/blob/master/t/139-ssl-cert-by.t#L8

Will you update your test file accordingly? Many thanks!

ghedo:
@agentzh ok, done. Since the required OpenSSL version hasn't been release yet, I added a note to the travis config, to add the `enable-tls1_3` option to the OpenSSL build once it gets added, so it won't be forgotten.

ghedo:
Also, for the record, I tested this locally with a git checkout of OpenSSL master and the test works.

t/155-tlsv1_3.t
```diff
@@ -0,0 +1,109 @@
...
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        server_name   test.com;
+        ssl_certificate ../html/test.crt;
+        ssl_certificate_key ../html/test.key;
+        ssl_protocols TLSv1.2 TLSv1.3;
+
+        server_tokens off;
+        location /foo {
+            default_type 'text/plain';
+            content_by_lua 'ngx.status = 201 ngx.say("foo") ngx.exit(201)';
```
agentzh: Newly added test cases should use the `*_by_lua_block` directives for inlined Lua code instead.

t/155-tlsv1_3.t
```diff
@@ -0,0 +1,109 @@
...
+    }
+--- config
+    server_tokens off;
+    lua_ssl_trusted_certificate ../html/test.crt;
+    lua_ssl_protocols TLSv1.2 TLSv1.3;
+    location /t {
+        #set $port 5000;
+        set $port $TEST_NGINX_MEMCACHED_PORT;
+
+        content_by_lua '
```
agentzh: Ditto.

agentzh:
Just merged with various minor edits and some docs. Thanks!

PR #1194, opened 2017-11-21
optimize: destroyed the Lua VM and avoided running any init_worker_by_lua* inside cache processes.

spacewander:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1487,6 +1487,8 @@ This hook is often used to create per-worker reoccurring timers (via the [ngx.ti
 
 This directive was first introduced in the `v0.9.5` release.
 
+After `v0.10.12` release, the specified Lua code doesn't run in cache processes anymore.
+
 [Back to TOC](#directives)
 
 init_worker_by_lua_block
--- a/src/ngx_http_lua_initworkerby.c
+++ b/src/ngx_http_lua_initworkerby.c
@@ -40,10 +40,33 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
     lmcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_lua_module);
 
-    if (lmcf == NULL
-        || lmcf->init_worker_handler == NULL
-        || lmcf->lua == NULL)
-    {
+    if (lmcf == NULL) {
+        return NGX_OK;
+    }
+
+    if (lmcf->lua != NULL) {
+        /* disable init_worker_by_lua* and destroy lua VM in cache processes */
+        if (ngx_process == NGX_PROCESS_HELPER) {
+#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
+            if (!ngx_is_privileged_agent) {
+#endif
+
+                ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                               "lua close the global Lua VM %p in "
+                               "cache process %P", lmcf->lua, ngx_pid);
+                lmcf->vm_cleanup->handler(lmcf->vm_cleanup->data);
+                lmcf->vm_cleanup->handler = NULL;
+                return NGX_OK;
+#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
+            }
+#endif
+        }
+
+    } else {
+        return NGX_OK;
+    }
+
+    if (lmcf->init_worker_handler == NULL) {
         return NGX_OK;
     }
 
--- a/src/ngx_http_lua_initworkerby.c
+++ b/src/ngx_http_lua_initworkerby.c
@@ -40,11 +40,7 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
     lmcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_lua_module);
 
-    if (lmcf == NULL) {
-        return NGX_OK;
-    }
-
-    if (lmcf->lua != NULL) {
+    if (lmcf != NULL && lmcf->lua != NULL) {
         /* disable init_worker_by_lua* and destroy lua VM in cache processes */
         if (ngx_process == NGX_PROCESS_HELPER) {
 #if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
@@ -61,12 +57,12 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
             }
 #endif
         }
-
-    } else {
-        return NGX_OK;
     }
 
-    if (lmcf->init_worker_handler == NULL) {
+    if (lmcf == NULL
+        || lmcf->init_worker_handler == NULL
+        || lmcf->lua == NULL)
+    {
         return NGX_OK;
     }
 
--- a/src/ngx_http_lua_initworkerby.c
+++ b/src/ngx_http_lua_initworkerby.c
@@ -40,29 +40,30 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
     lmcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_lua_module);
 
-    if (lmcf != NULL && lmcf->lua != NULL) {
-        /* disable init_worker_by_lua* and destroy lua VM in cache processes */
-        if (ngx_process == NGX_PROCESS_HELPER) {
-#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
-            if (!ngx_is_privileged_agent) {
-#endif
-
-                ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
-                               "lua close the global Lua VM %p in "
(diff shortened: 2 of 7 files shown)


src/ngx_http_lua_initworkerby.c
```diff
@@ -40,10 +40,33 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
     lmcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_lua_module);
 
-    if (lmcf == NULL
```
agentzh: I think we should still keep this original `if` statement to make the code below easier to understand.

src/ngx_http_lua_util.c
```diff
@@ -3768,6 +3768,10 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
 
     cln->data = state;
 
+    if (lmcf->vm_cleanup == NULL) {
+        lmcf->vm_cleanup = cln;
```
agentzh: Hmm, this may lead to use-after-free when the VM is per-request (when `lua_code_cache off` is configured). But maybe this is fine since the main VM is always the first to be created. Better add a code comment here to be clearer?

src/ngx_http_lua_util.c
```diff
@@ -3768,6 +3768,11 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
 
     cln->data = state;
 
+    /* this VM cleanup should be only called in the main vm */
```
agentzh: grammar: "should be only" => "should only be"

spacewander:
@agentzh 
The code is updated.

src/ngx_http_lua_initworkerby.c
```diff
@@ -40,6 +40,25 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
     lmcf = ngx_http_cycle_get_module_main_conf(cycle, ngx_http_lua_module);
 
+    if (lmcf != NULL && lmcf->lua != NULL) {
```
agentzh: How about the following instead?

```C
    if (lmcf == NULL || lmcf->lua == NULL) {
        return NGX_OK;
    }

    /* lmcf != NULL && lmcf->lua != NULL */

    /* disable init_worker_by_lua* and destroy lua VM in cache processes */
    if (ngx_process == NGX_PROCESS_HELPER) {
#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
        if (!ngx_is_privileged_agent) {
#endif

            ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
                           "lua close the global Lua VM %p in "
                           "cache process %P", lmcf->lua, ngx_pid);
            lmcf->vm_cleanup->handler(lmcf->vm_cleanup->data);
            lmcf->vm_cleanup->handler = NULL;
            return NGX_OK;
#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
        }
#endif
    }

    if (lmcf->init_worker_handler == NULL) {
        return NGX_OK;
    }
```

This way we can save 2 duplicate checks.

t/124-init-worker.t
```diff
@@ -756,3 +770,24 @@ GET /t
 ok
 --- no_error_log
 [error]
+
+
+
+=== TEST 20: destory Lua VM in cache processes
+--- skip_eval
```
agentzh: We should not skip this in valgrind test mode since this test case passes successfully in the HUP reload + valgrind combined test mode. It's fine for this test case to fail in the sole valgrind mode since our EC2 test cluster does support whitelist of expected failures in certain test modes or architectures:

https://github.com/openresty/opsboy/blob/master/misc/logs/parse-logs#L19

src/ngx_http_lua_initworkerby.c
spacewander: @agentzh
Good point!

t/124-init-worker.t
```diff
@@ -16,8 +16,22 @@ $ENV{TEST_NGINX_RESOLVER} ||= '8.8.8.8';
 
 our $ServerRoot = server_root();
 
+add_block_preprocessor(sub {
```
agentzh: This applies to all the test cases in this test file. Maybe a bit overkill?

spacewander:
@agentzh
Also updated the test cases.

t/124-init-worker.t
```diff
@@ -18,6 +18,7 @@ our $ServerRoot = server_root();
 
 #no_diff();
 no_long_string();
+master_on();
```
agentzh: I think we should just uncomment the `master_on()` line on line 6 :)

src/ngx_http_lua_initworkerby.c
```diff
@@ -40,10 +40,30 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
...
+    if (lmcf == NULL || lmcf->lua == NULL) {
+        return NGX_OK;
+    }
+
+    /* lmcf != NULL && lmcf->lua != NULL */
+
+    /* disable init_worker_by_lua* and destroy lua VM in cache processes */
+    if (ngx_process == NGX_PROCESS_HELPER) {
+#if defined(HAVE_PRIVILEGED_PROCESS_PATCH) && !NGX_WIN32
+        if (!ngx_is_privileged_agent) {
```
agentzh: Seems like we can just use `&&` instead of `if` here. This way we can also save the `#if/#endif` pair below.

t/124-init-worker.t
```diff
@@ -756,3 +757,42 @@ GET /t
...
+--- config
+    location = /t {
+        return 200;
+    }
+--- request
+GET /t
+--- error_log eval
+[
+qr/lua close the global Lua VM .+ in cache process/,
+qr/lua close the global Lua VM .+ in cache process/,
```
agentzh: I think we should just use a single regex here instead of an array since it's identical to all the repeated requests.

README.markdown
```diff
@@ -1487,6 +1487,8 @@ This hook is often used to create per-worker reoccurring timers (via the [ngx.ti
 
 This directive was first introduced in the `v0.9.5` release.
 
+After `v0.10.12` release, the specified Lua code doesn't run in cache processes anymore.
```
agentzh: Maybe "cache helper processes" is better than "cache processes"?

t/124-init-worker.t
agentzh: Also, the use of `grep_error_log` would be better here since we can actually count the number of VM destructions (there should be two here, one for the cache loader and one for the cache manager).

agentzh: Additionally, we may want to add more tests for the cases when `init_worker_by_lua*` is used and when the privileged agent process is enabled?

agentzh:
@spacewander I made some edits in this PR and committed to this branch:

https://github.com/openresty/lua-nginx-module/tree/pcre-init-mem-issues

When running the `t/124-init-worker.t` test file with valgrind and hup reload mode, I'm seeing a lot of valgrind memory issues like below:

```
https://gist.github.com/agentzh/22ee4bad191b6eacdf024fba29b189d4
```

It's not likely to be caused by this PR but still a problem that needs to be triaged.

spacewander:
@agentzh 
I also found another two variants of this Valgrind warning:
```
==27672== Invalid read of size 8
==27672==    at 0x142A66: ngx_log_error_core (ngx_log.c:128)
==27672==    by 0x143AE4: ngx_pfree (ngx_palloc.c:191)
==27672==    by 0x59B762D: ??? (in /lib/x86_64-linux-gnu/libpcre.so.3.13.3)
==27672==    by 0x59B948B: pcre_free_study (in /lib/x86_64-linux-gnu/libpcre.so.3.13.3)
==27672==    by 0x20D2CD: ngx_http_lua_ffi_destroy_regex (ngx_http_lua_regex.c:2413)
```
and
```
==27672== Invalid read of size 8
==27672==    at 0x142C1B: ngx_log_error_core (ngx_log.c:201)
==27672==    by 0x143AE4: ngx_pfree (ngx_palloc.c:191)
==27672==    by 0x59B762D: ??? (in /lib/x86_64-linux-gnu/libpcre.so.3.13.3)
==27672==    by 0x59B948B: pcre_free_study (in /lib/x86_64-linux-gnu/libpcre.so.3.13.3)
==27672==    by 0x20D2CD: ngx_http_lua_ffi_destroy_regex (ngx_http_lua_regex.c:2413)
```

However, after reading the related code, I have not found evidence to show there is memory leak in regex methods. And,  none of the variable in `ngx_log_error_core` is allocated in heap.

Particularly, after comment out two debug log below, Valgrind doesn't complain any more in my side:
https://github.com/nginx/nginx/blob/master/src/core/ngx_palloc.c#L55
https://github.com/nginx/nginx/blob/master/src/core/ngx_palloc.c#L284

I guess it is a false positive report. How do you think about this issue?

agentzh:
@spacewander Memory invalid reads on freed memory are usually real problems unless they are page-aligned memory read optimizations. This looks like a real issue to me. I think it might just be `ngx_log_t` allocated in the wrong pool. Maybe we should allocate it ourselves?

agentzh:
Already committed to master with the proper fixes.

PR #1198, opened 2017-11-27
feature : $installation_prefix for NGX_PREFIX

chronolaw:
I add a new notation '$installation_prefix' which can be used in lua_package_cpath or lua_package_path.

With this  notation, we do not need to write :
lua_package_path    "/usr/local/openresty/lualib/?.lua;"；

and we can write like below:
 lua_package_path    "$installation_prefix/../lualib/?.lua;"

It may be more flexiable than absolute path.

### DIFF
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -147,7 +147,16 @@ ngx_http_lua_set_path(ngx_cycle_t *cycle, lua_State *L, int tab_idx,
     prefix = lua_tostring(L, -1);
     tmp_path = luaL_gsub(L, tmp_path, "$prefix", prefix);
     tmp_path = luaL_gsub(L, tmp_path, "${prefix}", prefix);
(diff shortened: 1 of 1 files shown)


spacewander:
@chronolaw 
AFAIK, `/usr/local/openresty/lualib/?.lua;` is in the default `package.path` search paths. And you could use `;;` to represent it without using the absolute path name. Like this, `lua_package_path "$prefix/your/?.lua;;"` will contain `/usr/local/openresty/lualib/?.lua`.

chronolaw:
It could be useful for the users who makes their own nginx+ngx_lua bundles, I think.

agentzh:
@chronolaw No, for people doing their own bundle, they should use the `-DLUA_DEFAULT_PATH='directory-list'` compilation option instead. This is much more flexible and general than using the NGINX installation prefix. OpenResty itself uses this technique.

Issue #1172, opened 2017-10-21
Compilation fails because OpenSSL 1.1 is not supported


alex-mashin:
While trying to build openresty-1.11.2.5 under Ubuntu 17.10, a compilation error happens:
```
src/event/ngx_event_openssl.c: In function ‘ngx_ssl_connection_error’:
src/event/ngx_event_openssl.c:2048:21: error: ‘SSL_R_NO_CIPHERS_PASSED’ undeclared (first use in this function); did you mean ‘SSL_R_NO_CIPHERS_SPECIFIED’?
             || n == SSL_R_NO_CIPHERS_PASSED                          /*  182 */
                     ^~~~~~~~~~~~~~~~~~~~~~~
                     SSL_R_NO_CIPHERS_SPECIFIED
```

Did you mean `SSL_R_NO_CIPHERS_SPECIFIED`?

spacewander:
Maybe the problem is the same with #903. What openssl version do you use under Ubuntu 17.10?

alex-mashin:
OpenSSL 1.1.0f  25 May 2017

spacewander:
@alex-mashin 
OpenSSL 1.1 is not supported yet. You could use OpenSSL 1.0.2 instead, or install OpenResty via official package: http://openresty.org/en/linux-packages.html#ubuntu

spacewander:
@alex-mashin 
Seems you are not lucky enough to use the official package...
You could download OpenSSL 1.0.2 and use `./configure --with-openssl=/path/to/lib/the_openssl_version(for example, openssl-1.0.2h)` to compile OpenResty, instead of changing system-wide OpenSSL.

spacewander:
Sorry for missing two things:
1. You need to apply patch before compiling the OpenSSL.
2. You need to compile the OpenSSL to get the library file before running `./configure --with-openssl=...`.

See the [.travis.yml](https://github.com/openresty/lua-nginx-module/blob/master/.travis.yml#L113).

agentzh:
@alex-mashin Seems like there is something already listening on the 80 port? Please stop those daemons and retry installing the openresty binary package. This is because installing openresty will start the daemon automatically.

agentzh:
@alex-mashin Or better, tune your system to avoid starting the service immediately after installing a deb package in general: https://major.io/2014/06/26/install-debian-packages-without-starting-daemons/

agentzh:
[...]

PR #922, opened 2016-11-28
OpenSSL 1.1.0 support - work in progress

doujiang24:
Base on branch openss-1.1 (https://github.com/openresty/lua-nginx-module/pull/761)

changes in this patch: avoid using RC4 & used the new error string.
But some test cases still failed,
1. seems `lua_ssl_verify_depth` doesn't work
2. can't yield in session fetch call back

I'll continue to step in. @ghedo @chipitsine @ctrochalakis will you also take a look at this? Thanks very much :)

I hereby granted the copyright of the changes in this pull request
to the authors of this lua-nginx-module project.

Issue #903, opened 2016-11-01
Build Fail with OpenSSL 1.1.0b


smoghaddas:
dear all
i compile openresty 1.11.2.1 with 1.1.0b and got this error ...

-I /proxy/source/openssl-1.1.0b/.openssl/include -I objs \
        -o objs/src/event/ngx_event_openssl.o \
        src/event/ngx_event_openssl.c
src/event/ngx_event_openssl.c: In function 'ngx_ssl_connection_error':
src/event/ngx_event_openssl.c:2048:21: error: 'SSL_R_NO_CIPHERS_PASSED' undeclared (first use in this function)
             || n == SSL_R_NO_CIPHERS_PASSED                          /*  182 */
                     ^
src/event/ngx_event_openssl.c:2048:21: note: each undeclared identifier is reported only once for each function it appears in
objs/Makefile:1287: recipe for target 'objs/src/event/ngx_event_openssl.o' failed
make[2]: *** [objs/src/event/ngx_event_openssl.o] Error 1
make[2]: Leaving directory '/proxy/source/openresty-1.11.2.1/build/nginx-1.11.2'
Makefile:8: recipe for target 'build' failed
make[1]: *** [build] Error 2
make[1]: Leaving directory '/proxy/source/openresty-1.11.2.1/build/nginx-1.11.2'
Makefile:4: recipe for target 'all' failed
make: *** [all] Error 2

bjoe2k4:
See #757 and #761.

agentzh:
Use the OpenSSL 1.0.2 series instead.

KeithTt:
So terrible... I just upgrade openssl to 1.1.0f few days ago.

thibaultcha:
FYI, support for OpenSSL 1.1.0 has recently been merged and is pending for release: https://github.com/openresty/lua-nginx-module/pull/1213

agentzh:
Already in the master branches.

PR #1213, opened 2017-12-22
feature: supported OpenSSL 1.1.0

spacewander:

For #903 and #1172.

Changes contains:
1. Take changes from #761 and #922.
2. Update the error message.
3. Compared to 1.0.2, OpenSSL 1.1.0 SSL_CTX_set_verify_depth verifies
one more level. Since Nginx doesn't change the default value for
ssl_verfiy_depth, I just update the test cases instead of updating the
C code.
4. Add OpenSSL 1.1.0c to the travis build.

### DIFF
--- a/src/ngx_http_lua_ssl_session_storeby.c
+++ b/src/ngx_http_lua_ssl_session_storeby.c
@@ -172,6 +172,8 @@ int
 ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
     ngx_ssl_session_t *sess)
 {
+    const u_char                    *sess_id;
+    unsigned int                     sess_id_len;
     lua_State                       *L;
     ngx_int_t                        rc;
     ngx_connection_t                *c, *fc = NULL;
@@ -247,11 +249,13 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
         }
     }
 
+    sess_id = SSL_SESSION_get_id(sess, &sess_id_len);
+
     cctx->connection = c;
     cctx->request = r;
     cctx->session = sess;
-    cctx->session_id.data = sess->session_id;
-    cctx->session_id.len = sess->session_id_length;
+    cctx->session_id.data = (u_char *) sess_id;
+    cctx->session_id.len = sess_id_len;
     cctx->done = 0;
 
     dd("setting cctx");
--- a/t/129-ssl-socket.t
+++ b/t/129-ssl-socket.t
@@ -108,10 +108,10 @@ sent http request: 59 bytes.
 received: HTTP/1.1 (?:200 OK|302 Found)
 close: 1 nil
 \z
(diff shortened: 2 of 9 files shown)


spacewander:
I will send a pr to stream-lua-nginx-module if this one is able to merge.

t/129-ssl-socket.t
```diff
@@ -693,7 +693,7 @@ SSL reused session
     server_tokens off;
     resolver $TEST_NGINX_RESOLVER ipv6=off;
     lua_ssl_trusted_certificate ../html/trusted.crt;
-    lua_ssl_verify_depth 1;
+    lua_ssl_verify_depth 0;
```
agentzh: Hmm, why is this necessary for OpenSSL 1.1?

t/129-ssl-socket.t
```diff
@@ -771,7 +772,7 @@ SSL reused session
...
```
agentzh: Ditto.

t/129-ssl-socket.t
spacewander: @agentzh 
`SSL_CTX_set_verify_depth` in OpenSSL 1.1.0 verifies
one more level. Since Nginx doesn't change the default value for
ssl_verfiy_depth, I just update the test cases instead of updating the
C code.

agentzh: @spacewander I see. Thanks for your explanation!

agentzh:
Merged. Thanks! I have to (temporarily) comment out the openssl 1.1.0c test case in the travis ci config file since `ngx_stream_lua` does not work with this openssl version series yet.

neilstuartcraig:
This is awesome news! I'll be super excited to see this work in a release :-)
Thanks so much @spacewander, you are an absolute legend for doing this - i am sure it wasn't simple or quick! Thanks also to everyone else who helped with it.

src/ngx_http_lua_ssl_session_storeby.c
```diff
@@ -247,11 +249,13 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
         }
     }
 
+    sess_id = SSL_SESSION_get_id(sess, &sess_id_len);
```
agentzh: This change breaks compatibility with older OpenSSL like 0.9.x.

agentzh: Just fixed in commit 7c8c503001.

S4mw1s3:
A big thank you to all of the people who helped implementing support for OpenSSL 1.1.0 !

thibaultcha:
Great job @ghedo @doujiang24 @spacewander, thank you! 👍

frankgould:
I am trying to upgrade my RPi mate 16.04 with openssl 1.1.0f but after doing so, I cannot log in because I get 'failed to start session' error. Is there something I did wrong in the install and a different way to prevent this?

PR #1216, opened 2017-12-25
bugfix: ssl by lua handlers might be NULL

haorenfsa:
ssl by lua handlers might be NULL. This happens when 2 or more nginx server block listen on a same port with different server_name; one is configured with 'xxx_by_lua' and another is not.

this is because the 2 server block in fact use the same ssl ctx, and openssl will call back once any of the server block set 'xxx_by_lua'. For those without such config, the code may end with calling a null function ptr, segment fault for ip=0x0.

-----
add some error screenshots
note that the nginx is not in the newest version, but same problem still exists

### DIFF
--- a/src/ngx_http_lua_ssl_session_fetchby.c
+++ b/src/ngx_http_lua_ssl_session_fetchby.c
@@ -307,8 +307,12 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn, u_char *id,
     L = ngx_http_lua_get_lua_vm(r, NULL);
 
     c->log->action = "fetching SSL session by lua";
-
-    rc = lscf->srv.ssl_sess_fetch_handler(r, lscf, L);
+    
+    if(lscf->srv.ssl_sess_fetch_handler) {
+        rc = lscf->srv.ssl_sess_fetch_handler(r, lscf, L);
+    } else {
+        rc = 0;
+    }
 
     if (rc >= NGX_OK || rc == NGX_ERROR) {
         cctx->done = 1;
--- a/src/ngx_http_lua_ssl_session_storeby.c
+++ b/src/ngx_http_lua_ssl_session_storeby.c
@@ -269,8 +269,11 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
     L = ngx_http_lua_get_lua_vm(r, NULL);
 
     c->log->action = "storing SSL session by lua";
-
-    rc = lscf->srv.ssl_sess_store_handler(r, lscf, L);
+    if(lscf->srv.ssl_sess_store_handler) {
+        rc = lscf->srv.ssl_sess_store_handler(r, lscf, L);
+    } else {
+        rc = 0;
+    }
 
     if (rc >= NGX_OK || rc == NGX_ERROR) {
         cctx->done = 1;


haorenfsa:
the test failed for clang compiler seems due to connection timeout?

spacewander:
@haorenfsa 
Good catch. Could you add a test case in `t/143-ssl-session-fetch.t` to verify your fix?

src/ngx_http_lua_ssl_session_fetchby.c
```diff
@@ -307,8 +307,12 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn, u_char *id,
     L = ngx_http_lua_get_lua_vm(r, NULL);
 
     c->log->action = "fetching SSL session by lua";
-
-    rc = lscf->srv.ssl_sess_fetch_handler(r, lscf, L);
+    
+    if(lscf->srv.ssl_sess_fetch_handler) {
```
agentzh: Style: need a space after `if`. Also, we should test the pointer value against `NULL` instead of testing it as a boolean as per nginx's coding convention.

src/ngx_http_lua_ssl_session_fetchby.c
```diff
@@ -307,8 +307,12 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn, u_char *id,
...
+        rc = lscf->srv.ssl_sess_fetch_handler(r, lscf, L);
+    } else {
```
agentzh: Style: needs a blank line before this `} else {` line.

src/ngx_http_lua_ssl_session_storeby.c
```diff
@@ -269,8 +269,11 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
     L = ngx_http_lua_get_lua_vm(r, NULL);
 
     c->log->action = "storing SSL session by lua";
-
-    rc = lscf->srv.ssl_sess_store_handler(r, lscf, L);
+    if(lscf->srv.ssl_sess_store_handler) {
```
agentzh: Ditto.

src/ngx_http_lua_ssl_session_storeby.c
```diff
@@ -269,8 +269,11 @@ ngx_http_lua_ssl_sess_store_handler(ngx_ssl_conn_t *ssl_conn,
...
+        rc = lscf->srv.ssl_sess_store_handler(r, lscf, L);
+    } else {
```
agentzh: Ditto.

agentzh:
I just had a closer look, it seems that you are using an older version of ngx_lua? The `ssl_session_fetch_by_lua*` directives are no longer allowed in the `server {}` blocks. See

https://github.com/openresty/lua-nginx-module/#ssl_session_fetch_by_lua_block

and also

https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_module.c#L546

haorenfsa:
@agentzh Yes, this happens when I'm using lua module version 0.10.6. No longer an issue after 0.10.7 
when commit 37e5362088bd659e318aae568b268719bd0d6707 change these to `http {}` context.

By the way, I also notice that `ssl_certificate_by_lua` used to have the same problem, and now it's fixed with null pointer check like what I did above ( by commit 8827a7f8332b9556c04fe91153ba55d677489907 according to issue #1055 ). 

A little suggest that maybe we shoud keep these in a same style? Or maybe current state is better, then I shall close this request.

agentzh:
@haorenfsa They are different. `ssl_certificate_by_lua*` cannot be allowed in the `http {}` scope only since certificates have to be bound to `server_name` per se. I'm closing this.

PR #1218, opened 2017-12-27
feature: ngx.worker.parent_pid

chronolaw:
Nginx 1.13.8 adds a new feature : ngx_parent, which is the pid of master process.

I think it is an useful feature, with this we can identify multi nginx instance, so I add a new API : ngx.worker.parent_pid to ngx_lua.

I also writed a lua-resty-masterpid opm package, which is the same function but works in Lua. Which you think is better?

### DIFF
--- a/src/ngx_http_lua_worker.c
+++ b/src/ngx_http_lua_worker.c
@@ -18,6 +18,7 @@
 
 static int ngx_http_lua_ngx_worker_exiting(lua_State *L);
 static int ngx_http_lua_ngx_worker_pid(lua_State *L);
+static int ngx_http_lua_ngx_worker_parent_pid(lua_State *L);
 static int ngx_http_lua_ngx_worker_id(lua_State *L);
 static int ngx_http_lua_ngx_worker_count(lua_State *L);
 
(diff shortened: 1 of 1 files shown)


agentzh:
@chronolaw I think `master_pid` is a better method name here. Also, will you please add this new API function to the `ngx.worker` Lua module in lua-resty-core instead:

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/process.md

We do not want the old Lua API in the giant `ngx` global table to grow forever (which incurs more and more GC overhead).

agentzh:
@chronolaw Also, we need a test case for this new feature in the existing test suite.

chronolaw:
OK, I will see the documents then rewrite the code.

agentzh:
@chronolaw Thanks!

chronolaw:
I have a question : should I still keep the ffi API in ngx_lua(and remove ngx.worker.parent_pid) then call it in ngx.process , or I write pure Lua code in ngx.process?

agentzh:
@chronolaw No more API functions should be added to the giant `ngx` global table.

agentzh:
@chronolaw That's the whole point of doing new APIs in separate Lua modules.

agentzh:
@chronolaw Otherwise we're just making the situation even worse ;)

chronolaw:
Yes, I understand your meaning，so I will write pure lua code, and abandon this pull request, OK?

And would is this `master_pid` in ngx.process or ngx.woker ? I think it should be in ngx.process.

chronolaw:
I will write in Chinese, sorry, but you can answer me in Enginsh:
我明白你的意思，不要用C的栈调用方式向ngx_lua里加接口，但是否要在ngx_lua里写ffi函数让resty.core去调用，也就是说要实现一个ngx_http_lua_ffi_master_id的C函数。
对目前OpenResty的开发思路还不是太确定，是应当尽量写纯Lua代码，还是写ffi C函数，再在Lua里调用。
另外，这个`master_pid`接口是应该放在`ngx.process`里吧，但之前的回复却是说的ngx.worker。

agentzh:
It should be in the ngx.process Lua module and you should mimic the current implementation of those existing Lua API functions in that Lua module.

chronolaw:
OK， I will reimplement a C function ngx_http_lua_ffi_master_pid，then call it in ngx.process.

Please close this PR, I will open a new one.

PR #1219, opened 2017-12-27
feature: ngx_http_lua_ffi_master_pid

chronolaw:
Nginx 1.13.8 adds a new feature : ngx_parent, which is the pid of master process.

I think it is an useful feature, with this we can identify multi nginx instance, so I add a ffi C function : ngx_http_lua_ffi_master_pid, which could works in `lua-resty-core`.

### DIFF
--- a/src/ngx_http_lua_worker.c
+++ b/src/ngx_http_lua_worker.c
@@ -105,7 +105,7 @@ ngx_http_lua_ffi_master_pid(void)
 #if (nginx_version >= 1013008)
     return (int) ngx_parent;
 #else
-    return (int) ((ngx_process == NGX_PROCESS_SINGLE) ? ngx_pid : getppid());
+    return -1;
 #endif
 }
 
--- a/src/ngx_http_lua_worker.c
+++ b/src/ngx_http_lua_worker.c
@@ -100,6 +100,17 @@ ngx_http_lua_ffi_worker_pid(void)
 
 
 int
+ngx_http_lua_ffi_master_pid(void)
+{
+#if (nginx_version >= 1013008)
+    return (int) ngx_parent;
+#else
+    return (int) ((ngx_process == NGX_PROCESS_SINGLE) ? ngx_pid : getppid());
+#endif
+}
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_worker.c
```diff
@@ -99,6 +99,17 @@ ngx_http_lua_ffi_worker_pid(void)
 }
 
 
+int
+ngx_http_lua_ffi_master_pid(void)
+{
+#if (nginx_version >= 1013008)
+    return (int) ngx_parent;
+#else
+    return (int) ((ngx_process == NGX_PROCESS_SINGLE) ? ngx_pid : getppid());
```
agentzh: I think we should return an error here instead of returning a potential wrong value here.

chronolaw:
I have updated the code as the review say ,pleas check it ,thanks.

src/ngx_http_lua_worker.c
```diff
@@ -99,6 +99,17 @@ ngx_http_lua_ffi_worker_pid(void)
 }
 
 
+int
+ngx_http_lua_ffi_master_pid(void)
+{
+#if (nginx_version >= 1013008)
+    return (int) ngx_parent;
+#else
+    return -1;
```
agentzh: Using `NGX_ERROR` here is better?

src/ngx_http_lua_worker.c
```diff
@@ -99,6 +99,17 @@ ngx_http_lua_ffi_worker_pid(void)
 }
 
 
+int
+ngx_http_lua_ffi_master_pid(void)
+{
+#if (nginx_version >= 1013008)
+    return (int) ngx_parent;
```
agentzh: Wen `master_process off;` is configured inside `nginx.conf`, then this will yield a wrong value (a different value than the one in the `nginx.pid` file). In this case, we should return `ngx_pid` instead.

agentzh:
Merged by fixing the aforementioned issues myself. Thanks!

PR #1226, opened 2018-01-08
feature: added the "exptime" argument to the pure C function of the shdict:incr() API.

thibaultcha:

Sister PR for openresty/lua-resty-core#165

### DIFF
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -2623,11 +2623,12 @@ ngx_http_lua_ffi_shdict_get(ngx_shm_zone_t *zone, u_char *key,
 int
 ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
     size_t key_len, double *value, char **err, int has_init, double init,
-    int *forcible)
+    long init_ttl, int *forcible)
 {
     int                          i, n;
     uint32_t                     hash;
     ngx_int_t                    rc;
+    ngx_time_t                  *tp = NULL;
     ngx_http_lua_shdict_ctx_t   *ctx;
     ngx_http_lua_shdict_node_t  *sd;
     double                       num;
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_shdict.c
```diff
@@ -2623,11 +2623,12 @@ ngx_http_lua_ffi_shdict_get(ngx_shm_zone_t *zone, u_char *key,
 int
 ngx_http_lua_ffi_shdict_incr(ngx_shm_zone_t *zone, u_char *key,
     size_t key_len, double *value, char **err, int has_init, double init,
-    int *forcible)
+    int exptime, int *forcible)
```
agentzh: Should be `init_ttl` to avoid any ambiguities here. Also, we should use `long` type here. `int` is 32-bit usually and can easily get overflown.

thibaultcha: Sounds like a saner name to me :+1: 

Should we do the same modification to this value's type to the already existing implementations of it? Like `ngx_http_lua_ffi_shdict_set_expire`, `ngx_http_lua_ffi_shdict_store`, etc... In another PR of course.

agentzh: @thibaultcha No, they do not have such ambiguities as the `incr()` method.

thibaultcha: What I mean is that they do use the `int` type as well (I agree their naming is fine).

thibaultcha: Initial comment addressed.

agentzh: @thibaultcha Oh, right. Better in a separate PR?

agentzh: @thibaultcha Thanks for the catch!

agentzh:
@thibaultcha I've already merged this. Will you create another PR to document this feature properly? Thanks!

thibaultcha:
Will do

Issue #1232, opened 2018-01-11
log bug: log a wrong phase


yandongxiao:
this is my test code
```
daemon off;
error_log logs/error.log info;
worker_processes 1;

events{}

http {
    server {
        location / {
            rewrite_by_lua_block {
                ngx.ctx.name = "hello"
            }

            content_by_lua_block {
                ngx.log(ngx.INFO, ngx.ctx.name)
            }

            body_filter_by_lua_block {
                ngx.log(ngx.INFO, ngx.ctx.name)
            }
        }
    }
}
```

and this is my log
```
2018/01/11 17:41:36 [notice] 69083#2161562: start worker process 69084
2018/01/11 17:41:37 [info] 69084#2161563: *1 [lua] content_by_lua(test.conf:16):2: hello, client: 127.0.0.1, server: , request: "GET / HTTP/1.1", host: "127.0.0.1:8000"
2018/01/11 17:41:37 [info] 69084#2161563: *1 [lua] content_by_lua(test.conf:16):2: hello, client: 127.0.0.1, server: , request: "GET / HTTP/1.1", host: "127.0.0.1:8000"
2018/01/11 17:41:37 [info] 69084#2161563: *1 kevent() reported that client 127.0.0.1 closed keepalive connection
```
>
- the phase is logged wrong
- the log problem is triggered  when lua code is the same but at different phase
- both mac and linux system have this problem

spacewander:
@yandongxiao
Just provide a quick fix for the bug you found.

qlees:
Logging it in contet_by_lua, isn't it?

agentzh:
It's just that the Lua code chunk name is wrong. Not sure if we can swap the loaded Lua code chunks' names on-the-fly. The easiest way is to stop sharing identical inlined Lua code chunks across different `*_by_lua` and `*_by_lua_block` handlers. For nginx.conf having many identical Lua code snippets (especially when using nginx's `include` directive), @spacewander's fix will lead to more Lua GC objects at runtime and also more GC CPU overhead. Well, maybe it's not a big deal.

agentzh:
Adding support for swapping Lua code check names for already loaded Lua code chunks would require patching the LuaJIT core and expose such an API, which is nontrivial and may not worth the troubles.

spacewander:
@agentzh 
The GC overhead is acceptable, since it could not be too complex `include` rules with many identical Lua code snippets for nginx configuration files written by human.

[...]

PR #1234, opened 2018-01-11
bugfix: avoided messing up Lua inline code cache across different phases

spacewander:
For #1232 

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -173,13 +173,15 @@ ngx_http_lua_balancer_by_lua(ngx_conf_t *cf, ngx_command_t *cmd,
 
         lscf->balancer.src = value[1];
 
-        p = ngx_palloc(cf->pool, NGX_HTTP_LUA_INLINE_KEY_LEN + 1);
+        p = ngx_palloc(cf->pool,
+                       sizeof("balancer_by_lua") + NGX_HTTP_LUA_INLINE_KEY_LEN);
         if (p == NULL) {
             return NGX_CONF_ERROR;
         }
 
         lscf->balancer.src_key = p;
 
+        p = ngx_copy(p, "balancer_by_lua", sizeof("balancer_by_lua") - 1);
         p = ngx_copy(p, NGX_HTTP_LUA_INLINE_TAG, NGX_HTTP_LUA_INLINE_TAG_LEN);
         p = ngx_http_lua_digest_hex(p, value[1].data, value[1].len);
         *p = '\0';
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -271,13 +271,15 @@ ngx_http_lua_set_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 
     filter_data->size = filter.size;
 
-    p = ngx_palloc(cf->pool, NGX_HTTP_LUA_INLINE_KEY_LEN + 1);
+    p = ngx_palloc(cf->pool,
+                   sizeof("set_by_lua") + NGX_HTTP_LUA_INLINE_KEY_LEN);
     if (p == NULL) {
         return NGX_CONF_ERROR;
     }
 
     filter_data->key = p;
 
+    p = ngx_copy(p, "set_by_lua", sizeof("set_by_lua") - 1);
     p = ngx_copy(p, NGX_HTTP_LUA_INLINE_TAG, NGX_HTTP_LUA_INLINE_TAG_LEN);
     p = ngx_http_lua_digest_hex(p, value[2].data, value[2].len);
     *p = '\0';
@@ -448,6 +450,7 @@ ngx_http_lua_rewrite_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
 char *
 ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 {
+    size_t                       chunkname_len;
     u_char                      *p, *chunkname;
     ngx_str_t                   *value;
     ngx_http_lua_main_conf_t    *lmcf;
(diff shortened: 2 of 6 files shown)


src/ngx_http_lua_directive.c
```diff
@@ -487,19 +490,22 @@ ngx_http_lua_rewrite_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             return NGX_CONF_ERROR;
         }
 
+        chunkname_len = strlen((char *) chunkname);
```
agentzh: The use of `strlen` is unfortunate (even if we have to, we should use `ngx_strlen()` instead). Can we make `ngx_http_lua_gen_chunk_name` simply return the string length directly? The string length was readily there and was simply discarded.

src/ngx_http_lua_balancer.c
```diff
@@ -173,13 +173,15 @@ ngx_http_lua_balancer_by_lua(ngx_conf_t *cf, ngx_command_t *cmd,
...
-        p = ngx_palloc(cf->pool, NGX_HTTP_LUA_INLINE_KEY_LEN + 1);
+        p = ngx_palloc(cf->pool,
+                       sizeof("balancer_by_lua") + NGX_HTTP_LUA_INLINE_KEY_LEN);
         if (p == NULL) {
             return NGX_CONF_ERROR;
         }
 
         lscf->balancer.src_key = p;
 
+        p = ngx_copy(p, "balancer_by_lua", sizeof("balancer_by_lua") - 1);
```
agentzh: This does not look right. Seems like we should prepend the chunk name here as well?

src/ngx_http_lua_directive.c
```diff
@@ -271,13 +271,15 @@ ngx_http_lua_set_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
-    p = ngx_palloc(cf->pool, NGX_HTTP_LUA_INLINE_KEY_LEN + 1);
+    p = ngx_palloc(cf->pool,
+                   sizeof("set_by_lua") + NGX_HTTP_LUA_INLINE_KEY_LEN);
     if (p == NULL) {
         return NGX_CONF_ERROR;
     }
 
     filter_data->key = p;
 
+    p = ngx_copy(p, "set_by_lua", sizeof("set_by_lua") - 1);
```
agentzh: Ditto.

src/ngx_http_lua_directive.c
```diff
@@ -597,19 +604,22 @@ ngx_http_lua_access_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
             return NGX_CONF_ERROR;
         }
 
+        chunkname_len = strlen((char *) chunkname);
```
agentzh: Ditto.

src/ngx_http_lua_directive.c
```diff
@@ -711,6 +722,7 @@ ngx_http_lua_content_by_lua(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
...
```
agentzh: Ditto.

src/ngx_http_lua_directive.c
```diff
@@ -936,13 +954,17 @@ ngx_http_lua_header_filter_by_lua(ngx_conf_t *cf, ngx_command_t *cmd,
...
+        p = ngx_palloc(cf->pool,
+                       sizeof("header_filter_by_lua") +
+                       NGX_HTTP_LUA_INLINE_KEY_LEN + 1);
         if (p == NULL) {
             return NGX_CONF_ERROR;
         }
 
         llcf->header_filter_src_key = p;
 
+        p = ngx_copy(p, "header_filter_by_lua",
```
agentzh: We should generate a chunk name here and use that. It seems to be another issue that we failed to use the new `ngx_http_lua_gen_chunk_name()` mechanism for these directives.

spacewander: @agentzh 
Yes, the code in `header_filter_by_lua` uses a hard code `=header_filter_by_lua` instead of generated chunk name.
So with `=balancer_by_lua` and some other phases.
Should we change the hard code chunk names and use the `ngx_http_lua_gen_chunk_name()` here too?

src/ngx_http_lua_directive.c
spacewander: Good point. Maybe we could change it to:
```
static size_t
ngx_http_lua_gen_chunk_name(ngx_conf_t *cf, const char *tag, size_t tag_len, u_char *chunkname)
```

src/ngx_http_lua_directive.c
agentzh: @spacewander Yes, please.

src/ngx_http_lua_directive.c
agentzh: @spacewander `chunkname` pointer is allocated inside this C function, so your prototype cannot return a new pointer but only use a caller-supplied buffer, which does not look right to me (since in that case, you also need to pass in the caller buffer size as an argument).

The following makes more sense to me:

```C
static u_char *
ngx_http_lua_gen_chunk_name(ngx_conf_t *cf, const char *tag, size_t tag_len,
size_t *chunkname_len);
```

It's similar to the standard Lua C API function, `lua_tolstring()`:

```C
const char *lua_tolstring (lua_State *L, int index, size_t *len);
```

src/ngx_http_lua_directive.c
spacewander: All `xxx_by_lua:...:` need to be rewritten to `xxx_by_lua(nginx.conf:...):...`.

agentzh: @spacewander Yes, definitely. I think we can first merge this PR and address the chunk name thing in a later separate PR.

agentzh:
@spacewander Merged. Thanks!

@dndx  ngx_meta_lua and ngx_stream_lua may need similar fixes.

PR #1238, opened 2018-01-22
bugfix: corrected the table size of req socket

spacewander:
The req socket uses two slots in array and three slots in hash at most.

At first it uses one array slot and one hash slot:
array part: upstream udata
hash part: __index

After calling settimeouts, it uses two array slot and three hash slot:
array part:
    [1] upstream udata
    [2] connect timeout
hash part:
    __index
    [4] send timeout
    [5] read timeout

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -4390,7 +4390,7 @@ ngx_http_lua_req_socket(lua_State *L)
         r->request_body = rb;
     }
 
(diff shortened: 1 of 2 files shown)


agentzh:
@spacewander Will you please add a test case to the existing `t/062-count.t` test file to cover this table size?

agentzh:
@spacewander Merged with edits in your commit log message. Now the title reads:

"optimize: corrected the initial table size of req socket objects."

Also fixed some places where "slots" should be used instead of "slot" in the message body :)

Thanks!

PR #1241, opened 2018-01-25
NDK moved to 'simplresty' account

mclyne:
The NDK can now be found at http://github.com/simplresty/ngx_devel_kit, along with other Nginx modules that were under the 'simpl' account.

Forked and sync'ed copies will still be available at http://github.com/simpl/ngx_devel_kit to prevent issues.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -270,7 +270,7 @@ It is *highly* recommended to use [OpenResty releases](http://openresty.org) whi
 Alternatively, ngx_lua can be manually compiled into Nginx:
 
 1. Install LuaJIT 2.0 or 2.1 (recommended) or Lua 5.1 (Lua 5.2 is *not* supported yet). LuaJIT can be downloaded from the [LuaJIT project website](http://luajit.org/download.html) and Lua 5.1, from the [Lua project website](http://www.lua.org/).  Some distribution package managers also distribute LuaJIT and/or Lua.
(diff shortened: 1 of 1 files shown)


agentzh:
@mclyne Ah, renaming is indeed a big inconvenient given that we have several modules using NDK, for example:

https://github.com/openresty/set-misc-nginx-module

https://github.com/openresty/encrypted-session-nginx-module

https://github.com/calio/iconv-nginx-module

https://github.com/calio/form-input-nginx-module

agentzh:
@mclyne Why don't you simply *move* the old repo over to your new github account? That way, github still provides a redirect for the old address and we do not have to update all the docs for all the NDK user modules, which is a lot of work for something not interesting at all...

mclyne:
@agentzh I'd already transfered the repo over to the simplresty account.  I didn't know about the automatic redirects, but have now deleted the forked repos, and redirects are working as you suggest.

Don't worry about updating the docs yourself if you don't want to.  I don't mind going through all the modules and making the changes myself when I have time - and send you pull requests.  It's not a vital thing to do, and it's my choice to make the change after all.

Sorry for the inconvenience.  I thought for a long time before changing, but given my plans going forward, it made more sense to move it over to the simplresty organization.

Obviously, let me know if you have any issues.

agentzh:
@mclyne To quote the file header of `README.markdown`:

```
<!---
-->
```

mclyne:
@agentzh Sorry about that - I obviously didn't notice it.  I'll make the edits in another pull request, then.

PR #1250, opened 2018-02-02
improvement: removed antiquated feature test

spacewander:
The NGX_HTTP_LUA_HAVE_CONSTRUCTOR is used to preserve memory for LuaJIT.
Since we don't do it anymore, this feature test could be removed.


agentzh:
@spacewander It's fine to keep it there just in case we will need it again in the future :)

PR #1254, opened 2018-02-08
Socket gettimeouts

hongliang5316:
We can get tcp socket `timeouts` setting more convenient.

 hereby granted the copyright of the changes in this pull request


agentzh:
@hongliang5316 Sorry, I'm not convinced that this is very useful for the general user community. Will you elaborate your use case for this? We are careful about adding new Lua APIs to the core since every new addition adds a little bit of GC overhead and startup overhead.

PR #1259, opened 2018-02-16
feature: implemented a configure_by_lua phase

thibaultcha:

This phase's purpose would be to allow the configuration of NGINX/ngx_lua with Lua scripts. This PR proposes a first draft with _just a few_ APIs, hoping that more APIs can be added progressively.

Ultimately, I cherish the vision to some day being able to dynamically configure `http`, `location`, `stream`, and other blocks right from Lua.

Example of what is currently implemented:
```nginx
http {
    configure_by_lua_block {
        local ngx_configure = require "ngx.configure"

        -- hard-coded, but could be retrieved from I/O or env
        local shms = {
            { "dogs", "12k" },
            { "cats", "24k" }
        }

        -- create shms
        for _, shm in ipairs(shms) do
            ngx_configure.shared_dict(shm[1], shm[2])
        end

        -- conditionally expose env variables for workers
        ngx_configure.env("PROXY_PASS_URL")

        -- configure the Lua VM
        ngx_configure.max_pending_timers(1024)
        ngx_configure.max_running_timers(128)
    }

    init_by_lua_block {
        ngx.shared.dogs:set("hello", true)
    }
}
```

List of what is currently implemented:

- [x] a new `configure_by_lua` phase (block/file) for `http` context
- ngx_lua directives
  - [x] `ngx_configure.shared_dict()` -> `lua_shared_dict`
  - [x] `ngx_configure.max_pending_timers()` -> `lua_max_pending_timers`
  - [x] `ngx_configure.max_running_timers()` -> `lua_max_running_timers`
- NGINX directives
  - [x] `ngx_configure.env()` -> `env`
- [x] no NGINX patch required
- [x] naming convention of APIs strips `lua_` prefix from directives

Everything is implemented via the LuaJIT FFI in the lua-resty-core sister PR.

My current goal is to get feedback, reviews, guidance... anything :) In the long term, I wish to gradually add more relevant API to this phase and make OpenResty more dynamic. If anyone can think of anything in the current implementation would eventually impact this goal or limit future possibilities, please let me know.


spacewander:
@thibaultcha 
Sorry for mentioning one thing: Why don't you try to make the Nginx configure file as a template, and render it before starting/reloading the server?
The team I last worked with did this in production over one year ago, and it works.

spacewander:
PS. You could use [priviledge process](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/process.md#enable_privileged_agent) to detect configure change and send reload signal. And the template could be implemented in Lua, so no other dependency is required.

thibaultcha:
@spacewander This is precisely what we have been doing for the past few years, and what this PR tries to get rid of :) 

The goal here is to avoid having to depend so heavily on templating all the time to configure our applications. Additionally, templating causes huge headaches to users who need to make their own modifications to their Nginx configuration file, not only to learn it but also for each upgrade with configuration changes. Really, the goal here is to not have to do that anymore.

thibaultcha:
Especially considering the maintainable cost seems to be relatively low. No Nginx patch required, a dependency on an init, process that is very unlikely to ever change, and a self-contained patch in the codebase. I believe the ability to configure servers and location blocks would be possible by respecting the same constraints as well.

src/ngx_http_lua_configureby.c
```diff
@@ -0,0 +1,287 @@
...
+#include "ngx_http_lua_util.h"
+#include "ngx_http_lua_shdict.h"
+#include "ngx_http_lua_api.h"
+
+
+static ngx_conf_t *cfp;
+
+
+char *
+ngx_http_lua_configure_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
```
agentzh: I don't like all the code duplications in this C source file. It makes the code base harder to maintain and sync new changes.

src/ngx_http_lua_shdict.c
```diff
@@ -320,93 +321,113 @@ ngx_http_lua_shdict_expire(ngx_http_lua_shdict_ctx_t *ctx, ngx_uint_t n)
...
-ngx_http_lua_inject_shdict_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
+ngx_http_lua_create_shdict_mt(lua_State *L)
 {
-    ngx_http_lua_shdict_ctx_t   *ctx;
-    ngx_uint_t                   i;
-    ngx_shm_zone_t             **zone;
+    lua_createtable(L, 0 /* narr */, 18 /* nrec */);
+        /* ngx ngx.shared shmt */
 
-    if (lmcf->shdict_zones != NULL) {
```
agentzh: This also makes such optimizations no longer exist. I'm kinda proud of the concept that you do not pay for the features you never use :)

thibaultcha: What do you mean? We only create the shdict metatable if we need it still (via `lua_shared_dict` or `ngx_configure.shared_dict()`)

src/ngx_http_lua_configureby.c
thibaultcha: I can definitely improve it :) Which area? The context runner and/or the shm configure API?

src/ngx_http_lua_shdict.c
agentzh: @thibaultcha We used to save the metatable when lua_shared_dict is not configured in `nginx.conf`. This is a minor thing though.

src/ngx_http_lua_configureby.c
agentzh: @thibaultcha Almost all the code in this C source file is duplicated C code.

src/ngx_http_lua_shdict.c
thibaultcha: > We used to save the metatable when lua_shared_dict is not configured in nginx.conf.

And we still do, I made sure of that, this method has been renamed to reflect the change but both of its calls in the codebase are enclosed within conditions. I could re-add the safeguard condition in here too just in case.

> It seems that you essentially created a second parallel configuring code path for those config directives we already have.

All of the Lua phases code is duplicated already, I just followed the pattern.
Regarding the add shm API, I can certainly improve it - I originally didn't want to modify existing code too much by fear the patch would be refused if so.
Regarding the timers code, I was thinking of coming up with a more generic way of setting integer values onto the `lmcf` structure (we could follow the NGINX pattern with `ngx_conf_set_num_slot` and the likes) but that seems unnecessary for only 2 fields as of today and considering it would result in a lot more code.

src/ngx_http_lua_configureby.c
thibaultcha: @agentzh I just added a new commit which introduces `ngx_http_lua_shared_dict_add`, which factorizes all of the shared code for shm create for `lua_shared_dict` and `ngx_configure.shared_dict()`.

As stated before, the code related to parsing and running the `_by_*_block` and `_by_*_file` directives of this module are already always duplicated - this feels outside the scope of this PR, but if you want me to address this, please advise.

src/ngx_http_lua_directive.c
```diff
@@ -73,84 +73,38 @@ enum {
...
+        }
 
-    zp = ngx_array_push(lmcf->shdict_zones);
-    if (zp == NULL) {
         return NGX_CONF_ERROR;
     }
 
-    *zp = zone;
-
-    lmcf->requires_shm = 1;
```
thibaultcha: This was duplicated (as in the master branch) - we now only set it from `ngx_http_lua_shared_memory_add`.

src/ngx_http_lua_api.c
```diff
@@ -122,9 +121,7 @@ ngx_http_lua_shared_memory_add(ngx_conf_t *cf, ngx_str_t *name, size_t size,
         return &ctx->zone;
     }
 
-    n = sizeof(ngx_http_lua_shm_zone_ctx_t);
-
-    ctx = ngx_pcalloc(cf->pool, n);
```
thibaultcha: Just cleaning up here but I can revert if not desired.

src/ngx_http_lua_configureby.c
agentzh: @thibaultcha 

>All of the Lua phases code is duplicated already, I just followed the pattern.

What exactly are you referring to?

The thing is if the config phase handlers can call these functions, then we should make it that way. I don't want to maintain 2 duplicated copies of configurations. Also, how do you handle config directive overriding and inheritance here? Seems like you already bypass the nginx config system entirely?

agentzh:
@thibaultcha I appreciate your exploration in this direction. I'm thinking about maybe we could make this work on a much lower level, by dynamically configuring *all* the nginx main/http/stream/server/location blocks in Lua. Your approach looks like we will simply duplicate all the existing nginx configuraiton directives in this new facility, which does not have a future IMHO (not to mention there might be an infinite number of 3rd-party nginx modules out there providing their infinite number of possible directive that could interact with OpenResty) :)

If we want to get rid of nginx.conf, then let's do it in a thorough way :) Just my 2 cents.

src/ngx_http_lua_configureby.c
thibaultcha: > how do you handle config directive overriding and inheritance here? Seems like you already bypass the nginx config system entirely?

There is no inheritance involved here since we only support configuration of `http` (or above) level directives this is deliberate.

Ultimately, I do wish to support lower levels of configuration as explained in the PR (`server`, `location`...), but without it there are no need for supporting inheritance yet.

> What exactly are you referring to?

I was reffering to the only part of the code that I see is duplicated from other parts of ngx_lua, and that is the `ngx_http_lua_configure_by_lua` and other handlers that run the Lua code for this phase.

If not, then I am not sure what you are referring to when you are saying things are duplicated?

thibaultcha:
> we could make this work on a much lower level, by dynamically configuring all the nginx main/http/stream/server/location blocks in Lua

Yes, this is exactly my point and my intent on the long term (as explained above). The current patch is more of an intermediate phase for the nginx `main` and `http` context directives, because I am of the opinion that we should build this step-by-step, and add more APIs (including APIs dynamically adding server and location blocks) to it in the future.

> Your approach looks like we will simply duplicate all the existing nginx configuraiton directives in this new facility, which does not have a future IMHO

Maybe not _all_ :) This approach would already be quite helpful imho - we've been wanting for years to only allocate the lua_shared_dict that our customers' dynamically configured instances need, without requiring of them that they learn or modify our nginx configuration templating system (that we wish to get rid of in the long term). I believe a lot of users would be happy to allocate shared dicts from the Lua-land (I have witnessed many asks for it over the years in the mailing list, issues, or chats...).

> not to mention there might be an infinite number of 3rd-party nginx modules out there providing their infinite number of possible directive that could interact with OpenResty

Agreed. But it is not because we cannot please everybody that we should please nobody :) Covering the directives that are used the most often (like `env`, `lua_shared_dict` for starters) would already satisfy a lot of users imho. Besides, how do you intend to support the much larger amount of available directives in the `server` and `location` contexts?

One solution I thought of (I did not explore its feasibility) was to avoid creating structures representing the state of the configuration like this:

```lua
ngx.server.my_server:add_location("/")
```

Which would become a maintenance nightmare, but instead supporting a more lose API, like:

```lua
ngx.configure.append [[
server {
    listen :8080;
    
    location {
        return 200;
    }
}
]]
```

Maybe the final solution will be something in the middle.

thibaultcha:
I very much want to highlight the following sentences from this PR's opening thread:

> This phase's purpose would be to allow the configuration of NGINX/ngx_lua with Lua scripts. This PR proposes a first draft with just a few APIs, hoping that more APIs can be added progressively.

> Ultimately, I cherish the vision to some day being able to dynamically configure http, location, stream, and other blocks right from Lua.

The current proposal is to iterate slowly towards that goal with a few APIs that would already be helpful. If you prefer that idea I have outlined above, or if you have a better one altogether that you wish to share, or if dynamically configuring nginx is not something that OpenResty wishes to achieve, I am fine with this, please do let me know! :)

agentzh:
@thibaultcha I do understand your intention but unfortunately I don't see this PR as an iterative approach to the right direction from the perspective of an implementator. This patch adds quite some maintenance overhead with very little real world merit IMHO.

Someguynamedpie:
Down the line, being able to append or "execute" blocks of NGINX config (even as strings to lower the implementation requirements) from init_by_lua would be a godsend for server management tools.

PR #1266, opened 2018-02-21
Switch out TLS1_get_version for SSL_version

tmthrgd:
`TLS1_get_version` is a simple wrapper for `SSL_version` that returns 0 when used with DTLS. However, it was removed from BoringSSL in google/boringssl@593047fd8056e40b690f5b695ad35a4b09f9f9bb (2015), which breaks [`ngx.ssl.get_tls1_version`](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.md#get_tls1_version). This PR replaces it with `SSL_version` directly instead.

`ngx_http_lua_ffi_ssl_get_tls1_version` can never be reached with DTLS so the behaviour is the same.

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -553,13 +553,6 @@ ngx_http_lua_ssl_cert_by_chunk(lua_State *L, ngx_http_request_t *r)
 int
 ngx_http_lua_ffi_ssl_get_tls1_version(ngx_http_request_t *r, char **err)
 {
-#ifndef TLS1_get_version
-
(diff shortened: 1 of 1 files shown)


spacewander:
One extra note: we are able to support DTLS with this change in the future, by adding DTLS's version constants in:
https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.lua#L291

Well, just in the future.

agentzh:
@tmthrgd Merged with minor edits in your commit log message. Thanks! But please note: BoringSSL is never an officially supported target for this module :)

PR #1269, opened 2018-02-28
bugfix: avoid memory leak when set_serialized_session called repeatly

spacewander:

The way to reproduce the memory leak:
Run `TEST_NGINX_USE_VALGRIND=1 prove t/ssl-session-fetch.t` in the pull request https://github.com/openresty/lua-resty-core/pull/177


src/ngx_http_lua_ssl_session_fetchby.c
```diff
@@ -597,8 +598,13 @@ ngx_http_lua_ffi_ssl_set_serialized_session(ngx_http_request_t *r,
         return NGX_ERROR;
     }
 
+    old_session = cctx->session;
     cctx->session = session;
 
+    if (old_session) {
```
agentzh: We should test a pointer as a pointer instead of as a boolean. The recommended way is to write `old_session != NULL` here. I fixed it in the merged version.

agentzh:
@spacewander Merged with minor fixes. Thanks!

PR #1270, opened 2018-03-01
feature: added queue to connection pool

spacewander:
Now we could pass 'pool_size' and 'backlog' options in connect method to
specify the pool size and the queue size of connection pool.

When there is more connect operations than the pool_size, new connect
operation will be queued until former connect operation fails or
connection finalizes.

When there is more queueing connect operations than the backlog, new
connect operation will be dropped.

This pr also migrated connection pool creation to connect method.
Note that the meaning of `active_connections` is changed from the previous
setkeepalive mechanism. The connection pool is destroyed only if all related
connections closed now.

The implementation of tcp cosocket is quite complex, therefore both strict
code review and extreme testing are required before merging this feature into
master.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -7091,6 +7091,43 @@ An optional Lua table can be specified as the last argument to this method to sp
 * `pool`
 	specify a custom name for the connection pool being used. If omitted, then the connection pool name will be generated from the string template `"<host>:<port>"` or `"<unix-socket-path>"`.
 
+* `pool_size`
+	specify the size of the connection pool. If omitted and no
+	`backlog` option was provided, no pool will be created. If omitted
+	but `backlog` was provided, the pool will be created with a default
+	size equal to the value of the [lua_socket_pool_size](#lua_socket_pool_size)
+	directive.
+	The connection pool holds up to `pool_size` alive connections
+	ready to be reused by subsequent calls to [connect](#tcpsockconnect), but
+	note that there is no upper limit to the total number of opened connections
+	outside of the pool. If you need to restrict the total number of opened
+	connections, specify the `backlog` option.
+	When the connection pool would exceed its size limit, the least recently used
+	(kept-alive) connection already in the pool will be closed to make room for
+	the current connection.
+	Note that the cosocket connection pool is per Nginx worker process rather
+	than per Nginx server instance, so the size limit specified here also applies
+	to every single Nginx worker process. Also note that the size of the connection
+	pool cannot be changed once it has been created.
+	This option was first introduced in the `v0.10.14` release.
+
+* `backlog`
+	if specified, this module will limit the total number of opened connections
+	for this pool. No more connections than `pool_size` can be opened
+	for this pool at any time. If the connection pool is full, subsequent
+	connect operations will be queued into a queue equal to this option's
+	value (the "backlog" queue).
+	If the number of queued connect operations is equal to `backlog`,
+	subsequent connect operations will fail and return `nil` plus the
+	error string `"too many waiting connect operations"`.
+	The queued connect operations will be resumed once the number of connections
+	in the pool is less than `pool_size`.
+	The queued connect operation will abort once they have been queued for more
+	than `connect_timeout`, controlled by
+	[settimeouts](#tcpsocksettimeouts), and will return `nil` plus
+	the error string `"timeout"`.
+	This option was first introduced in the `v0.10.14` release.
+
 The support for the options table argument was first introduced in the `v0.5.7` release.
 
 This method was first introduced in the `v0.5.0rc1` release.
@@ -7422,13 +7459,31 @@ Puts the current socket's connection immediately into the cosocket built-in conn
 
 The first optional argument, `timeout`, can be used to specify the maximal idle timeout (in milliseconds) for the current connection. If omitted, the default setting in the [lua_socket_keepalive_timeout](#lua_socket_keepalive_timeout) config directive will be used. If the `0` value is given, then the timeout interval is unlimited.
 
-The second optional argument, `size`, can be used to specify the maximal number of connections allowed in the connection pool for the current server (i.e., the current host-port pair or the unix domain socket file path). Note that the size of the connection pool cannot be changed once the pool is created. When this argument is omitted, the default setting in the [lua_socket_pool_size](#lua_socket_pool_size) config directive will be used.
-
-When the connection pool exceeds the available size limit, the least recently used (idle) connection already in the pool will be closed to make room for the current connection.
-
-Note that the cosocket connection pool is per Nginx worker process rather than per Nginx server instance, so the size limit specified here also applies to every single Nginx worker process.
-
-Idle connections in the pool will be monitored for any exceptional events like connection abortion or unexpected incoming data on the line, in which cases the connection in question will be closed and removed from the pool.
+The second optional argument `size` is considered deprecated since
+the `v0.10.14` release of this module, in favor of the
+`pool_size` option of the [connect](#tcpsockconnect) method.
+Since the `v0.10.14` release, this option will only take effect if
+the call to [connect](#tcpsockconnect) did not already create a connection
+pool.
+When this option takes effect (no connection pool was previously created by
+[connect](#tcpsockconnect)), it will specify the size of the connection pool,
+and create it.
+If omitted (and no pool was previously created), the default size is the value
+of the [lua_socket_pool_size](#lua_socket_pool_size) directive.
+The connection pool holds up to `size` alive connections ready to be
+reused by subsequent calls to [connect](#tcpsockconnect), but note that there
+is no upper limit to the total number of opened connections outside of the
+pool.
+When the connection pool would exceed its size limit, the least recently used
+(kept-alive) connection already in the pool will be closed to make room for
+the current connection.
+Note that the cosocket connection pool is per Nginx worker process rather
+than per Nginx server instance, so the size limit specified here also applies
+to every single Nginx worker process. Also note that the size of the connection
+pool cannot be changed once it has been created.
+If you need to restrict the total number of opened connections, specify both
+the `pool_size` and `backlog` option in the call to
+[connect](#tcpsockconnect).
 
 In case of success, this method returns `1`; otherwise, it returns `nil` and a string describing the error.
 
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -100,12 +100,28 @@ static int ngx_http_lua_req_socket(lua_State *L);
 static void ngx_http_lua_req_socket_rev_handler(ngx_http_request_t *r);
 static int ngx_http_lua_socket_tcp_getreusedtimes(lua_State *L);
 static int ngx_http_lua_socket_tcp_setkeepalive(lua_State *L);
+static void ngx_http_lua_socket_tcp_create_socket_pool(lua_State *L,
+    ngx_http_request_t *r, ngx_str_t key, ngx_int_t pool_size,
+    ngx_int_t backlog, ngx_http_lua_socket_pool_t **spool);
 static ngx_int_t ngx_http_lua_get_keepalive_peer(ngx_http_request_t *r,
-    lua_State *L, int key_index,
     ngx_http_lua_socket_tcp_upstream_t *u);
 static void ngx_http_lua_socket_keepalive_dummy_handler(ngx_event_t *ev);
+static int ngx_http_lua_socket_tcp_connect_helper(lua_State *L,
+    ngx_http_lua_socket_tcp_upstream_t *u, ngx_http_request_t *r,
+    ngx_http_lua_ctx_t *ctx, u_char *host_ref, size_t host_len, in_port_t port,
+    unsigned resuming);
+static void ngx_http_lua_socket_tcp_conn_op_timeout_handler(
+    ngx_event_t *ev);
+static int ngx_http_lua_socket_tcp_conn_op_timeout_retval_handler(
+    ngx_http_request_t *r, ngx_http_lua_socket_tcp_upstream_t *u, lua_State *L);
+static void ngx_http_lua_socket_tcp_resume_conn_op(
+    ngx_http_lua_socket_pool_t *spool);
+static void ngx_http_lua_socket_tcp_conn_op_ctx_cleanup(void *data);
+static void ngx_http_lua_socket_tcp_conn_op_resume_handler(ngx_event_t *ev);
 static ngx_int_t ngx_http_lua_socket_keepalive_close_handler(ngx_event_t *ev);
 static void ngx_http_lua_socket_keepalive_rev_handler(ngx_event_t *ev);
+static int ngx_http_lua_socket_tcp_conn_op_resume_retval_handler(
+    ngx_http_request_t *r, ngx_http_lua_socket_tcp_upstream_t *u, lua_State *L);
 static int ngx_http_lua_socket_tcp_upstream_destroy(lua_State *L);
 static int ngx_http_lua_socket_downstream_destroy(lua_State *L);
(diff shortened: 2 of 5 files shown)


agentzh:
@spacewander Needs some doc updates to reflect the API changes in this PR (backward-incompatible changes, actually).

src/ngx_http_lua_socket_tcp.c
```diff
@@ -148,6 +165,7 @@ enum {
 
 
 enum {
+    SOCKET_OP_RESUME_CONN,
```
agentzh: Better add new enum constants to the end so that we do not change the values of those existing enum constants.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -5332,15 +5862,21 @@ ngx_http_lua_socket_tcp_resume_helper(ngx_http_request_t *r, int socket_op)
 
     dd("coctx: %p", coctx);
 
-    u = coctx->data;
-
     switch (socket_op) {
+    case SOCKET_OP_RESUME_CONN:
```
agentzh: Please add a blank line before the first `case` statement to better follow the nginx coding style. I understand the original code seems to already have the same problem, but it's good to fix things when we touch the same part of the code.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -5354,6 +5890,13 @@ ngx_http_lua_socket_tcp_resume_helper(ngx_http_request_t *r, int socket_op)
                    "u:%p", prepare_retvals, u);
 
     nret = prepare_retvals(r, u, ctx->cur_co_ctx->co);
+    if (socket_op == SOCKET_OP_CONNECT && nret > 1 && !u->conn_closed
+        && u->socket_pool != NULL)
```
agentzh: When the if condition is split, please use a more readable form like below:

```C
if (A
    && B
    && C
    && D)
{
```

t/068-socket-keepalive.t
```diff
@@ -1551,3 +1552,1239 @@ done
...
+received: OK
+connected: 1, reused: 1
+request sent: 11
+received: OK
+reused: 1
+reused: 0
+--- no_error_log eval
+["[error]",
+"lua tcp socket keepalive: free connection pool for ",
+"lua tcp socket connection pool size: 20"]
```
agentzh: We'd better issue a warning log message for ignoring the pool size specified in the old way (i.e., via `:setkeepalive()` method call's argument).

t/068-socket-keepalive.t
```diff
@@ -1551,3 +1552,1239 @@ done
...
+--- grep_error_log eval: qr/(start|continue) to handle \w+/
+--- grep_error_log_out
+start to handle timer
+start to handle cosocket
+continue to handle timer
+continue to handle cosocket
+
+
+
+=== TEST 32: don't count failed connect operation
```
agentzh: "operation" => "operations"

src/ngx_http_lua_socket_tcp.c
```diff
@@ -414,31 +432,414 @@ ngx_http_lua_socket_tcp(lua_State *L)
...
+        r->write_event_handler = ngx_http_core_run_phases;
+    }
+
+    if (resuming) {
+        return NGX_AGAIN;
+    }
+
+    return lua_yield(L, 0);
+
+failure:
```
agentzh: The nginx core usually uses the code label `failed`. Let's follow that convention here as well.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4656,67 +5216,31 @@ ngx_http_lua_socket_tcp_setkeepalive(lua_State *L)
...
-        key_len = ngx_align(key.len + 1, sizeof(void *));
-
-        size = sizeof(ngx_http_lua_socket_pool_t) + key_len - 1
-               + sizeof(ngx_http_lua_socket_pool_item_t)
-               * pool_size;
-
-        spool = lua_newuserdata(L, size);
-        if (spool == NULL) {
-            return luaL_error(L, "no memory");
+        if (pool_size <= 0) {
```
agentzh: I think we should issue a warning log message saying `argument "pool_size" ignored; use new() method instead`.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4886,7 +5387,6 @@ ngx_http_lua_get_keepalive_peer(ngx_http_request_t *r, lua_State *L,
             cln = ngx_http_lua_cleanup_add(r, 0);
             if (cln == NULL) {
                 u->ft_type |= NGX_HTTP_LUA_SOCKET_FT_ERROR;
-                lua_settop(L, top);
```
agentzh: Why remove this? This is a security measure. Just to make sure the Lua stack is always balanced for this C function call.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4895,16 +5395,12 @@ ngx_http_lua_get_keepalive_peer(ngx_http_request_t *r, lua_State *L,
             u->cleanup = &cln->handler;
         }
 
-        lua_settop(L, top);
```

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4895,16 +5395,12 @@ ngx_http_lua_get_keepalive_peer(ngx_http_request_t *r, lua_State *L,
...
-
         return NGX_OK;
     }
 
     ngx_log_debug0(NGX_LOG_DEBUG_HTTP, pc->log, 0,
                    "lua tcp socket keepalive: connection pool empty");
 
-    lua_settop(L, top);
```

src/ngx_http_lua_socket_tcp.c
There is no need to call `lua_settop`, since we don't touch Lua VM in `ngx_http_lua_get_keepalive_peer` now. We just use `u->socket_pool` directly.

spacewander:
@agentzh 
Updated.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -165,10 +165,10 @@ enum {
 
 
 enum {
-    SOCKET_OP_RESUME_CONN,
     SOCKET_OP_CONNECT,
     SOCKET_OP_READ,
-    SOCKET_OP_WRITE
+    SOCKET_OP_WRITE,
+    SOCKET_OP_RESUME_CONN,
```
agentzh: Can we remove the trailing comma? It's new in C99.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -5230,6 +5230,10 @@ ngx_http_lua_socket_tcp_setkeepalive(lua_State *L)
          * so we increase connections here for backward compatibility.
          */
         spool->connections++;
+
+    } else if (n >= 3) {
+        ngx_log_error(NGX_LOG_WARN, pc->log, 0, "argument \"pool_size\" "
+                      "already specified in connect() method");
```
agentzh: @spacewander So you still want to support the pool_size argument in the `setkeepalive()` method?

Yes. The `pool_size` argument in `connect` method is experimental. Maybe it is better to be conservative.

src/ngx_http_lua_socket_tcp.c
Comma removed.

spacewander:
@agentzh 
I reverted the `pool_size` specified warning, because calling `setkeepalice` twice will also trigger it.
And there is not existent way to distinguish them.

agentzh:
@spacewander OK.

agentzh:
@spacewander Documentation is still missing for this change?

spacewander:
@agentzh 
Documentation is added.
Because the version of next release has not been determined, I use `<code>dev</code> release` as placeholder.

agentzh:
@spacewander I'd really like to include it in OpenResty 1.13.6.2 final :)

README.markdown
```diff
@@ -7074,6 +7074,24 @@ An optional Lua table can be specified as the last argument to this method to sp
...
 	specify a custom name for the connection pool being used. If omitted, then the connection pool name will be generated from the string template `"<host>:<port>"` or `"<unix-socket-path>"`.
 
+* `pool_size`
+	specify the size of connection pool being used. Note that the size of the connection pool cannot be changed once the pool is created. If omitted,
+then the default setting in the [lua_socket_pool_size](#lua_socket_pool_size) config directive will be used.
+When the connection pool exceeds the available size limit, the least recently used (idle) connection already in the pool will be closed to make room for the current connection.
+Note that the cosocket connection pool is per Nginx worker process rather than per Nginx server instance, so the size limit specified here also applies to every single Nginx worker process.
+If the `backlog` is not specified, then the size of connection pool is only
+for idle connections. If you need to restrict the number of active connections in the pool, specify the `backlog` as well.
+This option was first introduced in the `dev` release.
```
agentzh: We need to add one level of indentation to these paragraphs otherwise they would break the current bullet list.

README.markdown
```diff
@@ -7074,6 +7074,24 @@ An optional Lua table can be specified as the last argument to this method to sp
...
+
+* `backlog`
+	specify the size of queue being used. If specified, once the number of connections in the pool is equal to `pool_size`, new connect operations will be queued.
+If the number of queued connect operations is equal to `backlog`, new connect operations will fail and return `nil` plus
+`"too many waiting connect operations"`.
+The queued connect operations will be resumed once the number of connections in the pool is less than `pool_size`.
+The queued connect operation will abort once they have been queued for more than `connect_timeout` controlled by
+[settimeouts](#tcpsocksettimeouts), return `nil` plus `"timeout"`.
+This option was first introduced in the `dev` release.
```
agentzh: Ditto.

README.markdown
```diff
@@ -7074,6 +7074,24 @@ An optional Lua table can be specified as the last argument to this method to sp
...
+for idle connections. If you need to restrict the number of active connections in the pool, specify the `backlog` as well.
+This option was first introduced in the `dev` release.
+
+* `backlog`
+	specify the size of queue being used. If specified, once the number of connections in the pool is equal to `pool_size`, new connect operations will be queued.
+If the number of queued connect operations is equal to `backlog`, new connect operations will fail and return `nil` plus
+`"too many waiting connect operations"`.
+The queued connect operations will be resumed once the number of connections in the pool is less than `pool_size`.
+The queued connect operation will abort once they have been queued for more than `connect_timeout` controlled by
+[settimeouts](#tcpsocksettimeouts), return `nil` plus `"timeout"`.
```
agentzh: Grammar: "return" => "and will also return".

spacewander:
@agentzh 
Documentation updated.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4800,48 +5320,29 @@ ngx_http_lua_socket_tcp_setkeepalive(lua_State *L)
     ngx_http_lua_socket_tcp_finalize(r, u);
 #endif
 
+    ngx_http_lua_socket_tcp_resume_conn_op(spool);
```
thibaultcha: The above `ngx_http_lua_socket_tcp_finalize` already calls `ngx_http_lua_socket_tcp_resume_conn_op` (granted the pool is not empty). Is this second call intentional then?

spacewander: @thibaultcha 
Yes, the call is intentional.
The `ngx_http_lua_socket_tcp_resume_conn_op` won't be called by `ngx_http_lua_socket_tcp_finalize`, because we have set the `u->peer->connection` to `NULL` above. So we need to call it after `ngx_http_lua_socket_tcp_finalize`.
I will add a comment to clarify it.

spacewander:
@thibaultcha 
Fixed with your advice.

deem0n:
Any chance this feature will be merged ?
Currently I need to run HAProxy for queuing requests to upstream. That works but adds latency.

aloisbarreras:
+1, anything I can do to help get this merged?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -414,31 +432,413 @@ ngx_http_lua_socket_tcp(lua_State *L)
...
+failed:
+
+    if (spool != NULL) {
+        spool->connections--;
+        ngx_http_lua_socket_tcp_resume_conn_op(spool);
+    }
+
+    return 2;
+
+no_memory_and_not_resuming:
```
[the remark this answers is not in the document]
spacewander: The helper may be called in `ngx_http_lua_socket_tcp_connect` directly, without resuming. If it is called in a resumption, we should not go into this label.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -466,13 +866,54 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
...
 
         /* found the last optional option table */
 
+        lua_getfield(L, n, "pool_size");
+
+        if (lua_isnumber(L, -1)) {
+            pool_size = (ngx_int_t) lua_tointeger(L, -1);
+
+            if (pool_size <= 0) {
+                msg = lua_pushfstring(L, "bad \"pool_size\" option value: %i",
```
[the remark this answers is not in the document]
thibaultcha: Yes, agreed, it isoutside of this PR’s scope.

spacewander:
@thibaultcha 
Rebased and updated.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -414,31 +432,413 @@ ngx_http_lua_socket_tcp(lua_State *L)
...
+           + sizeof(ngx_http_lua_socket_pool_item_t) * pool_size;
+
+    /* before calling this function, the Lua stack is:
+     * -1 key
+     * -2 pools
+     */
+    sp = lua_newuserdata(L, size);
+    if (sp == NULL) {
+        luaL_error(L, "no memory");
+        return;
```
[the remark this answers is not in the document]
thibaultcha: Ok, makes perfect sense. Yes, I know it is just an idiom.

thibaultcha:
@spacewander Thanks! I will have a final look ASAP, currently a bit unavailable as I am traveling.

spacewander:
@thibaultcha 
Updated.

README.markdown
```diff
@@ -7422,13 +7438,8 @@ Puts the current socket's connection immediately into the cosocket built-in conn
...
-
-When the connection pool exceeds the available size limit, the least recently used (idle) connection already in the pool will be closed to make room for the current connection.
-
-Note that the cosocket connection pool is per Nginx worker process rather than per Nginx server instance, so the size limit specified here also applies to every single Nginx worker process.
-
-Idle connections in the pool will be monitored for any exceptional events like connection abortion or unexpected incoming data on the line, in which cases the connection in question will be closed and removed from the pool.
+The second optional argument `size` is deprecated. It is used to specify the maximal number of connections allowed in the connection pool for the current server (i.e., the current host-port pair or the unix domain socket file path). Its use is discouraged since the dev release, and the `pool_size` option of the [connect](#tcpsockconnect) method is preferred instead.
```
thibaultcha: note for merge: update `dev` release

src/ngx_http_lua_socket_tcp.c
```diff
@@ -3571,6 +3877,257 @@ ngx_http_lua_socket_tcp_finalize_write_part(ngx_http_request_t *r,
...
+    coctx->cleanup = NULL;
+    /* note that we store conn_op_ctx in coctx->data instead of u */
+    coctx->data = conn_op_ctx;
+    u->write_co_ctx = NULL;
+
+    llcf = ngx_http_get_module_loc_conf(r, ngx_http_lua_module);
+
+    if (llcf->log_socket_errors) {
+        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                      "lua tcp socket queued connect timed out,"
```
thibaultcha: note for merge: remove comma

agentzh: @thibaultcha You can address such minor things yourself during the merge?

thibaultcha: Yes, will do.

Issue #1207, opened 2017-12-11
Deadlock in ngx_http_lua_shdict_set_helper


gohai:
Hello,

We am using OpenResty 1.13.6.1 (1.el7.centos) to terminate HTTPS traffic for multiple domains, using the [lua-resty-auto-ssl](https://github.com/GUI/lua-resty-auto-ssl) module.

One of our instances randomly stopped responding to traffic, and did not cover on its own. It turned out that both nginx worker processes appear to be wedged waiting for a futex.

Stacktraces on both processes look rather similar, although the details vary. Here is one: `ngx_http_lua_ssl_cert_handler -> ngx_http_lua_ssl_cert_by_chunk -> ngx_http_lua_run_thread -> ngx_http_lua_shdict_get_helper -> lua_pushlstring -> lj_gc_step -> gc_onestep -> gc_finalize -> gc_call_finalizer -> lj_BC_FUNCC -> ngx_http_lua_shdict_set_helper -> ngx_shmtx_lock -> sem_wait@@GLIBC_2.2.5`.

I am wondering whether the apparent recursion between `ngx_http_lua_shdict_get_helper` and `ngx_http_lua_shdict_set_helper` might have caused an issue, as I see both functions acquiring the same lock? Or is there some subtle race condition between the two worker processes?

Here is the stacktrace of the first worker process:

```
#0  0x00007f1ace2fda0b in do_futex_wait.constprop.1 ()
   from /lib64/libpthread.so.0
No symbol table info available.
#1  0x00007f1ace2fda9f in __new_sem_wait_slow.constprop.0 ()
   from /lib64/libpthread.so.0
No symbol table info available.
#2  0x00007f1ace2fdb3b in sem_wait@@GLIBC_2.2.5 () from /lib64/libpthread.so.0
No symbol table info available.
#3  0x000000000042e5f8 in ngx_shmtx_lock (mtx=0x7f1ac5de1068)
    at src/core/ngx_shmtx.c:111
        i = <optimized out>
        n = <optimized out>
#4  0x00000000004ef772 in ngx_http_lua_shdict_set_helper (L=0x413061a8, 
    flags=0) at ../ngx_lua-0.10.11/src/ngx_http_lua_shdict.c:1016
        i = <optimized out>
        n = <optimized out>
        key = {len = 31, data = 0x40dc1d70 "issue_cert:www.domainA.com"}
        rc = <optimized out>
        ctx = 0x23df390
[...]

PR #1272, opened 2018-03-08
7Feature/shdict zset

ZigzagAK:

Hello!
May be interresting DICT.zset feature (but without ffi)?


spacewander:
Thank you for your pull request.
But there are some problems I have to point out:
1. We are trying hard to migrate Lua C API to FFI, so adding more non-FFI API is not desiring. Unless there is not FFI infrastructure like cosocket stuff, but shdict is not. Not even mentioning that the current shdict's Lua C API implementation is unsafe, discussed in https://github.com/openresty/lua-nginx-module/issues/1207.
2. Implement zset with doubly linked list doesn't work well when people need to do `zrange` (`zrange` is not implemented yet, but we need to do it one day if we add zset to shdict type). Redis implements zset with skip list.

spacewander:
@ZigzagAK 
Apologize for my mistake. The `zset` is actually built on top of rbtree, so my second problem doesn't make sense at all.

ZigzagAK:
@spacewander 
I don't use ffi (lua-resty-core) api. In my use case i can't found performance improvements with FFI based api. Integral performance of my application with FFI looks like a performance with standart LUA C API (luajit).

spacewander:
@ZigzagAK 
Maybe the bottleneck of the application is not JIT.
Besides the performance, the shdict API implemented in Lua C API is unsafe. They hold shm lock when calling the Lua C API. When LuaJIT runs the Lua C API, it may do GC check and trigger the GC. If the shdict is accessed again via the `__gc` metamethod, there will be a deadlock.

PR #1273, opened 2018-03-09
feature: exposed ngx_http_lua_ffi_str_t for other Nginx modules.

spacewander:

### DIFF
--- a/src/api/ngx_http_lua_api.h
+++ b/src/api/ngx_http_lua_api.h
@@ -15,6 +15,7 @@
 #include <lua.h>
 #include <stdint.h>
 
+#include "ngx_http_lua_api_common.h"
 
 /* Public API for other Nginx modules */
 
--- a/config
+++ b/config
@@ -400,6 +400,7 @@ HTTP_LUA_DEPS=" \
             $ngx_addon_dir/src/ngx_http_lua_shdict.h \
             $ngx_addon_dir/src/ngx_http_lua_socket_tcp.h \
             $ngx_addon_dir/src/api/ngx_http_lua_api.h \
+            $ngx_addon_dir/src/api/ngx_http_lua_api_common.h \
             $ngx_addon_dir/src/ngx_http_lua_logby.h \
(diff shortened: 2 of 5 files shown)


src/api/ngx_http_lua_api.h
```diff
@@ -15,6 +15,7 @@
 #include <lua.h>
 #include <stdint.h>
 
+#include "ngx_http_lua_api_common.h"
```
agentzh: It's wrong to include `common.h` in `api.h` since it exposes all `ngx_lua`'s internal types and symbols, which is not desired.

spacewander: @agentzh 
No, `ngx_http_lua_api_common.h` is not `ngx_http_lua_common.h`.

agentzh: @spacewander Why introduce this new `common.h` file? It's confusing ;) Let's simply put everything inside this `api.h`. This file is small anyway.

spacewander: @agentzh 
The `ngx_http_lua_ffi_str_t` is required inside of the lua-nginx-module too. Put it into `api.h` will require the other headers to include the public API header. I think it could be more flexible by adding a middle layer.

agentzh: @spacewander Hmm, I disagree here. Internally we should use `ngx_str_t` directly.

spacewander: @agentzh 
`ngx_http_lua_ffi_str_t` is used in `ngx_http_lua_util.h`:
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_util.h#L25

If we put `ngx_http_lua_ffi_str_t` into `api.h`, then we should include the public API header into the internal utility header, which is strange.

agentzh: @spacewander I don't want that `api_common.h` file. It's horrible and confusing. You can include the whole `api.h` file if necessary. I don't mind.

agentzh:
@spacewander Committed 809192c instead. Thanks!

PR #1274, opened 2018-03-09
feature: enabled sending boolean and nil in cosocket.

spacewander:

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -2462,6 +2462,20 @@ ngx_http_lua_socket_tcp_send(lua_State *L)
             len = ngx_http_lua_calc_strlen_in_table(L, 2, 2, 1 /* strict */);
             break;
 
+        case LUA_TNIL:
+            len = sizeof("nil") - 1;
+            break;
+
+        case LUA_TBOOLEAN:
+            if (lua_toboolean(L, 2)) {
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_socket_tcp.c
```diff
@@ -2462,6 +2462,20 @@ ngx_http_lua_socket_tcp_send(lua_State *L)
...
+
+        case LUA_TBOOLEAN:
+            if (lua_toboolean(L, 2)) {
+                len = sizeof("true") - 1;
+
+            } else {
+                len = sizeof("false") - 1;
+            }
+
+            break;
```
agentzh: Sending booleans and nils are more likely due to bugs in the application code? I'm not sure if this is the right thing to do.

It's for the sake of duck type. API like `ngx.log` and `ngx.say` handle boolean and nil by converting them to string. If we finally migrate cosocket to FFI implementation, we will use `tostring` to non-string type argument too. BTW, it also conforms the description in the error message: "string, number, boolean, nil, or array table expected".

agentzh: @spacewander OK then.

agentzh:
We're migrating to meta-lua-nginx-module and we'll need existing PRs like this to be ported to that repo instead.

agentzh:
@spacewander Merged. Thanks!

PR #1282, opened 2018-03-21
balancer: Add ngx_http_lua_ffi_balancer_set_ssl_ctx

james-callahan:
This function allows setting an upstream's SSL_CTX* to a user-provided object

It comes with a sister pull-request to lua-resty-core


src/ngx_http_lua_balancer.c
```diff
@@ -760,4 +760,88 @@ ngx_http_lua_ffi_balancer_get_last_failure(ngx_http_request_t *r,
...
 
+
+#if NGX_HTTP_SSL
+
+int
+ngx_http_lua_ffi_balancer_set_ssl_ctx(ngx_http_request_t *r,
+    SSL_CTX* ssl_ctx, char **err)
+{
+    ngx_http_lua_ctx_t  *ctx;
+    ngx_http_upstream_t *u;
```
p0pr0ck5: Pedantic style: two spaces between `ngx_http_upstream_t` and `*u`

daurnimator: I thought the style was to vertically align `ctx` and `u`?

james-callahan: I thought the style was to vertically align `ctx` and `u`?

p0pr0ck5: Based on http://nginx.org/en/docs/dev/development_guide.html#code_style_variables (and it looks like this repo follows this guide), there are two spaces separating the columns (types and names), with the right column left-aligned based on deference operators.

PR #1284, opened 2018-03-23
misc: suppressed known false positive leak from cache manager.

dndx:

### DIFF
--- a/valgrind.suppress
+++ b/valgrind.suppress
@@ -173,6 +173,20 @@
 }
 {
    <insert_a_suppression_name_here>
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:ngx_alloc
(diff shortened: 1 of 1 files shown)


valgrind.suppress
```diff
@@ -171,6 +171,20 @@
    fun:ngx_create_pool
    fun:main
 }
+{
+   <insert_a_suppression_name_here>
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:ngx_alloc
+   fun:ngx_event_process_init
```
spacewander: IMHO, we already have this suppression: https://github.com/openresty/lua-nginx-module/blob/master/valgrind.suppress#L94

dndx: Didn't notice that. Thanks for pointing out. Closing now.

PR #1286, opened 2018-03-26
feature: Add api to fetch raw nginx ssl pointer

james-callahan:
This allows introspection of the `SSL*` object from lua code.

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -1319,6 +1319,17 @@ failed:
 }
 
 
+ngx_ssl_conn_t *
+ngx_http_lua_ffi_get_ssl_pointer(ngx_http_request_t *r)
+{
+    if (r->connection == NULL || r->connection->ssl == NULL) {
+        return NULL;
+    }
+
+    return r->connection->ssl->connection;
+}
+
+
 #endif  /* NGX_LUA_NO_FFI_API */
 
 
--- a/t/140-ssl-c-api.t
+++ b/t/140-ssl-c-api.t
@@ -57,6 +57,8 @@ ffi.cdef[[
     int ngx_http_lua_ffi_set_priv_key(void *r,
         void *cdata, char **err);
 
+    void *ngx_http_lua_ffi_get_ssl_pointer(void *r);
+
     void ngx_http_lua_ffi_free_cert(void *cdata);
 
     void ngx_http_lua_ffi_free_priv_key(void *cdata);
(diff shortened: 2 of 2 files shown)


james-callahan:
Test seems to be failing for unrelated reasons.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1319,6 +1319,21 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
 }
 
 
+int
+ngx_http_lua_ffi_get_ssl_pointer(ngx_http_request_t *r,
```
agentzh: We are moving to the meta-lua-nginx-module:

https://github.com/openresty/meta-lua-nginx-module

So this change deserves supporting other subsystems as well (like the `stream` subsystem) if we indeed want to have this.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1319,6 +1319,21 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+    ngx_ssl_conn_t **ssl_conn, char **err)
+{
+    if (r->connection == NULL || r->connection->ssl == NULL) {
+        *err = "bad request";
+        return NGX_ERROR;
```
agentzh: It's better to just return the SSL pointer directly and return NULL in case of errors.

src/ngx_http_lua_ssl_certby.c
james-callahan: Won't this get added when the stream subsystem gains `ssl_certificate_by_lua_block` support?

src/ngx_http_lua_ssl_certby.c
james-callahan: Will fix.

james-callahan:
Test added.

james-callahan:
Is there anything blocking this PR?

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1319,6 +1319,17 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
 }
 
 
+ngx_ssl_conn_t *
+ngx_http_lua_ffi_get_ssl_pointer(ngx_http_request_t *r)
```
agentzh, line 1323: `ngx_http_lua_ffi_get_req_ssl_ptr` is a better name.

agentzh (changes requested):

zhuizhuhaomeng:
merged

PR #1287, opened 2018-03-26
optimize: implement rwlock around shared dictionary access

### LINKED ISSUE #656: Is there any plan to implement a read-wirte lock for lua_shared_dict
I was working on a dynamic upstream feature for Marathon-docker framework  with balancer_by_lua_block,  which upstream server list stored in lua_shared_dict.  At this case, server list reading must be blocked when reloading the server list, and it's not a good idea to get a lock at each http request. I think, that could be properly sovled with a read-write lock.
So, do you have any plan to implement a read-wirte lock for lua_shared_dict

p0pr0ck5:
When defined at compile time, use the native read-write lock implementation available in Nginx to guard concurrent access to shared dictionary memory pools.

Fixes #656.


.travis.yml
```diff
@@ -97,7 +97,7 @@ before_script:
 script:
   - sudo iptables -I OUTPUT 1 -p udp --dport 10086 -j REJECT
   - cd luajit2/
-  - make -j$JOBS CCDEBUG=-g Q= PREFIX=$LUAJIT_PREFIX CC=$CC XCFLAGS='-DLUA_USE_APICHECK -DLUA_USE_ASSERT -msse4.2' > build.log 2>&1 || (cat build.log && exit 1)
+  - make -j$JOBS CCDEBUG=-g Q= PREFIX=$LUAJIT_PREFIX CC=$CC XCFLAGS='-DNGX_HTTP_LUA_SHDICT_HAVE_RWLOCK -DLUA_USE_APICHECK -DLUA_USE_ASSERT -msse4.2' > build.log 2>&1 || (cat build.log && exit 1)
```
p0pr0ck5: i suspect this may not be desirable in a final commit; this just demonstrates the validity of the locking mechanism in unit tests.

src/ngx_http_lua_shdict.c
```diff
@@ -56,6 +56,31 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
 #define NGX_HTTP_LUA_SHDICT_RIGHT       0x0002
 
 
+#ifdef NGX_HTTP_LUA_SHDICT_HAVE_RWLOCK
+#define NGX_HTTP_LUA_SHDICT_RLOCK 0
+#define NGX_HTTP_LUA_SHDICT_WLOCK 1
+
+#define ngx_http_lua_shdict_lock(ctx, t)        \
+    if (t == NGX_HTTP_LUA_SHDICT_RLOCK) {       \
```
agentzh: Please make sure all macro parameters always get enclosed by parentheses to avoid potential operator precedence issues in macro expansion.

Also, I wonder can we make this a static C inline function instead?

src/ngx_http_lua_shdict.c
```diff
@@ -56,6 +56,31 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
...
+        ngx_rwlock_rlock(&ctx->rwlock);         \
+    } else {                                    \
```
agentzh: Style: needs a blank line before the `} else {` line.

src/ngx_http_lua_shdict.c
```diff
@@ -56,6 +56,31 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
 #define NGX_HTTP_LUA_SHDICT_RIGHT       0x0002
 
 
+#ifdef NGX_HTTP_LUA_SHDICT_HAVE_RWLOCK
+#define NGX_HTTP_LUA_SHDICT_RLOCK 0
+#define NGX_HTTP_LUA_SHDICT_WLOCK 1
+
+#define ngx_http_lua_shdict_lock(ctx, t)        \
```
agentzh: Style: `\` should always be on the 78th column.

src/ngx_http_lua_shdict.c
p0pr0ck5: Yeah, refactored to make a `static ngx_inline void`.

agentzh:
I'm kinda worried about the lack of semaphore fallbacks. This may lead to undesired high CPU usage when lock contention is high (like lots of writes on huge trees).

agentzh:
Can we make this configurable at runtime? Like via an nginx config directive? The user must choose at build time is very sad especially for pre-built binary packages.

src/ngx_http_lua_shdict.c
```diff
@@ -1515,14 +1558,14 @@ ngx_http_lua_shared_dict_get(ngx_shm_zone_t *zone, u_char *key_data,
 
     ctx = zone->data;
 
-    ngx_shmtx_lock(&ctx->shpool->mutex);
+    ngx_http_lua_shdict_lock(ctx, NGX_HTTP_LUA_SHDICT_RLOCK);
```
agentzh: This looks wrong to me. You know that the `ngx_http_lua_shdict_unlock()` call is not really read only? It also always updates the LRU queue, which is not safe when multiple readers trying to change it at the same time. We need finer grained locking here.

p0pr0ck5:
> I'm kinda worried about the lack of semaphore fallbacks. This may lead to undesired high CPU usage when lock contention is high (like lots of writes on huge trees).

> Can we make this configurable at runtime? Like via an nginx config directive? The user must choose at build time is very sad especially for pre-built binary packages.

Yep, I intended to couple this with a configure-time option (hence leaving the cpp def logic in place, not simply removing the shmtx calls entirely). Shall I make a sister PR to `openresty/openresty` now? I wanted to hold off until this had some eyeballs.

> This looks wrong to me. You know that the ngx_http_lua_shdict_unlock() call is not really read only? It also always updates the LRU queue, which is not safe when multiple readers trying to change it at the same time. We need finer grained locking here.

Yep, I realized this about an hour after submitting this PR. Just now getting to fix it :p I have an ammended commit in place that addresses this.

src/ngx_http_lua_shdict.c
p0pr0ck5: I think this comment is correct in a number of places, but not in `ngx_http_lua_shdict_get_keys` in particular because there is no LRU queue adjustment that I can see.

agentzh: @p0pr0ck5 True. Wrong place.

agentzh:
@p0pr0ck5 I think you can just make the locking and unlocking calls function pointers? So that we can switch them at runtime? Extra C function calls may have a little bit of performance penalty though. We could measure it anyway :)

p0pr0ck5:
I felt like a runtime choice was unnecessary, hence defining them as macros (since there is no lock downgrade for shmtx I reverted to macros instead of static inline because the shmtx downgrade would have been a useless empty function call). Let me know your thoughts, this seems like the most performant approach.

agentzh:
@p0pr0ck5 Runtime selection is necessary for pre-built binary packages of OpenResty otherwise it is not doable.

p0pr0ck5:
@agentzh so you want each `ngx.shared` to have its own lock implementation?

agentzh:
@p0pr0ck5 BTW, inline functions should be the same as macros performance wise but much cleaner and safer. But here it seems like we really need function pointers or extra branch to do runtime selection (another more space demanding approach is to generate two different code copies of shdict via templates as in meta-lua-nginx-module).

p0pr0ck5:
@agentzh so to clarify you want to see an additional API for each shared dictionary to do mutex vs rwlock lock implementation? Or by 'runtime config' do you just mean an Nginx config directive that controls how all lua_shared_dictionary instances implement the lock?

agentzh:
@p0pr0ck5 It's sufficient if we can control that at config time. We don't need to do this at Lua runtime.

p0pr0ck5:
@agentzh done. rwlocks can now be implemented per dictionary by adding the token `rwlock` to the directive, e.g.

`lua_share_dict foo 128m rwlock;`

If `rwlock` is defined implemented then the existing behavior is used (thus existing `lua_shared_dictionary foo 128m` style definitions still work, of course).

thibaultcha:
@p0pr0ck5 Have you had a chance to benchmark some use-cases? I am quite curious to hear about it, or maybe try some on my own if you haven't had a chance yet.

p0pr0ck5:
@thibaultcha I did some preliminary testing but don't have numbers on me at the moment.

The biggest change came from concurrent usages of `get_keys`, since that operation no longer needs to hold a mutex. Primitive operations (set, get, etc) were not impacted in high concurrency scenarios until many hundreds of worker processes were introduced to artificially force a lot of lock contention.

thibaultcha:
@p0pr0ck5 Gotcha.

> The biggest change came from concurrent usages of get_keys, since that operation no longer needs to hold a mutex.

Yes, I am particularly interested in this win :) Nice work!

p0pr0ck5:
@agentzh ping :)

stone-wind:
@p0pr0ck5 

In the rwlock implementation of nginx, the priority of read lock is higher than that of write. Under a large number of read concurrency, the write delay will be large. The following article describes similar phenomena:
https://fy.blackhats.net.au/blog/html/2018/10/19/rust_rwlock_and_mutex_performance_oddities.html

maybe, we should prevent writer starvation and optimize the rwlock implementation . The following article shows that rwlock is better than mutex at macOS:
https://www.slideshare.net/mitsunorikomatsu/performance-comparison-of-mutex-rwlock-and-atomic-types-in-rust

p0pr0ck5:
Yes that’s why we make the behavior change optional. :) it can suit particular use cases where the useage pattern warrants. 

The scope here is not to improve the design of the underlying Nginx rwlock implementation. That would be for the Nginx developers to handle. Here we’re concerned with how openresty wraps these primatives and aligns usage accordingly.

src/ngx_http_lua_shdict.h
```diff
@@ -11,6 +11,23 @@
 #include "ngx_http_lua_common.h"
 
 
+struct ngx_http_lua_shdict_ctx_s;
+typedef struct ngx_http_lua_shdict_ctx_s ngx_http_lua_shdict_ctx_t;
+typedef void (* memory_rlock)(ngx_http_lua_shdict_ctx_t *);
+typedef void (* memory_wlock)(ngx_http_lua_shdict_ctx_t *);
+typedef void (* memory_unlock)(ngx_http_lua_shdict_ctx_t *);
+typedef void (* memory_lock_downgrade)(ngx_http_lua_shdict_ctx_t *);
```
agentzh: Also, it's bad to pollute the global namespace with unqualified names like `memory_rlock`. Note that all the nginx global symbols (and also file-scoped static ones) use prefixes like `ngx_http_xxx` to avoid name space pollution.

src/ngx_http_lua_shdict.c
```diff
@@ -55,6 +55,49 @@ static ngx_inline ngx_shm_zone_t *ngx_http_lua_shdict_get_zone(lua_State *L,
...
+    ngx_shmtx_lock(&ctx->shpool->mutex);
+}
+
+ngx_inline void
+ngx_http_lua_shdict_shmtxunlock(ngx_http_lua_shdict_ctx_t *ctx)
+{
+    ngx_shmtx_unlock(&ctx->shpool->mutex);
+}
+
+ngx_inline void
```
agentzh: The `ngx_inline` flag is useless. These functions are only called as function pointers where inlining is impossible anyway. Am I missing anything here?

agentzh:
Need some actual benchmark numbers to make sure we have no measurable performance regression for the original use cases.

Also, I wonder if we can use debugging logs to cover this feature in some test cases. Right now this change is not covered by the test suite and it may not be doing what it is supposed to do.

agentzh:
Also, need some concrete benchmark numbers for the cases rwlock is a win.

src/ngx_http_lua_shdict.c
```diff
@@ -503,7 +546,7 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
                    "fetching key \"%V\" in shared dict \"%V\"", &key, &name);
 #endif /* NGX_DEBUG */
 
-    ngx_shmtx_lock(&ctx->shpool->mutex);
+    ctx->wlock(ctx);
```
agentzh: This defeats the benefit of the rwlock when looking up the key over huge rbtrees. I think we need finer granularity of locking. For the rbtree lookup, it should be reader lock while the LRU queue updating, it should be the writer lock.

src/ngx_http_lua_shdict.c
```diff
@@ -513,10 +556,12 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
 
     rc = ngx_http_lua_shdict_lookup(zone, hash, key.data, key.len, &sd);
 
+    ctx->lock_downgrade(ctx);
```
agentzh: Why downgrade the lock when we are still holding a write lock here? It looks wrong to me.

src/ngx_http_lua_shdict.c
agentzh: But for changing the rbtree (like updating nodes, deleting nodes, inserting nodes), then we need a write lock.

agentzh:
BTW, using `get_keys()` with large count argument on big shdict is a wrong use case anyway. No sane key-value stores would support such extreme operations very well anyway. So I won't consider it a valid use case for this PR.

This PR is only meaningful when there are many reads from a huge shdict (deep rbtrees) and there are many CPU cores and thus many worker processes actually competing for the locks. Otherwise it's meaningless.

agentzh:
I'm also worried that according to the current rwlock implementation, the writers may never have a chance to obtain the lock (at least before too long) for continuous flows of readers.

PR #1288, opened 2018-03-26
bugfix: sanity checking in ngx.timer.at delay param

### LINKED ISSUE #1278: If the timer value in ngx.timer.at() is too high, callback triggers immediately
> * The exact version of the related software, including but not limited to the OpenResty version (if any), the NGINX core version, the `ngx_lua` module version,
and your operating system version.

nginx version: openresty/1.11.2.5
built by gcc 6.2.0 (GCC) 
ngx_lua-0.10.10
Linux 4.13.2

> * A minimal and standalone test case that others can easily run on their side and
reproduce the issue you are seeing.

```
http {
    server {
        listen 5309;
        location = / {
            content_by_lua_block {
                local function fn()
                    ngx.log(ngx.ERR, "Hello from callback")
                end
                ngx.timer.at(1e16, fn)
                return 200;
            }
        }
    }
}
```

Once I hit the endpoint, `"Hello from callback"` shows up immediately in the logs. If I lower the timer to `1e15`, it doesn't.

p0pr0ck5:

Fixes #1278.

### DIFF
--- a/src/ngx_http_lua_timer.c
+++ b/src/ngx_http_lua_timer.c
@@ -16,6 +16,9 @@
 #include "ngx_http_lua_probe.h"
 
 
+#define MAX_DELAY 2147483648 /* 2 ^ 31 */
+
+
 typedef struct {
     void        **main_conf;
     void        **srv_conf;
@@ -134,6 +137,7 @@ ngx_http_lua_ngx_timer_helper(lua_State *L, int every)
     lua_State               *vm;  /* the main thread */
     lua_State               *co;
     ngx_msec_t               delay;
+    lua_Integer              arg;
     ngx_event_t             *ev = NULL;
     ngx_http_request_t      *r;
     ngx_connection_t        *saved_c = NULL;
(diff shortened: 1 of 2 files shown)


agentzh:
Just a side note: I wonder if we should omit all the API checks by macros like LuaJIT's `LUA_USE_APICHECK` macro. All such checks come with a cost and the accumulated overhead can be big if these API functions are called frequently.

p0pr0ck5:
If there were detailed benchmarks showing the impact of such sanity checks, I would totally be onboard. I would hope that a few small branches here wouldn't case a noticeable impact in the grand scheme of things.

p0pr0ck5:
@agentzh any thoughts here? I did some benchmarking with a macro like `LUA_USE_APICHECK` against this function and a few others (like `ngx_http_lua_shdict_get_helper`, `ngx_http_lua_ngx_sleep`, etc) and found no appreciable difference. Let me know if you want me to share my data or how you want to proceed on this PR.

dndx:
This particular change is probably fine to be enabled by default as the cost of comparison is practically negligible compared to the cost of creating a timer.

p0pr0ck5:
@thibaultcha @agentzh ping?

src/ngx_http_lua_timer.c
```diff
@@ -16,6 +16,9 @@
 #include "ngx_http_lua_probe.h"
 
 
+#define MAX_DELAY 2147483648 /* 2 ^ 31 */
```
agentzh: I'm not sure if this threshold makes sense since `ngx_msec_t` is defined as `ngx_uint_t`, which is a 64-bit integer on 64-bit systems, no?

agentzh:
The trend is to check integer overflow for all API functions? I'm not sure if we want to do this everywhere. It's too much a burden IMHO.

PR #1290, opened 2018-03-27
tests: remove google.com relevant ssl certificate test cases.

spacewander:
This pr should fix the recent random failure of lua-nginx-module test suite.

We could not control Google's certificate, and Google don't promise
their root certificate won't change, see https://pki.google.com/faq.html.

> Q: What roots should we trust for connecting to Google?
> A: Google may decide to have its intermediate signed by another root at any
point in time, ...

Because in TEST 8 we already tested passing verify case with
openresty.org, here I remove TEST 11 and change TEST 12 to test with
openresty.org.

### DIFF
--- a/t/129-ssl-socket.t
+++ b/t/129-ssl-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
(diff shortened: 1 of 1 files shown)


agentzh:
@spacewander Merged. Thanks!

PR #1295, opened 2018-04-04
style: replace ngx_http_parse_time with ngx_parse_http_time

p0pr0ck5:
The former is a compatibility definition for the latter introduced in Nginx 8b6fa4842133. Both uses are in place in the codebase; for consistency let's just use the function definition to keep things aligned.


agentzh:
@p0pr0ck5 No, this change breaks compatibility with older nginx versions like 1.6.0 while adding no real benefit.

p0pr0ck5:
Shoot, I misread the output of grep while looking for ngx_parse_http_time. Silly me!

p0pr0ck5:
btw @agentzh if you want to ensure backward compatibility that far, that should probably be part of the acceptance CI/test suite process.

PR #1296, opened 2018-04-10
feature: added an FFI API for setting the SA_RESTART flag on signals

thibaultcha:
This is one of 4 PRs to properly handle `EINTR` errors in resty-cli (and/or OpenResty if desired by the user).

- https://github.com/openresty/luajit2/pull/18 Revert the "retry on EINTR" patch
- (this PR) Implement an FFI API to enable `SA_RESTART` on signal handlers
- openresty/lua-resty-core#<TODO> Implement the Lua `sig_restart()` API
- openresty/resty-cli#<TODO> Enable the `SA_RESTART` flag on interrupting signals

If all PRs are merged:
- we get rid of the "retry on EINTR" LuaJIT patch which was refused upstream
- resty-cli users can safely use APIs like `io.popen()` https://github.com/openresty/resty-cli/issues/35
- OpenResty users can toggle this behaviour on/off as desired (off by default)


src/ngx_http_lua_worker.c
```diff
@@ -190,4 +190,33 @@ ngx_http_lua_ffi_process_signal_graceful_exit(void)
...
 }
+
+
+int
+ngx_http_lua_ffi_enable_sa_restart(int signum, int enabled)
+{
+    struct sigaction        act;
+
+    if (ngx_process != NGX_PROCESS_WORKER &&
+        ngx_process != NGX_PROCESS_SINGLE) {
```
agentzh: Style: `{` should be on its own line when the `if` condition is splitted.

src/ngx_http_lua_worker.c
```diff
@@ -190,4 +190,33 @@ ngx_http_lua_ffi_process_signal_graceful_exit(void)
 {
     ngx_quit = 1;
 }
+
+
+int
+ngx_http_lua_ffi_enable_sa_restart(int signum, int enabled)
+{
+    struct sigaction        act;
```
agentzh: I'm afraid this breaks compatibility with systems like Windows. Better do a feature test to make sure we have these signal stuffs?

thibaultcha: Both comments addressed. Is it worth starting to invest time in tests and documentation for this?

src/ngx_http_lua_worker.c
```diff
@@ -190,4 +190,46 @@ ngx_http_lua_ffi_process_signal_graceful_exit(void)
...
+
+int
+ngx_http_lua_ffi_enable_sa_restart(int signum, int enabled, u_char *err,
+    size_t *errlen)
+{
+#if !(NGX_WIN32)
+    struct sigaction        act;
+
+    if (ngx_process != NGX_PROCESS_WORKER
+        && ngx_process != NGX_PROCESS_SINGLE)
```
agentzh: We need to take into account the privileged agent helper process here.

agentzh:
@thibaultcha I think it's better to simply enable `SA_RESTART` in our lua module's init worker hook? It's a burden for the user to always remember calling a Lua API in their own `init_worker_by_lua*` handler.

thibaultcha:
@agentzh Does this list sound good for the set of signals the `SA_RESTART` flag is to be enabled by default: https://github.com/openresty/resty-cli/pull/41/files#diff-f40cddd047a559bb89d08b41dfdfd86dR625 ?

thibaultcha:
I must admit that for the sake of backwards-compatibility (in the event of an unforeseen issue), I was under the impression that we would not enable this flag on any signal by default - especially considering it is enabled by default in resty-cli (as per openresty/resty-cli#41), where it matters the most.

But I am fine enabling it by default in this module's init context, no problem (if the suggested list of signals is good to go).

agentzh:
@thibaultcha That list looks good to me. This is also important for non-resty-cli contexts. I don't like enabling by hand, which is too easy to miss for average users. And it does not look like there is any unwanted side effects so there's no backward compatibility that we need to preserve here?

thibaultcha:
@agentzh @dndx Ping here :) 

Also maybe https://github.com/openresty/lua-nginx-module/pull/1239 as well, which I still consider quite important for true Lua compatibility.

thibaultcha:
Patch updated:
- introduced a `lua_sa_restart` directive (`on` by default)
- removed `ngx_http_lua_set_sa_restart_helper(int signo, int enabled, u_char *err, size_t *errlen)`; now, the `SA_RESTART` flag is directly set by `ngx_http_lua_set_sa_restart` (which cannot remove the flag)
- removed `ngx_http_lua_ffi_set_sa_restart` (which invalidates openresty/lua-resty-core#183)

I did like the granularity offered by the FFI API (being able to set/unset this flag as desired on a per-signal basis), but I removed it to show what a minimal patch would look like. I can add the FFI back any time if desired.

src/ngx_http_lua_module.c
```diff
@@ -104,6 +104,13 @@ static ngx_command_t ngx_http_lua_cmds[] = {
       0,
       NULL },
 
+     { ngx_string("lua_sa_restart"),
+      NGX_HTTP_MAIN_CONF|NGX_CONF_FLAG,
+      ngx_conf_set_flag_slot,
+      NGX_HTTP_MAIN_CONF_OFFSET,
+      offsetof(ngx_http_lua_main_conf_t, set_sa_restart),
+      NULL },
```
agentzh: Inconsistent indentation above.

src/ngx_http_lua_util.c
```diff
@@ -70,6 +70,25 @@
 #endif
 
 
+#if !(NGX_WIN32)
```
agentzh: I suggest we do feature test instead of excluding just win32 here for better portability.

thibaultcha:
@agentzh Updated with a feature test. Mind having another look? If ready, I will port this PR to the meta-lua module as well.

thibaultcha:
Pushed some documentation.

src/ngx_http_lua_util.c
```diff
@@ -4195,4 +4214,32 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
...
+    int                    *signo;
+    int                     sigs[] = NGX_HTTP_LUA_SA_RESTART_SIGS;
+    struct sigaction        act;
+
+    for (signo = sigs; *signo != 0; signo++) {
+        ngx_log_debug1(NGX_LOG_DEBUG_HTTP, log, 0,
+                       "setting SA_RESTART for signal %d", *signo);
+
+        if (sigaction(*signo, NULL, &act) != 0) {
+            ngx_log_error(NGX_LOG_WARN, log, 0, "failed to get sigaction for "
```
agentzh: Should use the `errno` arg directly.

thibaultcha: @agentzh Updated

PR #1306, opened 2018-04-20
Truncated retval

agentzh:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4173,7 +4173,7 @@ For reading *request* headers, use the [ngx.req.get_headers](#ngxreqget_headers)
 
 ngx.resp.get_headers
 --------------------
-**syntax:** *headers = ngx.resp.get_headers(max_headers?, raw?)*
+**syntax:** *headers, err = ngx.resp.get_headers(max_headers?, raw?)*
 
 **context:** *set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, balancer_by_lua&#42;*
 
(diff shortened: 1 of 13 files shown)


agentzh:
The sister PR is here: https://github.com/openresty/lua-resty-core/pull/187

src/ngx_http_lua_headers.c
```diff
@@ -480,12 +483,25 @@ ngx_http_lua_ngx_req_get_headers(lua_State *L)
                        "lua request header: \"%V: %V\"",
                        &header[i].key, &header[i].value);
 
-        if (--count == 0) {
+        if (--count <= 0) {
```
dndx: Will this be clearer:

```C
for (...) {
    if (--count <= 0) {
        break;
    }
}

if (truncated) {
    goto truncated;
}

return 1;

truncated:

lua_pushliteral(L, "truncated");
return 2;
```

Similar code can be found in `ngx_http_lua_ngx_resp_get_headers` as well.

p0pr0ck5:
I do not think the phrase "one can choose to ignore or reject the current response here" as found in the example docs is appropriate. It implies these are these only two options available. Simply a note that the value could be present should suffice.

README.markdown
```diff
@@ -4535,20 +4547,23 @@ Here the `args` table will always look like
...
+     -- one can choose to ignore or reject the current request here
+ end
 ```
 
 This argument can be set to zero to remove the limit and to process all request arguments received:
 
 ```lua
 
- local args = ngx.req.get_uri_args(0)
+ local args, err = ngx.req.get_uri_args(0)
```
p0pr0ck5: since `err` would never be set here (at least in the current context), is it appropriate to leave this example as-is?

README.markdown
```diff
@@ -4637,20 +4657,23 @@ That is, they will take Lua boolean values `true`. However, they are different f
...
+     -- one can choose to ignore or reject the current request here
+ end
 ```
 
 This argument can be set to zero to remove the limit and to process all request arguments received:
 
 ```lua
 
- local args = ngx.req.get_post_args(0)
+ local args, err = ngx.req.get_post_args(0)
```
p0pr0ck5: since err would never be set here (at least in the current context), is it appropriate to leave this example as-is?

README.markdown
agentzh: @p0pr0ck5 That's for forward compatibility (we may add more error strings in the future).

p0pr0ck5: Yes but the context here is specifically about a unlimited number of args returned.

agentzh: @p0pr0ck5 It's better to always be prepared for an error string to be returned in our docs. It's  much better than assuming only one value can be returned.

agentzh:
Merged. Thanks all for the review.

PR #1310, opened 2018-04-24
 feature: implemented bsd style cosocket receive method. 

spacewander:
This feature is atop #861, with some minor issues fixed.


README.markdown
```diff
@@ -7206,6 +7206,7 @@ If a number-like argument is specified (including strings that look like numbers
 If a non-number-like string argument is specified, then it is interpreted as a "pattern". The following patterns are supported:
 
 * `'*a'`: reads from the socket until the connection is closed. No end-of-line translation is performed;
+* `'*b'`: the BSD style receive, return anything once socket receives;
```
agentzh: Better be consistent: "return" => "returns".

Also, this is better:

"returns anything the socket has already received (won't return if it has not yet received any data)."

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2027,6 +2032,38 @@ ngx_http_lua_socket_read_all(void *data, ssize_t bytes)
...
+    ngx_buf_t                   *b;
+
+#if (NGX_DEBUG)
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, u->request->connection->log, 0,
+                   "lua tcp socket read bsd");
+#endif
+
+    if (bytes == 0) {
+        u->ft_type |= NGX_HTTP_LUA_SOCKET_FT_CLOSED;
+        return NGX_ERROR;
```
agentzh: I'm not sure about this. Maybe it's better to return the empty Lua string when the reading part of the connection has been closed by the remote? This should not be considered a fatal error. And we need a test case to cover this code path.

If we do so, we could not detect if the remote is closed as the same way with `sock:receive('*l')`.

This case is already covered by this line: https://github.com/spacewander/lua-nginx-module/blob/77c34e1f3aec37acb84bf67661ae9cc22b21c2eb/t/058-tcp-socket.t#L3849

t/058-tcp-socket.t
```diff
@@ -3808,3 +3808,131 @@ received:
...
+        content_by_lua_block {
+            local sock = ngx.socket.tcp()
+            sock:settimeout(500)
+            assert(sock:connect("127.0.0.1", ngx.var.port))
+            local req = {
+                'GET /foo HTTP/1.0\r\n',
+                'Host: localhost\r\n',
+                'Connection: close\r\n\r\n',
+            }
+            sock:send(req)
```
agentzh: Better to check the return values of `send()` and handle any potential errors? Please fix other similar places (if any).

t/058-tcp-socket.t
```diff
@@ -3808,3 +3808,131 @@ received:
 received: truefalsenil
 --- no_error_log
 [error]
+
+
+
+=== TEST 64: *b pattern for receive
```
agentzh: I suggest you also duplicate this test case in a separate `.t` test file which *requires*` the use of mockeagain reading test mode for the extreme case of receiving one byte at a time:

```
t/058-tcp-socket.t .. 115/416
#   Failed test 'TEST 64: *b pattern for receive - response_body - response is expected (repeated req 0, req 0)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1555.
# @@ -1,3 +1,14 @@
#  1
# -22
# +2
# -hello world
# +2
# +h
# +e
# +l
# +l
# +o
# +
# +w
# +o
# +r
# +l
# +d

#   Failed test 'TEST 64: *b pattern for receive - response_body - response is expected (repeated req 1, req 0)'
#   at /home/agentzh/git/lua-nginx-module/../test-nginx/lib/Test/Nginx/Socket.pm line 1555.
# @@ -1,3 +1,14 @@
#  1
# -22
# +2
# -hello world
# +2
# +h
# +e
# +l
# +l
# +o
# +
# +w
# +o
# +r
# +l
# +d
t/058-tcp-socket.t .. 379/416 # Looks like you failed 2 tests of 416.
```

agentzh: Oh, and maybe mockeagain *writing* mode too :)

spacewander: @agentzh 
Additional test added.

spacewander:
Will change the method name and submit in another pr.

PR #1312, opened 2018-04-28
add balancer_ssl_session_fetchby and balancer_ssl_session_storeby

dongbeiouba:

Nginx support caching and reusing the upstream ssl session, but balancer_by_lua don't support reusing the ssl session.

So i add the hook balancer_ssl_session_fetchby and balancer_ssl_session_storeby; and then we can cache the ssl session with lua code. 

Cache and reuse the ssl session can reduce ssl full handshake with the proxied servers; especially for those nginx which not config keepalive.


gwittel:
Are there any plans to pick up this work?  Was debugging SSL induced performance issues when I verified this behavior (and later found this ticket).

doujiang24:
@spacewander @rainingmaster please help to take a look at this PR, thanks :)

nr17:
@spacewander @rainingmaster Any help in integrating this PR would be much appreciated.

PR #1318, opened 2018-05-12
optimize: don't calculate hash when clearing the header.

spacewander:
The `hash` will be set to `0` later when clearing the header. By saving the extra hash operation we could save the cost of `ngx.header.HEADER = nil` by 10%.

### DIFF
--- a/src/ngx_http_lua_headers_out.c
+++ b/src/ngx_http_lua_headers_out.c
@@ -485,7 +485,10 @@ ngx_http_lua_set_output_header(ngx_http_request_t *r, ngx_str_t key,
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    hv.hash = ngx_hash_key_lc(key.data, key.len);
+    if (value.len > 0) {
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_headers_out.c
```diff
@@ -485,7 +485,10 @@ ngx_http_lua_set_output_header(ngx_http_request_t *r, ngx_str_t key,
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    hv.hash = ngx_hash_key_lc(key.data, key.len);
+    if (value.len > 0) {
+        hv.hash = ngx_hash_key_lc(key.data, key.len);
+    }
```
agentzh: I think it's better to clear the `hv.hash` field when `value.len == 0` here?

spacewander:
@agentzh 
Updated.

zhuizhuhaomeng (approved):

zhuizhuhaomeng:
rebase & merged

PR #1323, opened 2018-05-18
feature: implemented receiveany cosocket method.

spacewander:
The `receiveany` method is named `receive('*b')` in previous pr. The name is renamed for the sake of comprehension.Additional `max` argument is added to avoid getting too much data in a single call.

TODO: add this method to `ngx.req.socket`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -7246,7 +7246,7 @@ This method is a synchronous operation just like the [send](#tcpsocksend) method
 In case of success, it returns the data received; in case of error, it returns `nil` with a string describing the error.
 
 If the received data is more than this size, this method will return with exactly this size of data.
-The remain received data could be returned in the next `receiveany` call.
+The remaining data in the underlying receive buffer could be returned in the next reading operation.
 
 Timeout for the reading operation is controlled by the [lua_socket_read_timeout](#lua_socket_read_timeout) config directive and the [settimeouts](#tcpsocksettimeouts) method. And the latter takes priority. For example:
 
@@ -7261,7 +7261,7 @@ Timeout for the reading operation is controlled by the [lua_socket_read_timeout]
  ngx.say("successfully read: ", data)
 ```
 
-This method doesn't automatically close the current connection when the read timeout error happens. For other connection errors, this method always automatically closes the connection.
+This method doesn't automatically close the current connection when the read timeout error occurs. For other connection errors, this method always automatically closes the connection.
 
 This feature was first introduced in the `dev` release.
 
--- a/README.markdown
+++ b/README.markdown
@@ -3221,6 +3221,7 @@ Nginx API for Lua
 * [tcpsock:sslhandshake](#tcpsocksslhandshake)
 * [tcpsock:send](#tcpsocksend)
 * [tcpsock:receive](#tcpsockreceive)
+* [tcpsock:receiveany](#tcpsockreceiveany)
 * [tcpsock:receiveuntil](#tcpsockreceiveuntil)
 * [tcpsock:close](#tcpsockclose)
 * [tcpsock:settimeout](#tcpsocksettimeout)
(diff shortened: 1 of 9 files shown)


README.markdown
```diff
@@ -7231,6 +7233,40 @@ This feature was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+tcpsock:receiveany
+------------------
+**syntax:** *data, err = tcpsock:receiveany(max?)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, ngx.timer.&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;*
+
+Return any data received by the connected socket, at most `max` bytes if the `max` argument is specified.
```
agentzh: Should use "Returns" here instead for the sake of consistency.

README.markdown
```diff
@@ -7231,6 +7233,40 @@ This feature was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+tcpsock:receiveany
+------------------
+**syntax:** *data, err = tcpsock:receiveany(max?)*
```
agentzh: I think we should make the `max` argument required for the sake of security. Otherwise large volumn of incoming data would crash the LuaJIT VM.

README.markdown
```diff
@@ -7231,6 +7233,40 @@ This feature was first introduced in the `v0.5.0rc1` release.
...
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, ngx.timer.&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;*
+
+Returns any data received by the connected socket, at most `max` bytes.
+
+This method is a synchronous operation just like the [send](#tcpsocksend) method and is 100% nonblocking.
+
+In case of success, it returns the data received; in case of error, it returns `nil` with a string describing the error.
+
+If the received data is more than this size, this method will return with exactly this size of data.
+The remain received data could be returned in the next `receiveany` call.
```
agentzh: Grammar: "The remain received data" => "The remaining data in the underlying receive buffer"

"in the next `receiveany` call" => "in the next reading operation" (it should not necessarily be another `receiveany` call, right? it could also be `receive()` or `receiveuntil()`).

README.markdown
```diff
@@ -7231,6 +7233,40 @@ This feature was first introduced in the `v0.5.0rc1` release.
...
+ sock:settimeouts(1000, 1000, 1000)  -- one second timeout for connect/read/write
+ local data, err = sock:receiveany(10 * 1024 * 1024) -- read any data, at most 10K
+ if not data then
+     ngx.say("failed to read any data: ", err)
+     return
+ end
+ ngx.say("successfully read: ", data)
+```
+
+This method doesn't automatically close the current connection when the read timeout error happens. For other connection errors, this method always automatically closes the connection.
```
agentzh: "happens" => "occurs". The latter reads better ;)

src/ngx_http_lua_input_filters.c
```diff
@@ -0,0 +1,134 @@
...
+
+ngx_int_t
+ngx_http_lua_read_any(ngx_buf_t *src, ngx_chain_t *buf_in, size_t *max,
+    ssize_t bytes, ngx_log_t *log)
+{
+    if (bytes == 0) {
+        return NGX_ERROR;
+    }
+
+    bytes = ((size_t) bytes >= *max) ? (ssize_t) *max : bytes;
```
agentzh: This line is not very readable. Better be

```C
if (bytes >= (ssize_t) *max) {
    bytes = (ssize_t) *max;
}
```

It's safer to convert unsigned values to signed values. Negative values can be much more troublesome when being converted into very large numbers.

spacewander:
@agentzh
Updated.

PR #1331, opened 2018-05-24
bugfix: when variable is cacheable and indexed, setting the request

dndx:
"Host" header should clear any existing cached $host variable value so
that subsequent read returns expected result.

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -227,6 +227,8 @@ struct ngx_http_lua_main_conf_s {
     ngx_int_t            busy_buf_ptr_count;
 #endif
 
+    ngx_int_t            host_index;
+
     unsigned             requires_header_filter:1;
     unsigned             requires_body_filter:1;
     unsigned             requires_capture_filter:1;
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_headers_in.c
```diff
@@ -449,6 +453,9 @@ ngx_http_set_host_header(ngx_http_request_t *r, ngx_http_lua_header_val_t *hv,
         r->headers_in.server = *value;
     }
 
+    var = &r->variables[lmcf->host_index];
+    var->valid = var->not_found = 0;
```
agentzh: Better use 2 assignment statements here for clarity. The whole nginx code base never uses this form of assignment anyway.

src/ngx_http_lua_common.h
```diff
@@ -227,6 +227,8 @@ struct ngx_http_lua_main_conf_s {
     ngx_int_t            busy_buf_ptr_count;
 #endif
 
+    ngx_int_t            host_index;
```
agentzh: I think `host_var_index` is a better field name here since it is clearer.

agentzh:
@dndx Merged. Thanks!

PR #1333, opened 2018-06-01
bugfix: supported base64 decoding string which inserts a newline after every 76 characters according to the RFC 2045 (MIME).

moonming:


p0pr0ck5:
I suspect this is going to break a loooot of existing code...

src/ngx_http_lua_util.c
```diff
@@ -4129,4 +4129,87 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
...
+        77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77,
+        77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77, 77
+    };
+
+    for (i = 0; i < src->len; i++) {
+        if (src->data[i] == '=') {
+            break;
+        }
+
+        if (src->data[i] == '\n') {
```
agentzh: Better save the value of `src->data[i]` into a local variable to make optimizing C compilers' lives easier.

thibaultcha:
@moonming It seems like you enabled the executable bits on those files you modified, probably by mistake?

moonming:
@thibaultcha yes, it's my mistake. thanks.

spacewander:
@moonming 
MIME's base64 encoding is different from standard base64 encoding. I just do some search and find many intereasting results. IMHO, we should write a separete method for MIME's base64 encoding, like the URL safe version.

The difference list could be found in https://en.wikipedia.org/wiki/Base64#Variants_summary_table

A quick reference here:
1. Standard base64 doesn't use CRLF as line separators.
2. MIME base64 will ignore any characters not in the table, including CRLF.

moonming:
@spacewander agreed.  we can add a method for MIME's base64 encoding, and I think this standard base64 support `\n` is enough, What do you think?

spacewander:
@moonming 
If we obey RFC 4648, which is the latest RFC of base64, we should **reject any characters not in** the base encoding alphabet.

Quoted from [RFC 4648](https://www.ietf.org/rfc/rfc4648):
> Implementations MUST reject the encoded data if it contains
   characters outside the base alphabet when interpreting base-encoded
   data, unless the specification referring to this document explicitly
   states otherwise.  Such specifications may instead state, as MIME
   does, that characters outside the base encoding alphabet should
   simply be ignored when interpreting data ("be liberal in what you
   accept").

Some implementations in the real world is broken, maybe they just want to keep backward compatibility with the previous implementation came before the RFC. But it is not a good reason to break our current implementation which complies with the RFC.

agentzh:
@moonming We should add a `decode_base64mime()` method to our existing `ngx.base64` module in lua-resty-core. See

https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/base64.md

I'm closing this.

agentzh:
But yeah, we still need a pure C implementation in the ngx_lua core.

PR #1344, opened 2018-06-20
feature: logged error when failed to run timer.

spacewander:
Previously no error will be logged when there is not resource to run an
expired timer.

### DIFF
--- a/src/ngx_http_lua_timer.c
+++ b/src/ngx_http_lua_timer.c
@@ -345,6 +345,9 @@ ngx_http_lua_ngx_timer_helper(lua_State *L, int every)
 
     ngx_add_timer(ev, delay);
 
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                   "created timer (co: %p delay: %M ms)", tctx->co, delay);
+
     lua_pushinteger(L, 1);
     return 1;
 
@@ -498,6 +501,10 @@ ngx_http_lua_timer_copy(ngx_http_lua_timer_ctx_t *old_tctx)
 
     ngx_add_timer(ev, tctx->delay);
 
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                   "created next timer (co: %p delay: %M ms)", tctx->co,
+                   tctx->delay);
+
     return NGX_OK;
 
 nomem:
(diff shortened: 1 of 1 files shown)


spacewander:
We could also log down the `delay` after introducing an extra bit to distinguish `ngx.timer.every` from `ngx.timer.at`.

src/ngx_http_lua_timer.c
```diff
@@ -568,6 +575,9 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
 
     c = ngx_http_lua_create_fake_connection(tctx.pool);
     if (c == NULL) {
+        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
```
agentzh: Better use the "error" level instead of "alert"?

src/ngx_http_lua_timer.c
```diff
@@ -568,6 +575,9 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
...
+                      "failed to create fake connection to run timer co:%p",
```
agentzh: Better use `(co: %p)` (note the space after colon and the surrounding parentheses) here?

src/ngx_http_lua_timer.c
spacewander: Just follow the level of the first two error message when timer failed to run.

spacewander:
@agentzh 
Updated.

src/ngx_http_lua_timer.c
```diff
@@ -579,6 +589,9 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
 
     r = ngx_http_lua_create_fake_request(c);
     if (r == NULL) {
+        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
+                      "failed to create fake request to run timer (co: %p)",
+                      tctx.co);
```
agentzh: Hmm, can we simply put this under the `failed` code label so that we do not have to repeat ourselves?

spacewander: @agentzh 
The error messages before `goto failed` are different. They are in `failed to do something to run timer` format, but the content in `do something` part is not repeated.

spacewander: @agentzh 
Not all the error messages before `failed` are in the same format, like this one:
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_timer.c#L575

thibaultcha: @agentzh Ping. This seems fine as is?

agentzh: I don't have strong opinion on this one. It's fine as is.

Issue #1354, opened 2018-07-24
segmentation fault might occur when SSL renegotiation happens


tokers:
Hello!

Recently I found one of our Nginx worker process exited  abnormally (segmentation fault). The backtrace is like:

```
nginx: worker process(ngx_http_lua_ssl_cert_handler+0x1c2) [0x5b8313]
nginx: worker process(tls_post_process_client_hello+0x1be) [0x7302e9]
nginx: worker process(ossl_statem_server_post_process_message+0x6c) [0x72dd0f]
nginx: worker process(read_state_machine) [0x71b25d]
nginx: worker process(state_machine) [0x71abc1]
nginx: worker process(ossl_statem_accept+0x1d) [0x71a74d]
nginx: worker process(ssl3_read_bytes+0xed4) [0x6e9846]
nginx: worker process(ssl3_read_internal) [0x6f33e9]
nginx: worker process(ssl3_read+0x38)
nginx: worker process(ssl_read_internal+0x165)
nginx: worker process(SSL_read+0x5b)
nginx: worker process(ngx_ssl_recv+0xc5)
nginx: worker process(ngx_http_v2_read_handler)
nginx: worker process(ngx_http_v2_idle_handler+0x116)
nginx: worker process(ngx_epoll_process_events)
nginx: worker process(ngx_process_events_and_timers+0xd3)
```

The coredump point is inside function `ngx_http_lua_ssl_cert_handler`.

You can reproduce this problem by the following way.

```nginx.conf
server {
    listen 8443 ssl http2;
    ssl_certificate_by_lua_block {
        return;
    }
     ssl_certificate /path/to/cert;
     ssl_certificate_key /path/to/pkey;
     location / {
          return 200;
     }
}
```
```bash
openssl s_client -connect 127.0.0.1:8443 -alpn h2 -reconnect
```
Then type "R" in the interactive mode (trigger the TLS reneogatitaion).

```
ALPN protocol: h2
SSL-Session:
    Protocol  : TLSv1.2
    Cipher    : ECDHE-RSA-AES256-GCM-SHA384
    Session-ID: 8C3B8082AEEDC0AF9894C28A7ED50F2B52AA5C2274B4E15509DFEA097B12CB42
    Session-ID-ctx:
    Master-Key: DABD0D31B6A799DB05E199DC89045833480E47EF8F43DC2DDD36A4851C2026DB05A242580928135214395734343EADF7
    PSK identity: None
    PSK identity hint: None
[...]

PR #1355, opened 2018-07-24
bugfix: a segmentation fault might occur when SSL renegotiation happens

tokers:
Please see #1354 for more details.

Since it is tough to distinguish whether current connection is upgraded
to HTTP/2 in the ngx_http_lua_ssl_cert_handler context. This PR just
returns 0 (marks failed) when SSL renegotiation happens. Nginx already
prohibited the SSL renegotiation, so it is safe.

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -201,6 +201,20 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 
     c = ngx_ssl_get_connection(ssl_conn);
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+
+#if !defined SSL_OP_NO_RENEGOTIATION || nginx_version < 1015002
+
+    if (c->ssl->renegotiation) {
+        ngx_log_error(NGX_LOG_WARN, c->log, 0,
+                      "lua_certificate_by_lua: SSL renegotiation refused");
+        return 0;
+    }
+
+#endif
+
+#endif
+
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, c->log, 0,
                    "ssl cert: connection reusable: %ud", c->reusable);
 
--- a/t/139-ssl-cert-by.t
+++ b/t/139-ssl-cert-by.t
@@ -6,16 +6,17 @@ repeat_each(3);
 
 # All these tests need to have new openssl
 my $NginxBinary = $ENV{'TEST_NGINX_BINARY'} || 'nginx';
-my $openssl_version = eval { `$NginxBinary -V 2>&1` };
+our $openssl_version = eval { `$NginxBinary -V 2>&1` };
 
 if ($openssl_version =~ m/built with OpenSSL (0|1\.0\.(?:0|1[^\d]|2[a-d]).*)/) {
     plan(skip_all => "too old OpenSSL, need 1.0.2e, was $1");
 } else {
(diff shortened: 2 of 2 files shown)


spacewander:
AFAIK, server side SSL renegotiation is disable in Nginx.
Maybe you are using OpenSSL 1.1.0+, so the renegotiation is not disable yet.
This was fixed in the master branch of Nginx, see https://trac.nginx.org/nginx/ticket/1376

tokers:
@spacewander The usage of `SSL_OP_NO_RENEGOTIATION` in Nginx was added recently.
It seems that the option can disable the SSL renegotiation completely. But for the Nginx older than `1.15.2`,  there is a time window between Nginx detects SSL renegotiation and takes the corresponding action, so this segmentation fault can still happens.

I don't know what is your new plan for compatible with the Nginx 1.15.x series, if there is no such plan, maybe it's better to fixup this?

spacewander:
@tokers 
> for the Nginx older than 1.15.2, there is a time window between Nginx detects SSL renegotiation and takes the corresponding action, so this segmentation fault can still happens.

Why not backport the patch to the Nginx older than `1.15.2`? Server SSL renegotiation is disable with the intention to avoid security risk. Even if nobody is using SSL 3.0, work around a fixed bug is not so nice.

tokers:
@spacewander The purpose is not "work around a fixed bug" but solves the segmentation fault problem. Of course put the patch for old Nginx is also ok.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -201,6 +201,10 @@ ngx_http_lua_ssl_cert_handler(ngx_ssl_conn_t *ssl_conn, void *data)
 
     c = ngx_ssl_get_connection(ssl_conn);
 
+    if (c->ssl->renegotiation) {
+        return 0;
```
spacewander: IMHO, we should `return -1` and log with error message.
In the ideal world, in which Nginx and OpenSSL work perfectly, server side renegotiation is disable.
Also, skip the `ssl_cert_by_lua` phase and use the dummy ssl certificate is not always expected in the renegotiation.

tokers:
@spacewander 

> What about wrapping this change with specific version macro?
This change is only required for some versions of Nginx built with some versions of OpenSSL.

Okay, I think we can check the Nginx version number and the `SSL_OP_NO_RENEGOTIATION ` macro.

src/ngx_http_lua_ssl_certby.c
tokers: Why should we return `-1`? `-1` represents the current stage is still in progress.

spacewander: @tokers 
My bad. Yes, it should return `0`.

agentzh:
@tokers Is it possible to add a test case to cover this?

tokers:
@agentzh 

It seems difficult for me, I don't know how to manipulate Test::Nginx to perform the SSL renegotiation. :(

src/ngx_http_lua_ssl_certby.c
tokers: @spacewander 
I have updated this change.

tokers:
@agentzh @spacewander test was added. This problem only occurs when OpenSSL version is between `1.1.0` to `1.1.0g`.

zhuizhuhaomeng:
> @agentzh @spacewander test was added. This problem only occurs when OpenSSL version is between `1.1.0` to `1.1.0g`.
This is a bug of openssl, so we close this PR directyly.

PR #1356, opened 2018-07-25
feature: check the number of parameters for ngx.thread.wait

rainingmaster:

When use `ngx.thread.wait` like this:  
```lua
    local ret, err = ngx.thread.wait()
    ngx.say("blabla")
```
the thread will dead because program reach [here](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_uthread.c#L202) but without any coroutine resume the parent coroutine. I think we should return an error to avoid this case.


zhuizhuhaomeng (approved):

zhuizhuhaomeng:
modified the test case and merged

PR #1358, opened 2018-07-26
feature: implemented ngx_http_lua_ffi_decode_base64mime.

moonming:
releated PR: https://github.com/openresty/lua-nginx-module/pull/1333


src/ngx_http_lua_string.c
```diff
@@ -737,6 +737,26 @@ ngx_http_lua_ffi_decode_base64(const u_char *src, size_t slen, u_char *dst,
...
+    in.data = (u_char *) src;
+    in.len = slen;
+
+    out.data = dst;
+
+    rc = ngx_http_lua_decode_base64mime(&out, &in);
+
+    *dlen = out.len;
+
+    return rc == NGX_OK;
```
spacewander: Why not return `rc` directly and check it with `FFI_OK` in the Lua land, just like other FFI functions?

src/ngx_http_lua_util.c
```diff
@@ -4129,4 +4129,84 @@ ngx_http_lua_cleanup_free(ngx_http_request_t *r, ngx_http_cleanup_pt *cleanup)
...
+
+    if (data_len % 4 == 1) {
+        return NGX_ERROR;
+    }
+
+    s = src->data;
+    d = dst->data;
+
+    for(i = 0; i < src->len; i++) {
+        if (s[i] == '=') {
```
spacewander: Maybe we could set `src->len` to `i+1` in the previous loop and skip this branch?

zhuizhuhaomeng:
merged

PR #1362, opened 2018-08-03
fetch keealive peer from last instead of head

rainingmaster:
In most cases, the expiration time at the last of the queue should be less than that at the head of the queue, so we should give priority to getting peers from the last of the queue to prevent too many expired items.


zhuizhuhaomeng:
keep the same policy with the nginx official.

zhuizhuhaomeng:
It is generally easier to succeed than to fail when using the latest connection.

PR #1370, opened 2018-08-24
Fix segfault when setting cache-control=nil

sharpobject:
Currently this segfaults in ngx_http_set_expires when it calls
ngx_strlcasestrn(NULL, NULL + 0, ...)
NULL + 0 is UB :(

### DIFF
--- a/src/ngx_http_lua_headers_out.c
+++ b/src/ngx_http_lua_headers_out.c
@@ -311,6 +311,17 @@ ngx_http_set_builtin_multi_header(ngx_http_request_t *r,
 
     pa = (ngx_array_t *) ((char *) &r->headers_out + hv->offset);
 
+    if (value->data == NULL) {
+        if (!hv->no_override) {
+            ph = pa->elts;
+            for (i = 0; i < pa->nelts; i++) {
+                ph[i]->hash = 0;
+                ph[i]->value.len = 0;
+            }
+        }
+        return NGX_OK;
+    }
+
     if (pa->elts == NULL) {
         if (ngx_array_init(pa, r->pool, 2, sizeof(ngx_table_elt_t *))
             != NGX_OK)


zhuizhuhaomeng:
cat not contact the author and can not reproduce the bug.

sharpobject:
Hello,

Whether this crashes depends on how your optimizing compiler is feeling that day. A totally reasonable way of compiling this function does not end up crashing.

You may still have trouble if you pass NULL to ngx_strlcasestrn, so I would check for that, rather than checking for a crash.

Thanks!

sharpobject:
It looks like my initial report may have been incorrect. If I recall correctly, this would crash searching for needle="private", which seems to happen in ngx_http_upstream_process_cache_control.

zhuizhuhaomeng:
@sharpobject  I add a test case like this, but can not reproduce this problem?
what am I missing?

```
=== TEST 95: clear response cache-control header
 --- http_config
 proxy_cache_path html/cache levels=1:2 keys_zone=cache_zone:10m inactive=1h max_size=512m use_temp_path=off;   --- config
     location /t {
         proxy_pass http://127.0.0.1:$server_port/upstream;
         proxy_cache cache_zone;
         expires 300;
         header_filter_by_lua_block {
             ngx.header.cache_control = ""
         }        
     }        
 
     location /upstream {
         content_by_lua_block {
             ngx.header.cache_control = "public, max-age=31536000"
             ngx.header.content_length=6
             ngx.say("Hello")
         }        
     }        
 --- request  
 GET /t
 --- response_body
 Hello
 --- ONLY
```

sharpobject:
"" is different from nil. I think with a value of "" you'll get an ngx_str_t with non-NULL data and 0 length. With nil you'll get {NULL, 0}. Relevant code seems to be here: https://github.com/openresty/lua-nginx-module/blob/59d39ca2f0963695052c2593f957053f1a1779a2/src/ngx_http_lua_headers.c#L907

zhuizhuhaomeng:
I have corrected the test case above. I have tried all kinds of test cases and that was one of them.
But none can not reproduce the problem.

sharpobject:
I understand. As I wrote above, your compiler is not forced to compile these functions in such a way that NULL+0 does not equal NULL, but it is allowed to. So a desirable manual testing approach is to somehow check if you are ever passing NULL and NULL+0 as the start and end of the haystack to ngx_strlcasestrn. This may not translate well to an automated perl test. Sorry.

zhuizhuhaomeng:
ngx_http_upstream_process_cache_control is called before ngx_http_set_builtin_multi_header
so ngx.header.cache_control = nil won't affect the ngx_http_upstream_process_cache_control.

sharpobject:
Got it. Sorry, it's been several years since I worked in this codebase, and I couldn't find how else it would end up calling ngx_strlcasestrn looking for "private". I wonder if that search for "private" was actually the result of a change to nginx internal to Cloudflare.

zhuizhuhaomeng:
> 

tks!

PR #1377, opened 2018-09-11
change: avoided running any init_by_lua* code inside signaller process and when testing Nginx configuration.

spacewander:
Previously, when Nginx is run as a signaller like `nginx -p . -s stop`, the code in `init_by_lua*` will be executed. Now this probably unexpected behavior is avoided.

### DIFF
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -647,6 +647,10 @@ ngx_http_lua_init(ngx_conf_t *cf)
 #endif
     ngx_str_t                   name = ngx_string("host");
 
+    if (ngx_process == NGX_PROCESS_SIGNALLER || ngx_test_config) {
+        return NGX_OK;
+    }
+
     lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module);
 
     lmcf->host_var_index = ngx_http_get_variable_index(cf, &name);
--- a/t/160-disable-init-by-lua.t
+++ b/t/160-disable-init-by-lua.t
@@ -5,6 +5,7 @@ repeat_each(2);
 plan tests => repeat_each() * (blocks() * 2);
 
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
+
 my $html_dir = $ENV{TEST_NGINX_HTML_DIR};
 my $http_config = <<_EOC_;
     init_by_lua_block {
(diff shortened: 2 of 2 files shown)


thibaultcha:
It should be doable to write a test for this behaviour with Test::Nginx (e.g. some side-effect triggered by `init_by_lua*`? At least for the sake of preventing a future regression.

Food for thought:

Executing `init_by_lua*` in signaller processes has caused us some pain, so I am all in for this change. However, I can't help but think of:

1. some application out there that depends on this behaviour for some reason,
2. this behaviour actually being desired in the future to run Lua in signaller processes

We could address 1. with a new directive to optionally run the handler in signaller processes (`lua_run_init_in_signaller on;`). And 2. could be addressed with a new handler altogether, if we don't have the init_in_signaller on/off directive (e.g. `signal_by_lua`).

spacewander:
@thibaultcha 
IMHO, people could run their Lua code via `resty` before calling `nginx -s`, so there is always an alternative.

Now tests are added, and I also disable running `init_by_lua*` when testing Nginx configuration in the new commits.

thibaultcha:
@spacewander Good call on `-t` and tests look fine to me :+1:

t/157-disable-init-by-lua.t
```diff
@@ -0,0 +1,132 @@
...
+            if not out then
+                ngx.log(ngx.ERR, err)
+
+            else
+                ngx.log(ngx.WARN, out)
+            end
+        }
+    }
+--- no_error_log
+[error]
```
thibaultcha: @spacewander Mind renaming this test suite to `159-disable-init-by-lua.t`? See #1394 which already deals with 2 test suites named `157-*.t`.

spacewander:
@thibaultcha 
Test file renamed.

t/159-disable-init-by-lua.t
```diff
@@ -0,0 +1,132 @@
...
+            if not out then
+                ngx.log(ngx.ERR, err)
+
+            else
+                ngx.log(ngx.WARN, out)
+            end
+        }
+    }
+--- no_error_log eval
+qr/\[error\] .+ (?!:nginx.pid" failed \(2: No such file or directory\))$/
```
thibaultcha: This test succeeds even without the patch, since the regex here does not match the line written by our test nginx configuration. Updating it myself.

thibaultcha:
I have pushed a new commit which fixes the false positive found in the first test case. It is less fancy, but more robust and simpler.

Lekensteyn:
For future reference:
This issue can only be reproduced with (1) `nginx -t`, or (2) `nginx -s ...` when a block is *not* using `lua_shared_dict`:
https://github.com/openresty/lua-nginx-module/blob/b6875300b7bd2425fd6f93de1e55e47685425b18/src/ngx_http_lua_module.c#L798

When `lua_shared_dict` is enabled, the init handler will not be called at that point and will be processed at a later stage. This oversight was also the cause for a crash with `nginx -t` (#1462) which was fixed by #1463.

thibaultcha:
@Lekensteyn yes, I also have it on my TODO list already to fix a few other edge-cases that I believe aren' addressed yet. PRs welcome!

PR #1378, opened 2018-09-11
bugfix: added missing dd definition in ngx_http_lua_util.h.

spacewander:

### DIFF
--- a/src/ngx_http_lua_util.h
+++ b/src/ngx_http_lua_util.h
@@ -9,6 +9,12 @@
 #define _NGX_HTTP_LUA_UTIL_H_INCLUDED_
 
 
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+
 #include "ngx_http_lua_common.h"
 #include "ngx_http_lua_api.h"
 


src/ngx_http_lua_util.h
```diff
@@ -9,6 +9,12 @@
 #define _NGX_HTTP_LUA_UTIL_H_INCLUDED_
 
 
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
```
agentzh: The thing is that we never include this snippet in our header files. It should only be included in individual compilation units since we need the ability to turn DDEBUG on and off in individial compilation units.

spacewander: @agentzh 
However, `dd` macro is used in `ngx_http_lua_util.h`:
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_util.h#L328

If we don't add the definition, those usages are no op.

thibaultcha: @spacewander Let's remove those maybe? It seems like they would be awfully noisy anyways, plust they could be invoked from the caller's POV instead (e.g., callers of `ngx_http_lua_get_lua_vm()`, wherever the developer is currently debugging).

agentzh: That `dd` line should be guarded by `#ifdef DDEBUG`/`#endif`.

thibaultcha: That's a good solution too :)

spacewander: I submit the change in another pull request： https://github.com/openresty/lua-nginx-module/pull/1418

spacewander:
Replaced by https://github.com/openresty/lua-nginx-module/pull/1418

PR #1379, opened 2018-09-18
feature: added support for ARM64.

spacewander:
now we require OpenResty's LuaJIT branch to work properly on ARM64 since
we make use of the thread.exdata Lua API.

On architectures other than ARM64, the thread.exdata API also saves the
per-request global env table and closure objects for each light thread,
which gives a nice ~10% speedup for the simplest Lua handler location
loaded by wrk over HTTP 1.1.

We now use proper userdata instead of lightuserdata for the
ngx.shared.DICT Lua objects.

we also mask off the bits higher than 47-bit of lightuserdata keys based
on C static variable addresses.

We also implemented the pure C API for the ndk.* API (to be used in
lua-resty-core).

Thanks Dejiang Zhu and Zexuan Luo for the development work of this patch.

Signed-off-by: Yichun Zhang (agentzh) <agentzh@gmail.com>

### DIFF
--- a/src/ngx_http_lua_cache.c
+++ b/src/ngx_http_lua_cache.c
@@ -35,11 +35,14 @@ static ngx_int_t
 ngx_http_lua_cache_load_code(ngx_log_t *log, lua_State *L,
     const char *key)
 {
+#ifndef OPENRESTY_LUAJIT
     int          rc;
     u_char      *err;
+#endif
 
     /*  get code cache table */
-    lua_pushlightuserdata(L, &ngx_http_lua_code_cache_key);
+    lua_pushlightuserdata(L, ngx_http_lua_lightudata_mask(
+                          code_cache_key));
     lua_rawget(L, LUA_REGISTRYINDEX);    /*  sp++ */
 
     dd("Code cache table to load: %p", lua_topointer(L, -1));
@@ -52,6 +55,10 @@ ngx_http_lua_cache_load_code(ngx_log_t *log, lua_State *L,
     lua_getfield(L, -1, key);    /*  sp++ */
 
     if (lua_isfunction(L, -1)) {
+#ifdef OPENRESTY_LUAJIT
+        lua_remove(L, -2);   /*  sp-- */
+        return NGX_OK;
+#else
         /*  call closure factory to gen new closure */
         rc = lua_pcall(L, 0, 1, 0);
         if (rc == 0) {
@@ -73,6 +80,7 @@ ngx_http_lua_cache_load_code(ngx_log_t *log, lua_State *L,
                       key, err);
         lua_pop(L, 2);
         return NGX_ERROR;
+#endif
     }
 
     dd("Value associated with given key in code cache table is not code "
@@ -102,10 +110,13 @@ ngx_http_lua_cache_load_code(ngx_log_t *log, lua_State *L,
 static ngx_int_t
 ngx_http_lua_cache_store_code(lua_State *L, const char *key)
 {
+#ifndef OPENRESTY_LUAJIT
     int rc;
+#endif
 
     /*  get code cache table */
-    lua_pushlightuserdata(L, &ngx_http_lua_code_cache_key);
+    lua_pushlightuserdata(L, ngx_http_lua_lightudata_mask(
+                          code_cache_key));
     lua_rawget(L, LUA_REGISTRYINDEX);
 
     dd("Code cache table to store: %p", lua_topointer(L, -1));
@@ -121,12 +132,14 @@ ngx_http_lua_cache_store_code(lua_State *L, const char *key)
     /*  remove cache table, leave closure factory at top of stack */
     lua_pop(L, 1); /* closure */
 
+#ifndef OPENRESTY_LUAJIT
     /*  call closure factory to generate new closure */
     rc = lua_pcall(L, 0, 1, 0);
     if (rc != 0) {
         dd("Error: failed to call closure factory!!");
         return NGX_ERROR;
     }
+#endif
 
     return NGX_OK;
 }
--- a/src/ngx_http_lua_clfactory.c
+++ b/src/ngx_http_lua_clfactory.c
@@ -15,11 +15,13 @@
 #include "ngx_http_lua_clfactory.h"
 
 
+#ifndef OPENRESTY_LUAJIT
 #define CLFACTORY_BEGIN_CODE "return function() "
 #define CLFACTORY_BEGIN_SIZE (sizeof(CLFACTORY_BEGIN_CODE) - 1)
 
 #define CLFACTORY_END_CODE "\nend"
 #define CLFACTORY_END_SIZE (sizeof(CLFACTORY_END_CODE) - 1)
+#endif
 
 
 /*
@@ -59,6 +61,7 @@
  * length(Instruction) = 4 or 8
  * little endian or big endian
 */
+#ifndef OPENRESTY_LUAJIT
 #define    LUA_LITTLE_ENDIAN_4BYTES_CODE                                     \
     "\x24\x00\x00\x00\x1e\x00\x00\x01\x1e\x00\x80\x00"
 #define    LUA_LITTLE_ENDIAN_8BYTES_CODE                                     \
@@ -75,6 +78,7 @@
 #define    LUA_BIG_ENDIAN_8BYTES_CODE_LEN           (8 + 8 + 8)
 #define    LUAC_HEADERSIZE         12
 #define    LUAC_VERSION            0x51
+#endif
 
 
 /*
@@ -147,6 +151,7 @@
  * ---------------------
 */
 
+#ifndef OPENRESTY_LUAJIT
 #define    POS_SOURCE_STR_LEN      LUAC_HEADERSIZE
 #define    POS_START_LINE          (POS_SOURCE_STR_LEN + sizeof(size_t))
 #define    POS_LAST_LINE           (POS_START_LINE + sizeof(int))
@@ -160,6 +165,7 @@
     (POS_BYTECODE + LUA_LITTLE_ENDIAN_8BYTES_CODE_LEN                        \
     + sizeof(int) + sizeof(int))
 #define    MAX_END_CODE_SIZE       (sizeof(int) + sizeof(int) + sizeof(int))
+#endif
 
 /*
  * taken from chaoslawful:
@@ -225,6 +231,7 @@
 
 /* bytecode for luajit 2.0 */
 
+#ifndef OPENRESTY_LUAJIT
 #define    LJ20_LITTLE_ENDIAN_CODE_STRIPPED                                  \
     "\x14\x03\x00\x01\x00\x01\x00\x03"                                       \
     "\x31\x00\x00\x00\x30\x00\x00\x80\x48\x00\x02\x00"                       \
@@ -275,6 +282,7 @@
 #define    LJ21_BCDUMP_VERSION        2
 #define    LJ20_BCDUMP_VERSION        1
 #define    LJ_SIGNATURE             "\x1b\x4c\x4a"
+#endif
 
 
 typedef enum {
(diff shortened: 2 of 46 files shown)


javierguerragiraldez:
we've tested this at Cloudflare, seems stable enough.

src/ngx_http_lua_common.h
```diff
@@ -213,6 +217,24 @@ struct ngx_http_lua_main_conf_s {
                      * thus it is safe to store the peer data in the main conf.
                      */
 
+    ngx_chain_t                            *body_filter_chain;
+                    /* body_filter_by_lua does not support yielding and
```
agentzh: Should use `body_filter_by_lua*` instead to cover other variants. Please address other similar places.

src/ngx_http_lua_common.h
```diff
@@ -213,6 +217,24 @@ struct ngx_http_lua_main_conf_s {
...
+                     * there cannot be any conflicts among concurrent requests,
+                     * thus it is safe to store the chain data in the main conf.
+                     */
+
+    ngx_http_variable_value_t              *setby_args;
+                    /* set_by_lua does not support yielding and
```
agentzh: Ditto.

src/ngx_http_lua_common.h
```diff
@@ -213,6 +217,24 @@ struct ngx_http_lua_main_conf_s {
...
+                     * there cannot be any conflicts among concurrent requests,
+                     * thus it is safe to store the args point in the main conf.
+                     */
+
+    size_t                                  setby_nargs;
+                    /* set_by_lua does not support yielding and
```
agentzh: Ditto.

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
     lua_setglobal(L, "ngx");
 
     ngx_http_lua_inject_coroutine_api(log, L);
+
+#ifdef OPENRESTY_LUAJIT
+    {
+        const char buf[] =
```
agentzh: We should put the `int rc` declaration here as well. `rc` is not used elsewhere and we should limit its lifetime.

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+        return;
+    }
+
+    rc = lua_pcall(L, 0, 0, 0);
+    if (rc != 0) {
+        ngx_log_error(NGX_LOG_ERR, log, 0,
+                      "failed to run Lua code for _G write guard: %s",
+                      rc, lua_tostring(L, -1));
+        lua_pop(L, 1);
+    }
```
agentzh: These 2 `if` statements should be in the same block as the `buf` variable.

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+            "end\n"
+            "setmetatable(_G, { __newindex  = newindex })\n"
+            ;
+
+        rc = luaL_loadbuffer(L, buf, sizeof(buf) - 1, "=_G write guard");
+    }
+
+    if (rc != 0) {
+        ngx_log_error(NGX_LOG_ERR, log, 0,
+                      "failed to load Lua code for _G write guard: %i: %s",
```
agentzh: grammar: "for _G write guard" => "for the _G write guard".

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+                                       "', value type: ', type(value), '\\n', "
+                                       "traceback())\n"
+                "else\n"
+                    "ngx_log(ngx_WARN, 'lua http setting global variable, "
+                                       "key type: ', key_type, "
+                                       "', value type: ', type(value), '\\n', "
+                                       "traceback())\n"
+                "end\n"
+            "end\n"
+            "setmetatable(_G, { __newindex  = newindex })\n"
```
agentzh: Style: 2 spaces before `=`. Only one space should be used there.

src/ngx_http_lua_cache.c
```diff
@@ -73,6 +80,7 @@ ngx_http_lua_cache_load_code(ngx_log_t *log, lua_State *L,
                       key, err);
         lua_pop(L, 2);
         return NGX_ERROR;
+#endif
```
agentzh: Better add a comment here after `#endif` to indicate what condition for this macro branch? This branch is a bit large.

src/ngx_http_lua_clfactory.c
```diff
@@ -75,6 +78,7 @@
 #define    LUA_BIG_ENDIAN_8BYTES_CODE_LEN           (8 + 8 + 8)
 #define    LUAC_HEADERSIZE         12
 #define    LUAC_VERSION            0x51
+#endif
```
agentzh: Ditto. Better add a comment to make it clear what condition is this `#endif` belonged to.

src/ngx_http_lua_clfactory.c
```diff
@@ -160,6 +165,7 @@
     (POS_BYTECODE + LUA_LITTLE_ENDIAN_8BYTES_CODE_LEN                        \
     + sizeof(int) + sizeof(int))
 #define    MAX_END_CODE_SIZE       (sizeof(int) + sizeof(int) + sizeof(int))
+#endif
```
agentzh: Ditto. Please address other similar places too.

src/ngx_http_lua_clfactory.c
```diff
@@ -275,6 +282,7 @@
 #define    LJ21_BCDUMP_VERSION        2
 #define    LJ20_BCDUMP_VERSION        1
 #define    LJ_SIGNATURE             "\x1b\x4c\x4a"
+#endif
```
agentzh: Ditto.

t/001-set.t
```diff
@@ -592,25 +592,29 @@ failed to run set_by_lua*: unknown reason
...
             return foo
-        ';
+        }
         echo $res;
     }
 --- request
 GET /lua
---- response_body
-1
+--- response_body_like
```
agentzh: Should chomp the last trailing newline.

t/001-set.t
```diff
@@ -592,25 +592,29 @@ failed to run set_by_lua*: unknown reason
...
+^[12]$
```
agentzh: Better use `\A` and `\z` to avoid any potential ambiguities?

t/014-bugs.t
```diff
@@ -593,12 +593,12 @@ $s
     }
 --- pipelined_requests eval
 ["GET /test", "GET /test", "GET /test"]
---- response_body eval
-["0", "0", "0"]
+--- response_body_like eval
+[qr/^(0|3|6)$/, qr/^(1|4|7)$/, qr/^(2|5|8)$/]
```
agentzh: Please do not use capturing groups in such regexes. We should use `(?: )` instead to avoid useless submatch capturing. Please address other similar places in the tests. Also, it seems that it's simpler to just use `[036]` instead of `(?:0|3|6)` here, for example?

t/014-bugs.t
```diff
@@ -613,12 +613,12 @@ $s
...
```
agentzh: Ditto.

t/001-set.t
```diff
@@ -592,25 +592,29 @@ failed to run set_by_lua*: unknown reason
...
 --- no_error_log
 [error]
+--- grep_error_log eval: qr/old foo: \d+/
+--- grep_error_log_out eval
+["", "old foo: 1\n"]
```
agentzh: This should generate write guard warnings. Let's check for them in these tests writing to globals in request contexts.

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+                "rawset(table, key, value)\n"
+                "local phase = ngx_get_phase()\n"
+                "if phase == 'init_worker' or phase == 'init' then\n"
+                    "return\n"
+                "end\n"
+                "local key_type = type(key)\n"
+                "if key_type == 'string' then\n"
+                    "ngx_log(ngx_WARN, 'lua http setting global variable, "
+                                       "key: ', key, "
+                                       "', value type: ', type(value), '\\n', "
```
agentzh: This error message is not very intuitive. `key` does not make much sense to the end users since they may just write to a global variable, not invoking the metamethod explicitly via a key. Something like below would be much clearer:

```
lua http attempting to write to the global variable 'foo'
```

It's not very useful to log the type of the value IMHO. Let's make it concise.

Also, it seems that there's no test cases covering this warning message at all?

src/ngx_http_lua_util.c
```diff
@@ -756,6 +767,57 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+                                       "traceback())\n"
+                "else\n"
+                    "ngx_log(ngx_WARN, 'lua http setting global variable, "
+                                       "key type: ', key_type, "
+                                       "', value type: ', type(value), '\\n', "
+                                       "traceback())\n"
```
agentzh: Like above, we should log something like this:

```
lua http attempting to write to the global variable 'table: 0x40e6ca60'
```

So we can actually merge these 2 code branches and just use `tostring(key)` instead. It works on both the string-typed variable name and other types.

spacewander:
@agentzh 
Updated.

src/ngx_http_lua_util.c
```diff
@@ -756,6 +763,50 @@ ngx_http_lua_inject_ngx_api(lua_State *L, ngx_http_lua_main_conf_t *lmcf,
...
+            "local type = type\n"
+            "local ngx_get_phase = ngx.get_phase\n"
+            "local traceback = require 'debug'.traceback\n"
+            "local function newindex(table, key, value)\n"
+                "rawset(table, key, value)\n"
+                "local phase = ngx_get_phase()\n"
+                "if phase == 'init_worker' or phase == 'init' then\n"
+                    "return\n"
+                "end\n"
+                "ngx_log(ngx_WARN, 'lua http attempting to write to the lua "
```
agentzh: Hmm, after some more thinking, it's better to use the prefix "http lua" instead? So basically the log message prefix could be

```
<subsys> <module>
```

So we should use "stream lua" in the `ngx_stream_lua_module`.

t/001-set.t
agentzh: we should cover the log prefix "http lua attempting to" here as well. Also we should check if the error log level is warn. The current pattern does not cover these bits.

t/157-global-var.t
```diff
@@ -0,0 +1,489 @@
...
+            end
+            return foo
+        }
+        echo $res;
+    }
+--- response_body_like eval
+qr/[12]/
+--- grep_error_log eval: qr/(old foo: \d+|write to the lua global variable '\w+')/
+--- grep_error_log_out eval
+["write to the lua global variable 'foo'\n", "old foo: 1\n"]
```
agentzh: Should cover the log level (`warn`) and the Lua backtrace (is there a backtrace logged at all?).

t/157-global-var.t
```diff
@@ -0,0 +1,489 @@
...
+            else
+                ngx.log(ngx.WARN, "old foo: ", foo)
+                foo = foo + 1
+            end
+            return foo
+        }
+        echo $res;
+    }
+--- response_body_like eval
+qr/[12]/
```
agentzh: This regex is too loose...Output like `hello, 1234` will also match this regex happily.

Issue #1371, opened 2018-08-27
Add UNSUPPORTED_MEDIA_TYPE constant


RussDragon:
Hello. Is it possible to add UNSUPPORTED_MEDIA_TYPE (415) constant to the list of HTTP constants? 

Thank you.

kwanhur:
see this PR[1380](https://github.com/openresty/lua-nginx-module/pull/1380)

PR #1380, opened 2018-09-20
feature: added support constant HTTP status code 415 #1371

kwanhur:


zhuizhuhaomeng:
we don't want to add more variables into  ngx since it is to large.

PR #1381, opened 2018-09-20
change: tcpsock:setkeepalive: do not put the connection to the pool when the process is exiting.

doujiang24:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -7453,8 +7453,6 @@ Idle connections in the pool will be monitored for any exceptional events like c
 
 In case of success, this method returns `1`; otherwise, it returns `nil` and a string describing the error.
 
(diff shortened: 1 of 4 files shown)


agentzh:
Making `setkeepalive()` return errors is too overkill and is not the right way to fix the shutdown problem.

Actually nginx already provides the `ngx_event_t.cancelable` flag for this purposes and nginx's own syslog transport already makes use of this. We should do that here as well.

BTW, the standard `ngx_http_upstream_keepalive` module also has the same problem in that its `keepalive_timeout` timer does not set the `cancelable` flag for its corresponding event construct.

tokers:
We also can check the return of `ngx.exit()` before calling `tcpsock:setkeepalive()`.

agentzh:
Merged with edits in your commit log message.

PR #1401, opened 2018-10-29
misc: updated _G write guard message to be more accurate.

thibaultcha:

* more appropriate message ("writing" instead of "attempting to")
* remove http/lua prefix (`ngx.log` will inject `[lua]` and `stream [lua]`
  already)
* remove unused cached global
* cache `tostring` global locally


thibaultcha:
About to push tests updates.

PR #1402, opened 2018-10-30
feature: implemented ngx.req module and its function add_header().

spacewander:
The `ngx.req.add_header` works mostly like `ngx.req.set_header`.
However, unlike `ngx.req.set_header`, this method appends new header to the old
one instead of overriding it if the header is not builtin.

For the builtin header, this method overrides it instead of appending to it.
IMHO, it is reasonable since appending to the builtin header like `User-Agent`
is confusing and semantically wrong.

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -955,9 +955,10 @@ nomem:
 
 
 int
-ngx_http_lua_ffi_req_set_header(ngx_http_request_t *r,
-    const u_char *key, size_t key_len, const u_char *value, size_t value_len,
-    ngx_http_lua_ffi_str_t *mvals, size_t mvals_len, int override)
+ngx_http_lua_ffi_req_set_header(ngx_http_request_t *r, const u_char *key,
+    size_t key_len, const u_char *value, size_t value_len,
+    ngx_http_lua_ffi_str_t *mvals, size_t mvals_len, int override,
+    char **errmsg)
 {
     u_char                      *p;
     size_t                       len;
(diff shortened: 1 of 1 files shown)


spacewander:
This is the sister pr of https://github.com/openresty/lua-resty-core/pull/213

src/ngx_http_lua_headers.c
```diff
@@ -1322,9 +1322,13 @@ ngx_http_lua_ffi_set_resp_header(ngx_http_request_t *r, const u_char *key_data,
 
 
 int
-ngx_http_lua_ffi_req_header_set_single_value(ngx_http_request_t *r,
-    const u_char *key, size_t key_len, const u_char *value, size_t value_len)
+ngx_http_lua_ffi_req_set_header(ngx_http_request_t *r,
```
thibaultcha: We should align this function's name with `ngx_http_lua_ffi_set_resp_header`. It should thus be `ngx_http_lua_ffi_set_req_header`.

src/ngx_http_lua_headers.c
```diff
@@ -1345,22 +1349,51 @@ ngx_http_lua_ffi_req_header_set_single_value(ngx_http_request_t *r,
...
+                }
+
+                ngx_memcpy(v.data, p, len);
+                v.data[len] = '\0';
+                v.len = len;
+
+                if (ngx_http_lua_set_input_header(r, k, v, override)
+                    != NGX_OK)
+                {
+                    return NGX_ERROR;
```
thibaultcha: `ngx_http_lua_ffi_set_resp_header` has a slightly different implementation. I believe this implementation should closely mirror it. Additionally, `ngx_http_lua_ffi_set_resp_header` also sets error messages on "failed to set header" or "no memory" errors. I think that we should do the same here.

thibaultcha:
Merged!

PR #1407, opened 2018-10-31
feature: implemented the pure C API to support the ngx.pipe API provi…

spacewander:
…ded by lua-resty-core.

Now we could spawn a process and interact with its stdin/stdout/stderr
without blocking any OS threads.

Signed-off-by: Yichun Zhang (agentzh) <agentzh@gmail.com>

### DIFF
--- a/src/ngx_http_lua_initworkerby.c
+++ b/src/ngx_http_lua_initworkerby.c
@@ -12,6 +12,7 @@
 
 #include "ngx_http_lua_initworkerby.h"
 #include "ngx_http_lua_util.h"
+#include "ngx_http_lua_pipe.h"
 
 
 static u_char *ngx_http_lua_log_init_worker_error(ngx_log_t *log,
@@ -65,6 +66,13 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
         return NGX_OK;
     }
+
+#ifdef HAVE_NGX_LUA_PIPE
+    if (ngx_http_lua_pipe_add_signal_handler(cycle) != NGX_OK) {
+        return NGX_ERROR;
+    }
+#endif
+
 #endif  /* NGX_WIN32 */
 
 #if (NGX_HTTP_LUA_HAVE_SA_RESTART)
--- a/src/ngx_http_lua_module.c
+++ b/src/ngx_http_lua_module.c
@@ -29,6 +29,7 @@
 #include "ngx_http_lua_ssl_session_storeby.h"
 #include "ngx_http_lua_ssl_session_fetchby.h"
 #include "ngx_http_lua_headers.h"
+#include "ngx_http_lua_pipe.h"
 
 
 static void *ngx_http_lua_create_main_conf(ngx_conf_t *cf);
@@ -742,6 +743,11 @@ ngx_http_lua_init(ngx_conf_t *cf)
 
     cln->data = lmcf;
     cln->handler = ngx_http_lua_sema_mm_cleanup;
+
+#ifdef HAVE_NGX_LUA_PIPE
+    ngx_http_lua_pipe_init();
+#endif
+
 #endif
 
 #if nginx_version >= 1011011
--- a/src/ngx_http_lua_pipe.c
+++ b/src/ngx_http_lua_pipe.c
@@ -1,0 +1,2487 @@
+
+/*
+ * Copyright (C) by OpenResty Inc.
+ */
+
+
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+
+#include "ngx_http_lua_common.h"
+#include "ngx_http_lua_input_filters.h"
+#include "ngx_http_lua_util.h"
+#include "ngx_http_lua_pipe.h"
+#if (NGX_HTTP_LUA_HAVE_SIGNALFD)
+#include <sys/signalfd.h>
+#endif
+
+
+#ifdef HAVE_NGX_LUA_PIPE
+static ngx_rbtree_node_t *ngx_http_lua_pipe_lookup_pid(ngx_rbtree_key_t key);
+#if !(NGX_HTTP_LUA_HAVE_SIGNALFD)
+static void ngx_http_lua_pipe_sigchld_handler(int signo, siginfo_t *siginfo,
+    void *ucontext);
+#endif
+static void ngx_http_lua_pipe_sigchld_event_handler(ngx_event_t *ev);
+static ssize_t ngx_http_lua_pipe_fd_read(ngx_connection_t *c, u_char *buf,
+    size_t size);
+static ssize_t ngx_http_lua_pipe_fd_write(ngx_connection_t *c, u_char *buf,
+    size_t size);
+static ngx_int_t ngx_http_lua_pipe_close_helper(
+    ngx_http_lua_pipe_ctx_t *pipe_ctx, ngx_event_t *ev, int forced);
+static ngx_int_t ngx_http_lua_pipe_close_stdin(ngx_http_lua_pipe_t *pipe,
+    int forced);
+static ngx_int_t ngx_http_lua_pipe_close_stdout(ngx_http_lua_pipe_t *pipe,
+    int forced);
+static ngx_int_t ngx_http_lua_pipe_close_stderr(ngx_http_lua_pipe_t *pipe,
+    int forced);
+static void ngx_http_lua_pipe_proc_finalize(ngx_http_lua_ffi_pipe_proc_t *proc,
+    int forced);
+static ngx_int_t ngx_http_lua_pipe_get_lua_ctx(ngx_http_request_t *r,
+    ngx_http_lua_ctx_t **ctx, u_char *errbuf, size_t *errbuf_size);
+static void ngx_http_lua_pipe_put_error(ngx_http_lua_pipe_ctx_t *pipe_ctx,
+    u_char *errbuf, size_t *errbuf_size);
+static void ngx_http_lua_pipe_put_data(ngx_http_lua_pipe_t *pipe,
+    ngx_http_lua_pipe_ctx_t *pipe_ctx, u_char **buf, size_t *buf_size);
+static ngx_int_t ngx_http_lua_pipe_add_input_buffer(ngx_http_lua_pipe_t *pipe,
+    ngx_http_lua_pipe_ctx_t *pipe_ctx);
+static ngx_int_t ngx_http_lua_pipe_read_all(void *data, ssize_t bytes);
+static ngx_int_t ngx_http_lua_pipe_read_bytes(void *data, ssize_t bytes);
+static ngx_int_t ngx_http_lua_pipe_read_line(void *data, ssize_t bytes);
+static ngx_int_t ngx_http_lua_pipe_read_any(void *data, ssize_t bytes);
+static ngx_int_t ngx_http_lua_pipe_read(ngx_http_lua_pipe_t *pipe,
+    ngx_http_lua_pipe_ctx_t *pipe_ctx);
+static ngx_int_t ngx_http_lua_pipe_init_ctx(
+    ngx_http_lua_pipe_ctx_t **pipe_ctx_pt, int fd, ngx_pool_t *pool,
+    u_char *errbuf, size_t *errbuf_size);
+static ngx_int_t ngx_http_lua_pipe_write(ngx_http_lua_pipe_t *pipe,
+    ngx_http_lua_pipe_ctx_t *pipe_ctx);
+static int ngx_http_lua_pipe_read_stdout_retval(
+    ngx_http_lua_ffi_pipe_proc_t *proc, lua_State *L);
+static int ngx_http_lua_pipe_read_stderr_retval(
+    ngx_http_lua_ffi_pipe_proc_t *proc, lua_State *L);
+static int ngx_http_lua_pipe_read_retval_helper(
+    ngx_http_lua_ffi_pipe_proc_t *proc, lua_State *L, int from_stderr);
+static int ngx_http_lua_pipe_write_retval(ngx_http_lua_ffi_pipe_proc_t *proc,
+    lua_State *L);
+static int ngx_http_lua_pipe_wait_retval(ngx_http_lua_ffi_pipe_proc_t *proc,
+    lua_State *L);
+static void ngx_http_lua_pipe_resume_helper(ngx_event_t *ev,
+    ngx_http_lua_co_ctx_t *wait_co_ctx);
+static void ngx_http_lua_pipe_resume_read_stdout_handler(ngx_event_t *ev);
+static void ngx_http_lua_pipe_resume_read_stderr_handler(ngx_event_t *ev);
+static void ngx_http_lua_pipe_resume_write_handler(ngx_event_t *ev);
+static void ngx_http_lua_pipe_resume_wait_handler(ngx_event_t *ev);
+static ngx_int_t ngx_http_lua_pipe_resume(ngx_http_request_t *r);
+static void ngx_http_lua_pipe_dummy_event_handler(ngx_event_t *ev);
+static void ngx_http_lua_pipe_cleanup_helper(
+    ngx_http_lua_co_ctx_t *wait_co_ctx);
+static void ngx_http_lua_pipe_proc_read_stdout_cleanup(void *data);
+static void ngx_http_lua_pipe_proc_read_stderr_cleanup(void *data);
+static void ngx_http_lua_pipe_proc_write_cleanup(void *data);
+static void ngx_http_lua_pipe_proc_wait_cleanup(void *data);
+
+
+static ngx_rbtree_t       ngx_http_lua_pipe_rbtree;
+static ngx_rbtree_node_t  ngx_http_lua_pipe_proc_sentinel;
+
+
+#if (NGX_HTTP_LUA_HAVE_SIGNALFD)
+static int                                ngx_http_lua_signalfd;
+static struct signalfd_siginfo            ngx_http_lua_pipe_notification;
+
+#define ngx_http_lua_read_sigfd           ngx_http_lua_signalfd
+
+#else
+static int                                ngx_http_lua_sigchldfd[2];
+static u_char                             ngx_http_lua_pipe_notification[1];
+
+#define ngx_http_lua_read_sigfd           ngx_http_lua_sigchldfd[0]
+#define ngx_http_lua_write_sigfd          ngx_http_lua_sigchldfd[1]
+#endif
+
+
+static ngx_connection_t                  *ngx_http_lua_sigfd_conn = NULL;
+
+
+/* The below signals are ignored by Nginx.
+ * We need to reset them for the spawned child processes. */
+ngx_http_lua_pipe_signal_t ngx_signals[] = {
+    { SIGSYS, "SIGSYS" },
+    { SIGPIPE, "SIGPIPE" },
+    { 0, NULL }
+};
+
+
+enum {
+    PIPE_ERR_CLOSED = 1,
+    PIPE_ERR_SYSCALL,
+    PIPE_ERR_NOMEM,
+    PIPE_ERR_TIMEOUT,
+    PIPE_ERR_ADD_READ_EV,
+    PIPE_ERR_ADD_WRITE_EV
+};
+
+
+enum {
+    PIPE_READ_ALL = 0,
+    PIPE_READ_BYTES,
+    PIPE_READ_LINE,
+    PIPE_READ_ANY
+};
+
+
+#define REASON_EXIT         "exit"
+#define REASON_SIGNAL       "signal"
+#define REASON_UNKNOWN      "unknown"
+
+#define REASON_RUNNING_CODE  0
+#define REASON_EXIT_CODE     1
+#define REASON_SIGNAL_CODE   2
+#define REASON_UNKNOWN_CODE  3
+
+
+void
+ngx_http_lua_pipe_init(void)
+{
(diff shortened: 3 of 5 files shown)


spacewander:
This is the sister pr of https://github.com/openresty/lua-resty-core/pull/215

thibaultcha:
This PR also needs rebasing; will you take care of it @spacewander? Thanks!

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2458 @@
...
+{
+    switch (pipe_ctx->err_type) {
+
+    case NGX_HTTP_LUA_PIPE_ET_CLOSED:
+        *errbuf_size =
+            ngx_snprintf(errbuf, *errbuf_size, "closed") - errbuf;
+        break;
+
+    case NGX_HTTP_LUA_PIPE_ET_ERROR:
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size, "error") - errbuf;
```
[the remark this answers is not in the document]
spacewander: @thibaultcha 
Well, errors which are rare and could not be classified will be marked as `NGX_HTTP_LUA_PIPE_ET_ERROR`. Because we always log the error when this type of error happened, it is acceptable to use `"error"` as the returned error message.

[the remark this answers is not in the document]
agentzh: @spacewander Can we avoid automatic logging but return the detailed error message to the Lua caller? It's much easier to handle errors on the Lua land than messing up with error log messages. The cosocket API also provides the `lua_socket_log_errors` directive to turn off automatic logging completely.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2458 @@
...
+failed:
+
+    if (ngx_http_lua_sigfd_conn != NULL) {
+        ngx_close_connection(ngx_http_lua_sigfd_conn);
+        ngx_http_lua_sigfd_conn = NULL;
+    }
+
+    if (close(ngx_http_lua_read_sigfd) == -1) {
+        ngx_log_error(NGX_LOG_EMERG, ngx_cycle->log, ngx_errno,
+                      "failed to close the read sigfd");
```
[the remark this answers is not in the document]
agentzh: +1

src/ngx_http_lua_pipe.h
```diff
@@ -0,0 +1,96 @@
...
+
+
+#include "ngx_http_lua_common.h"
+
+
+#define NGX_HTTP_LUA_PIPE_ET_ERROR         0x0001
+#define NGX_HTTP_LUA_PIPE_ET_CLOSED        0x0002
+#define NGX_HTTP_LUA_PIPE_ET_SYSCALL       0x0003
+#define NGX_HTTP_LUA_PIPE_ET_NOMEM         0x0004
+#define NGX_HTTP_LUA_PIPE_ET_TIMEOUT       0x0005
```
agentzh: What does `ET` stand for? "error tag"? Does not make sense...

agentzh: I wonder if we could just use an anonymous enum to define these, which can be easier and also cleaner.

spacewander: @agentzh 
The `ET` is an abbreviation for `err_type`, like the `FT` is an abbreviation for `ft_type` in cosocket API implementation.

spacewander:
@agentzh @thibaultcha 
Updated and rebased.

spacewander:
@thibaultcha 
Now we only enable ngx.pipe when the patch is applied, so that we have less situations to concern about.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2458 @@
...
+#endif
+
+    if (ngx_handle_read_event(rev, 0) == NGX_ERROR) {
+        goto failed;
+    }
+
+#if !(NGX_HTTP_LUA_HAVE_SIGNALFD)
+    ngx_memzero(&sa, sizeof(struct sigaction));
+    sa.sa_sigaction = ngx_http_lua_pipe_sigchld_handler;
+    sa.sa_flags = SA_SIGINFO;
```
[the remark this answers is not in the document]
spacewander: @thibaultcha 
I think this problem is fixed? As now `ngx_http_lua_pipe_add_signal_handler` runs earlier than the `ngx_http_lua_set_sa_restart`. And all tests are passed with `ngx.pipe` enabled.
ngx_http_lua_pipe_add_signal_handler: https://github.com/openresty/lua-nginx-module/blob/76acadac4d09e2aee735bbcd2c9a2287c3d51015/src/ngx_http_lua_initworkerby.c#L74
ngx_http_lua_set_sa_restart: https://github.com/openresty/lua-nginx-module/blob/76acadac4d09e2aee735bbcd2c9a2287c3d51015/src/ngx_http_lua_initworkerby.c#L83

spacewander:
@thibaultcha 
Updated again.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2476 @@
...
+ * We need to reset them for the spawned child processes */
+ngx_http_lua_pipe_signal_t ngx_signals[] = {
+    SIG(SYS),
+    SIG(PIPE),
+    { 0, NULL }
+};
+
+
+enum {
+    PIPE_ERR_CLOSED = 1,
```
thibaultcha: I think it makes more sense to let this enum's values be anonymous? In the cosocket implementation, enums start at 1 because they are used as indexes in Lua tables, but no reason for this enum not to start at 0.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2476 @@
...
+    PIPE_ERR_ADD_READ_EV,
+    PIPE_ERR_ADD_WRITE_EV
+};
+
+
+enum {
+    PIPE_READ_ALL     = 0,
+    PIPE_READ_BYTES   = 1,
+    PIPE_READ_LINE    = 2,
+    PIPE_READ_ANY     = 3
```
thibaultcha: Ditto, this enum's values could be anonymous as well. Besides, aligning the assignment operator typically makes for messy diffs in the future. But the problem doesn't even arise itself if we let those values (or even all but the first one) be anonymously defined. So that's one more reason.

spacewander:
@thibaultcha 
Updated again.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2476 @@
...
+    }
+
+    rc = sigaddset(&set, SIGCHLD);
+    if (rc != 0) {
+        ngx_log_error(NGX_LOG_ERR, cycle->log, ngx_errno,
+                      "lua pipe add SIGCHLD to signal set failed");
+        return NGX_ERROR;
+    }
+
+    rc = sigprocmask(SIG_BLOCK, &set, NULL);
```
thibaultcha: Since we are blocking `SIGCHLD` (or overriding its signal handler when we don't have `signalfd`), how come we are not respecting Nginx's original handling of the signal? https://github.com/nginx/nginx/blob/release-1.13.6/src/os/unix/ngx_process.c#L401

thibaultcha:
Another round of review, almost done here. I also pushed a small cleanup commit for all minor stuff.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2478 @@
...
+    }
+
+    ngx_http_lua_sigfd_conn->log = cycle->log;
+    ngx_http_lua_sigfd_conn->recv = ngx_http_lua_pipe_fd_read;
+    rev = ngx_http_lua_sigfd_conn->read;
+    rev->log = ngx_http_lua_sigfd_conn->log;
+    rev->handler = ngx_http_lua_pipe_sigchld_event_handler;
+
+#ifdef HAVE_SOCKET_CLOEXEC_PATCH
+    rev->skip_socket_leak_check = 1;
```
thibaultcha: What is this struct member? I cannot compile this branch in my environment (where I apply the socket_cloexec patch) because of this:

```
lua-nginx-module/src/ngx_http_lua_pipe.c:1460:12: error: ‘ngx_event_t’ {aka ‘struct ngx_event_s’} has no member named ‘skip_socket_leak_check’
     c->read->skip_socket_leak_check = 1;
            ^~
make[1]: *** [objs/Makefile:2540: objs/addon/src/ngx_http_lua_pipe.o] Error 1
make[1]: Leaving directory 'lua-nginx-module/buildroot/nginx-1.13.6'
make: *** [Makefile:8: build] Error 2
```

The socket_cloexec patch does not seem to have any mention of it: https://github.com/openresty/openresty/blob/master/patches/nginx-1.13.6-socket_cloexec.patch

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2476 @@
...
+    pipe->dead = 1;
+
+    ngx_http_lua_pipe_close_stdin(pipe, forced);
+    ngx_http_lua_pipe_close_stdout(pipe, forced);
+
+    if (!pipe->merge_stderr) {
+        ngx_http_lua_pipe_close_stderr(pipe, forced);
+    }
+
+    pipe->closed = 1;
```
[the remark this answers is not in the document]
spacewander: @thibaultcha 
* `dead` is used when the process is gone, but we are still able to read/write to the buffer.
* `closed` is used when the process is 'closed' manually, and we could not read/write to the buffer.

`pipe->dead` could be `1` when `pipe->closed` is `0`.

src/ngx_http_lua_pipe.c
spacewander: @thibaultcha 
Nginx's original handling only works in the master process, and we block `SIGCHLD` in the worker process. What we lose is only the NOTICE level error message: `signal %d (%s) received from xxx`. We could add the error message in our handler.

src/ngx_http_lua_pipe.c
spacewander: Err, you need to update the `socket_cloexec.patch` from this sister PR: https://github.com/openresty/openresty/pull/412

src/ngx_http_lua_pipe.c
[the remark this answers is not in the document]
spacewander: Sorry for the late reply, I just suffered a network problem. The reason why I chose `dead` instead of `exited` is that you could do something with a `dead` process, but you could not do something with an `exited` process. Admittedly, the `pipe->closed` is suitable for the read/write operation, but not for the `wait` operation.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2476 @@
...
+    if (rc != NGX_OK) {
+        return rc;
+    }
+
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                   "lua pipe read process:%p pid:%P", proc, proc->_pid);
+
+    pipe = proc->pipe;
+    if (pipe == NULL || pipe->closed) {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size, "closed") - errbuf;
```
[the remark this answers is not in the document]
spacewander: I could not really agree that distinguishing `exited` from `closed` is a good idea...
In caller's perspective, there is not much difference between `closed` and `exited`, and checking `err ~= 'closed` and err ~= 'exited'`  is cumbersome.

@agentzh, what is your idea?

src/ngx_http_lua_pipe.c
thibaultcha: Yeah, worker processes usually don’t have much to reap... I did have this NOTICE log in mind though when writing the above comment (after seeing it so much when working on the SA_RESTART fix).

src/ngx_http_lua_pipe.c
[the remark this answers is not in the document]
spacewander: > In caller's perspective, there is not much difference between closed and exited

I think I need to clarify my words...
Whether the process is exited or not doesn't affect to the read/write operations. The caller is still able to read/write the pipe buffer even after the target process is exited. (You will get a `EPIPE` when writing to such buffer, but writing to a closed stdin also gets the same error) The reason why I chose `pipe->closed` and returned `closed` is that the `pipe` is actually closed by the caller manually.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2473 @@
...
+    rc = ngx_http_lua_pipe_get_lua_ctx(r, &ctx, errbuf, errbuf_size);
+    if (rc != NGX_OK) {
+        return rc;
+    }
+
+    ngx_log_debug2(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                   "lua pipe wait process:%p pid:%P", proc, proc->_pid);
+
+    pipe = proc->pipe;
+    if (pipe == NULL || pipe->closed) {
```
thibaultcha: I think we should check for `pipe->dead` here instead?

src/ngx_http_lua_pipe.c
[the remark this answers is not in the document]
thibaultcha: > In caller's perspective, there is not much difference between closed and exited, and checking err ~= 'closed‘ and err ~= 'exited' is cumbersome.

Yes, that is exactly what I meant, and I wanted to avoid the some thus APIs had to check for `"exited"` and others for `"closed"` (which could easily lead to user errors...).

That said, I did realize that each error message had a precise meaning:

> There are some instances when we return "closed" in which it is fine, of course (e.g. a pipe is closed, which is different from the process being exited).

However, in hindsight to what was said in this thread and the other one, I do agree with this design. The documentation can be clarified to explain this, I will make sure to do so when reviewing the docs in particular (I have yet to push my changes and comments there for now, but it is ongoing).

src/ngx_http_lua_pipe.c
spacewander: @thibaultcha 
We may wait for a `dead` process (the process is dead just before we call `wait`).

thibaultcha: So we may wait for a `dead` process, but we may not wait when the pipe is `closed`? Why so?

src/ngx_http_lua_pipe.c
spacewander: @thibaultcha 
I prefer adding it in `ngx_http_lua_pipe_sigchld_event_handler`, since the signal handler will trigger this event handler eventually, so that we could avoid repeating the notice log.

src/ngx_http_lua_pipe.c
spacewander: The pipe is `closed` when it is manually 'closed' by the users (via `proc:kill` or `proc:wait`).

spacewander:
@thibaultcha 
Updated and added test in the sister PR.

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2473 @@
...
+            return NGX_ERROR;
+        }
+
+    } else {
+        pipe->stdin_ctx->err_type = 0;
+    }
+
+    pipe_ctx = pipe->stdin_ctx;
+    if (pipe_ctx->c == NULL) {
+        *errbuf_size = ngx_snprintf(errbuf, *errbuf_size, "closed") - errbuf;
```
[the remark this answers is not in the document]
thibaultcha: We still override the errors set by `ngx_http_lua_pipe_init_ctx` here. Let's either not set the error buffer there, or follow the suggestion elaborated in my previous comment here. The latter provides more precise error messages (`"closed"` can be misleading if, e.g., we are in fact hitting the `"no memory"` branch of `ngx_http_lua_pipe_init_ctx`.

src/ngx_http_lua_pipe.c
thibaultcha: Only `proc:wait` or `__gc` seem to set `closed`. Nonetheless, I see the reason :+1:

src/ngx_http_lua_pipe.c
```diff
@@ -0,0 +1,2486 @@
...
+
+            if (dup2(err[1], STDERR_FILENO) == -1) {
+                ngx_log_error(NGX_LOG_ERR, ngx_cycle->log, ngx_errno,
+                              "lua pipe child dup2 stderr failed");
+                exit(EXIT_FAILURE);
+            }
+        }
+
+        ngx_log_debug2(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                       "lua pipe spawn process:%s pid:%P", file, getpid());
```
thibaultcha: One of the two "spawn" debug log should be removed. They currently produce a duplicated output:
```
2019/01/14 11:44:03 [debug] 9552#0: lua pipe spawn process:0000000041529F38 pid:9553 merge_stderr:0 buffer_size:4096
2019/01/14 11:44:03 [debug] 9552#0: *1 lua pipe read process:0000000041529F38 pid:9553
2019/01/14 11:44:03 [debug] 9552#0: lua pipe init pipe ctx:000000000172CFC0 fd:*12
2019/01/14 11:44:03 [debug] 9552#0: lua pipe spawn process:openssl pid:9553
```

imho, we should keep the one logged from the worker process. Not only is it more detailed, but it is also safer not to log the `file` being executed.

src/ngx_http_lua_pipe.c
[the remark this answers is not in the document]
thibaultcha: @spacewander Ping here. This is my last concern for this PR.

src/ngx_http_lua_pipe.h
```diff
@@ -80,8 +80,10 @@ typedef struct {
 } ngx_http_lua_pipe_signal_t;
 
 
+#ifndef NGX_LUA_NO_FFI_API
 void ngx_http_lua_pipe_init(void);
 ngx_int_t ngx_http_lua_pipe_add_signal_handler(ngx_cycle_t *cycle);
+#endif
```
thibaultcha: We still need these symbols to be defined when `-DNGX_LUA_NO_FFI_API` is passed to the compiler. As of now, this option would not compile. Probably best to conditionally implement those to be "NOP" when the FFI API is disabled?

src/ngx_http_lua_module.c
```diff
@@ -738,6 +742,11 @@ ngx_http_lua_init(ngx_conf_t *cf)
 
     cln->data = lmcf;
     cln->handler = ngx_http_lua_sema_mm_cleanup;
+
+#if !(NGX_WIN32)
+    ngx_http_lua_pipe_init();
```
thibaultcha: Should this also be enclosed within `#ifdef HAVE_SOCKET_CLOEXEC_PATCH`? Otherwise there is no point initializing the pipe r/b/ tree if no pipe can be spawned.

I am starting to think that we should simplify our macro dependencies... E.g. come up with `HAVE_NGX_LUA_PIPE` defined by `ngx_http_lua_pipe.h` when `HAVE_SOCKET_CLOEXEC_PATCH` and `!(NGX_WIN32)` are met (and any other conditions, present or future). We then enclose those branches in existing init code paths within a single `#ifdef HAVE_NGX_LUA_PIPE`, which reads much cleaner.

Then, inside of `ngx_http_lua_pipe.c`, we have some extra work to do to ensure that we nop the exposed symbols when `#ifdef NGX_LUA_NO_FFI_API`. And we could enclose `ngx_http_lua_ffi_pipe_spawn` within `#ifdef HAVE_NGX_LUA_PIPE`, which also reads better.

src/ngx_http_lua_initworkerby.c
```diff
@@ -65,6 +69,13 @@ ngx_http_lua_init_worker(ngx_cycle_t *cycle)
 
         return NGX_OK;
     }
+
+#ifdef HAVE_SOCKET_CLOEXEC_PATCH
+    if (ngx_http_lua_pipe_add_signal_handler(cycle) != NGX_OK) {
```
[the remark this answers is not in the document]
spacewander: This is already guarded in `!NGX_WIN32`.

thibaultcha:
@spacewander Have you seen my comments at https://github.com/openresty/lua-nginx-module/pull/1407#discussion_r247250621?

src/ngx_http_lua_pipe.c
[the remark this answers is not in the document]
spacewander: > We still override the errors set by `ngx_http_lua_pipe_init_ctx` here.

Err... I am quite confused. When `ngx_http_lua_pipe_init_ctx` sets an error, it will return `NGX_ERROR`. Then the `ngx_http_lua_ffi_pipe_proc_write` will return from there:
https://github.com/openresty/lua-nginx-module/blob/36bb55ff7cef202de6ecbbc8cfb39c9b079defd0/src/ngx_http_lua_pipe.c#L1802

So this line:
https://github.com/openresty/lua-nginx-module/blob/36bb55ff7cef202de6ecbbc8cfb39c9b079defd0/src/ngx_http_lua_pipe.c#L1811
won't be hit, and errors won't be overridden.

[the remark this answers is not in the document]
thibaultcha: Err, I guess I am quite confused myself... marking this as resolved!

thibaultcha:
Commits have been squashed and I am happy to merge them anytime.

One last comment: the naming of this API is crucial, since it will stick with OpenResty forever. Are we sure that `ngx.pipe` is the right name for a nonblocking fork-exec API? To me, pipes are just an implementation detail of this fork-exec model, and naming this API after it sorts of diminishes the appeal of the API and blurs its meaning.

We cannot use `ngx.proc` (for "processes") by fear of confusion with the [ngx.process](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/process.md) API.

Therefore, I suggest:

- `ngx.child`
- `ngx.subproc`

Both names clearly hint at a fork-exec API, without leaking underlying (minor) implementation details such as means of IPC between the worker processes and the children processes.

This is our last chance before this API sticks with OpenResty forever, so let's at least consider some alternatives?

thibaultcha:
Manually merged to master!

PR #1415, opened 2018-11-06
 tests: updated the test suite so that it can be run in parallel mode. 

thibaultcha:

To be rebased once #1414 is merged. We can now run this module's regular test suite in about just 4 minutes :)


thibaultcha:
Updated to use the `$TEST_NGINX_SERVER_ROOT` variable, which is cleaner.

PR #1433, opened 2018-11-30
Generalise shared-dict argument error message

JoeARO:
Have reworded the error message for the lua_shared_dict directive to ensure there is no confusion when there are more than 2 arguments passed into it. 
The confusion here was around the use of the word 'only' which suggests you have received less arguments than required amount, however when passed 3 or more this causes the error message to become confusing.
```
failed to run set_by_lua*: set_by_lua:3: expecting exactly two arguments, but only seen 3
```
Have replaced, 'only seen' with 'received' to ensure that the error message is more generic and doesnt lead to confusion when reading the message.

I hereby grant the copyright of the changes in this pull request to the authors of the lua-nginx-module project.

### DIFF
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -470,7 +470,7 @@ ngx_http_lua_shdict_get_helper(lua_State *L, int get_stale)
 
     if (n != 2) {
         return luaL_error(L, "expecting exactly two arguments, "
-                          "but only seen %d", n);
+                          "but received %d", n);
     }
 
     if (lua_type(L, 1) != LUA_TTABLE) {


agentzh:
@JoeARO There are many places using this wording. We'd need to be consistent with ourselves.

JoeARO:
What if I replicate the changes elsewhere too?

agentzh:
@JoeARO I don't have strong opinions about it though I have some concern about breaking backward compatibility.

@thibaultcha What's your opinion on this?

thibaultcha:
@agentzh I am in favour of this. In my opinion, some error messages we should not change, they are those about runtime errors (e.g. the `shm` API returning `"exists"` or `"no memory"`).

But argument validations are unlikely to be tested for by developers for special error handling treatment, especially when they result in Lua errors such as this one; they are mostly to catch programmer mistakes, not runtime errors. I myself have been wanting to clarify other such args validation error messages from the cosocket/ngx.pipe/shm APIs at times (and still plan to, ideally).

thibaultcha:
This function does not exist anymore since we removed the CFunction API in https://github.com/openresty/lua-nginx-module/commit/947fa0088b7a0387f43eb016436a2e4462eb5746. Therefore, I'll be closing this PR. Thanks!

Issue #92, opened 2012-03-20
log level error in ngx_http_lua_conf.c 


mei-rune:
在 src/ngx_http_lua_conf.c  文件的第 51 行， 如下
            ngx_conf_log_error(NGX_ERROR, cf, 0, "Failed to initialize Lua VM");
其中日志级别错误， 可改为

```
        ngx_conf_log_error(NGX_LOG_ERR, cf, 0, "Failed to initialize Lua VM");
```

mruse:
我遇到了同样的错误：
nginx: [error] failed to initialize Lua VM

系统版本：Darwin CmsTopMacPro 14.5.0 Darwin Kernel Version 14.5.0: Wed Jul 29 02:26:53 PDT 2015; root:xnu-2782.40.9~1/RELEASE_X86_64 x86_64
软件版本：nginx-1.8.0, luajit 2.0.4 , nginx-lua-mudle-0.9.16

同时安装的模块还有：ngx_devel_kit， nginx-rtmp-module，redis2-nginx-module

貌似0.9.16版本里src/ngx_http_lua_conf.c 替换为 src/ngx_http_lua_config.c

报该错误有2个文件;
.//src/ngx_http_lua_module.c:                               "failed to initialize Lua VM");
if (lmcf->lua == NULL) {
            ngx_conf_log_error(NGX_LOG_ERR, cf, 0,
                               "failed to initialize Lua VM");
            return NGX_ERROR;
        }

.//src/ngx_http_lua_util.h:                          "failed to initialize Lua VM");

if (!llcf->enable_code_cache && r->connection->fd != -1) {
        lmcf = ngx_http_get_module_main_conf(r, ngx_http_lua_module);

        dd("lmcf: %p", lmcf);

        L = ngx_http_lua_init_vm(lmcf->lua, lmcf->cycle, r->pool, lmcf,
                                 r->connection->log, &cln);
        if (L == NULL) {
            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
                          "failed to initialize Lua VM");
            return NULL;
        }

        if (lmcf->init_handler) {
            if (lmcf->init_handler(r->connection->log, lmcf, L) != NGX_OK) {
                /* an error happened */
                return NULL;
            }
        }

水平有限，望春哥赐教。

agentzh:
[...]

PR #1445, opened 2018-12-28
change: avoided generating Content-Type header when getting response …

spacewander:
…headers.

The previous behavior assumed that Nginx will always generate the
Content-Type header if it is missing.
However, in some cases, like the upstream response doesn't have the Content-Type
header, Nginx will not generate the Content-Type header. So generating
Content-Type header when getting response headers may create an unexpected
Content-Type header in the response.

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -505,7 +505,6 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
     ngx_table_elt_t    *header;
     ngx_http_request_t *r;
     ngx_http_lua_ctx_t *ctx;
-    ngx_int_t           rc;
     u_char             *lowcase_key = NULL;
     size_t              lowcase_key_sz = 0;
     ngx_uint_t          i;
@@ -560,17 +559,6 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
                               headers_metatable_key));
         lua_rawget(L, LUA_REGISTRYINDEX);
         lua_setmetatable(L, -2);
-    }
-
-    if (!r->headers_out.content_type.len) {
-        rc = ngx_http_lua_set_content_type(r, ctx);
-        if (rc != NGX_OK) {
-            return luaL_error(L,
-                              "failed to set default content type: %d",
-                              (int) rc);
-        }
-
-        ctx->headers_set = 1;
     }
 
 #if 1
@@ -1405,21 +1393,12 @@ ngx_http_lua_ffi_get_resp_header(ngx_http_request_t *r,
         break;
 
     case 12:
-        if (ngx_strncasecmp(key_buf, (u_char *) "Content-Type", 12) == 0) {
-            if (!r->headers_out.content_type.len) {
-                if (ngx_http_lua_set_content_type(r, ctx) != NGX_OK) {
-                    *errmsg = "failed to set default content type";
-                    return NGX_ERROR;
-                }
-
-                ctx->headers_set = 1;
-            }
-
-            if (r->headers_out.content_type.len) {
-                values[0].data = r->headers_out.content_type.data;
-                values[0].len = r->headers_out.content_type.len;
-                return 1;
-            }
+        if (ngx_strncasecmp(key_buf, (u_char *) "Content-Type", 12) == 0
+            && r->headers_out.content_type.len)
+        {
+            values[0].data = r->headers_out.content_type.data;
+            values[0].len = r->headers_out.content_type.len;
+            return 1;
         }
 
         break;
--- a/src/ngx_http_lua_headers_out.c
+++ b/src/ngx_http_lua_headers_out.c
@@ -536,7 +536,6 @@ ngx_http_lua_get_output_header(lua_State *L, ngx_http_request_t *r,
 {
     ngx_table_elt_t            *h;
     ngx_list_part_t            *part;
-    ngx_int_t                   rc;
     ngx_uint_t                  i;
     unsigned                    found;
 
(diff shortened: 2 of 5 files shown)


src/ngx_http_lua_headers.c
```diff
@@ -1406,15 +1394,6 @@ ngx_http_lua_ffi_get_resp_header(ngx_http_request_t *r,
...
         if (ngx_strncasecmp(key_buf, (u_char *) "Content-Type", 12) == 0) {
-            if (!r->headers_out.content_type.len) {
-                if (ngx_http_lua_set_content_type(r) != NGX_OK) {
-                    *errmsg = "failed to set default content type";
-                    return NGX_ERROR;
-                }
-
-                ctx->headers_set = 1;
-            }
-
```
thibaultcha: The two `if` statements should now be merged:
```c
if (ngx_strncasecmp(key_buf, (u_char *) "Content-Type", 12) == 0
    && r->headers_out.content_type.len)
{

}
```

src/ngx_http_lua_headers_out.c
```diff
@@ -552,17 +551,6 @@ ngx_http_lua_get_output_header(lua_State *L, ngx_http_request_t *r,
...
-                rc = ngx_http_lua_set_content_type(r);
-                if (rc != NGX_OK) {
-                    return luaL_error(L,
-                                      "failed to set default content type: %d",
-                                      (int) rc);
-                }
-
-                ctx->headers_set = 1;
-            }
-
```
thibaultcha: ditto: merging the two `if` statements.

t/016-resp-header.t
```diff
@@ -1501,84 +1501,77 @@ test
...
+        header_filter_by_lua_block {
+            ngx.log(ngx.WARN, "Content-Type: ", ngx.header["content-type"])
         }
     }
 --- request
-GET /set/hello.png
+GET /t
 --- response_headers
-Content-Type: image/png
+Content-Type:
```
thibaultcha: `!Content-Type` might be more explicit? It's fine either way imo, up to you.

thibaultcha:
Are we fine with breaking the behavior described in #92? Using the nginx configuration provided in the issue, the current master branch produces the following responses (which seem correct):

```
$ http :8080/set/  
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/html
Date: Fri, 04 Jan 2019 00:05:06 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked
x-hello: world

text/html

$ http :8080/noset/                      
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/html
Date: Fri, 04 Jan 2019 00:05:12 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked

text/html

```

But with this patch, the `Content-Type` header isn't even set in the first case, and is not retrieved by the ngx_lua API at all (while `lua_use_default_type` is enabled):

```
$ http :8080/set/
HTTP/1.1 200 OK
Connection: keep-alive
Date: Fri, 04 Jan 2019 00:04:14 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked
x-hello: world

nil

$ http :8080/noset/
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/html
Date: Fri, 04 Jan 2019 00:04:20 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked

nil

```

spacewander:
@thibaultcha 
I have submitted another PR to fix the problem you pointed out: https://github.com/openresty/lua-nginx-module/pull/1449

thibaultcha:
@spacewander It seems to work well :+1: However, since breaking the use case mentioned above is a regression introduced by _this_ commit, the commit in your other PR (#1449) should be squashed into this PR (in the resulting commit). This is in order to ensure atomicity. After all, the behavior described above is not failing in the master branch, only because of this commit, so it is this commit that must be updated accordingly.

Will you close your other PR and squash its patch in this PR? Thanks!

thibaultcha:
When rebasing #1449 on top of this branch, we do get the Content-Type header in the response as expected:
```
$ http :8080/set/
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/html
Date: Sat, 05 Jan 2019 03:10:52 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked
x-hello: world

nil

chasum@x1c ~ $ http :8080/noset/
HTTP/1.1 200 OK
Connection: keep-alive
Content-Type: text/html
Date: Sat, 05 Jan 2019 03:10:56 GMT
Server: nginx/1.13.6 (no pool)
Transfer-Encoding: chunked

nil

```

But we still experience a regression in that accessing `ngx.header["content-type"]` returns `nil`.

spacewander:
>  However, since breaking the use case mentioned above is a regression introduced by this commit

It's my bad that I didn't mention that the regression is introduced by the https://github.com/openresty/lua-nginx-module/commit/017e00462ae1bea96f508365726f7cec3cbcc5ca. Currently, if one has set the response header (which doesn't need to be `Content-Type`), OpenResty will not generate the default Content-Type. This is OK before that commit, because we used to generate the Content-Type once the response header is set. You could reproduce the issue via replacing `ngx.print(ngx.header["content-type"])` with `ngx.print('ok')`.

> But we still experience a regression in that accessing ngx.header["content-type"] returns nil.

I think it is not a regression, but a better way to deal with missing Content-Type header.
Previously, `ngx.header["content-type"]` returns the default type when there is not Content-Type header, because it assumed the default type will be used eventually. However, this assumption is incorrect (for instance, in the test cases in this PR).

Instead of overdoing, I think it would be better to return the real Content-Type header. When the header is missing, return `nil` instead of the default one. If you run `ngx.header["content-type"]` in `header_filter_by_lua` phase, you could get the final, correct Content-Type header.

thibaultcha:
@spacewander Gotcha. I confirmed the regression indeed comes from https://github.com/openresty/lua-nginx-module/commit/017e00462ae1bea96f508365726f7cec3cbcc5ca and pushed fixes and tweaks to your proposed patch in #1449. Thank you for clarifying.

> I think it is not a regression, but a better way to deal with missing Content-Type header.

Yes, I have to say I agree.

t/016-resp-header.t
```diff
@@ -1501,84 +1501,11 @@ test
...
-    }
---- request
-GET /set/hello.png
---- response_headers
-Content-Type: image/png
---- response_body
-image/png
-image/png
---- no_error_log
-[error]
```
thibaultcha: I think we should keep the above, but update them with the new conservative behavior we elaborated on. That is:

> Instead of overdoing, I think it would be better to return the real Content-Type header. When the header is missing, return nil instead of the default one.

Our other tests ensure that the response header has (or not) the Content-Type header, but those tests should ensure that in the _Lua-land_, the return values of those lookups is `nil`.

So we get:
```
=== TEST 70: return nil if Content-Type is not set
--- config
    location /set/ {
        default_type "image/png";
        content_by_lua_block {
            ngx.log(ngx.WARN, "Content-Type: ", h["content-type"])
            ngx.say("Content-Type: ", ngx.header["content-type"])
        }
    }
--- request
GET /set/hello.png
--- response_headers
!Content-Type
--- response_body
Content-Type: nil
--- no_error_log
[error]
--- error_log
Content-Type: nil
```

Also, this test should be moved below, where other CT-related tests are.

thibaultcha: Test added in an updated version of the patch.

PR #1460, opened 2019-01-24
implement tcpsock:getreusedtimes to get tcp infos

jichengwang:
implement tcpsock:getreusedtimes to get tcp infos

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -2845,7 +2845,8 @@ ngx_http_lua_socket_tcp_info(lua_State *L) {
    struct tcp_info info;
    int length = sizeof(struct tcp_info);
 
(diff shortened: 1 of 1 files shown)


tokers:
@jichengwang It would be better to add some tests :)

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2802,6 +2811,109 @@ ngx_http_lua_socket_tcp_close(lua_State *L)
     return 1;
 }
 
+static int
+ngx_http_lua_socket_tcp_info(lua_State *L) {
```
agentzh: I think we just need to expose a Lua API to return the connection fd. Such information gathering features are usually platform-specific (we're not just supporting Linux or *NIX systems) and are better off to some 3rd-party Lua libraries distributed to opm (or if it's indeed good enough and popular enough, we can bundle it in OpenResty by default too). It does not have to be in the core.

piersoh: There is this PR: https://github.com/openresty/lua-nginx-module/pull/1147 to do that - it would be handy to have one or the other.

PR #1473, opened 2019-02-15
bugfix: fixed a race condition in the new conn queuing mechanism.

thibaultcha:
This issue appeared in our EC2 test cluster, which compiles Nginx with
`-DNGX_LUA_USE_ASSERT` and `-DNGX_LUA_ABORT_AT_PANIC`.

The lua-resty-redis test:

    === TEST 1: github issue #108: ngx.locaiton.capture + redis.set_keepalive

in t/bugs.t [1] would produce core dumps in the check leak testing mode.

The backtrace for these core dumps was:

    #0  0x00007fd417bee277 in raise () from /lib64/libc.so.6
    #1  0x00007fd417bef968 in abort () from /lib64/libc.so.6
    #2  0x00007fd417be7096 in __assert_fail_base () from /lib64/libc.so.6
    #3  0x00007fd417be7142 in __assert_fail () from /lib64/libc.so.6
    #4  0x000000000050d227 in ngx_http_lua_socket_tcp_resume_conn_op (spool=c/ngx_http_lua_socket_tcp.c:3963
    #5  0x000000000050e51a in ngx_http_lua_socket_tcp_finalize (r=r@entry=0x5628) at ../../src/ngx_http_lua_socket_tcp.c:4195
    #6  0x000000000050e570 in ngx_http_lua_socket_tcp_cleanup (data=0x7fd419p_lua_socket_tcp.c:3755
    #7  0x0000000000463aa5 in ngx_http_free_request (r=r@entry=0xbfaec0, rc=http_request.c:3508
    ...

Which was caused by the following assertion in ngx_http_lua_socket_tcp.c
with `NGX_DEBUG`:

    #if (NGX_DEBUG)
        ngx_http_lua_assert(spool->connections >= 0);

    #else

Thanks to Mozilla's rr, a recorded session showed that
`spool->connections` was `-1`.

Here is a reproducible test case:

    local sock1 = ngx.socket.tcp()
    local sock2 = ngx.socket.tcp()

    sock1:connect()
    sock2:connect()

    sock1:setkeepalive() -- pool created, connections: 1
    sock2:setkeepalive() -- connections: 1

    sock1:connect() -- connections: 1
    sock2:connect() -- connections: 1

    sock1:close() -- connections: 0
    sock2:close() -- connections: -1
    -- ngx_http_lua_socket_tcp_resume_conn_op gets called, assertion fails

In order to avoid this race condition, we must determine whether the
socket pool exists or not, not from the
`ngx_http_lua_socket_tcp_upstream` struct, but from the Lua Registry.
This way, when thread 2's socket enters the keepalive state, it will
respect the previous call to `ngx_http_lua_socket_free_pool` (which
unset the pool from the registry).

[1]: https://github.com/openresty/lua-resty-redis/blob/master/t/bugs.t

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -4132,6 +4132,7 @@ static void
 ngx_http_lua_socket_tcp_finalize(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u)
 {
+    lua_State                      *L;
     ngx_connection_t               *c;
     ngx_http_lua_socket_pool_t     *spool;
 
@@ -4181,6 +4182,19 @@ ngx_http_lua_socket_tcp_finalize(ngx_http_request_t *r,
         u->conn_closed = 1;
 
         spool = u->socket_pool;
+        if (spool == NULL) {
+            return;
+        }
+
+        L = spool->lua_vm;
+
+        lua_pushlightuserdata(L, ngx_http_lua_lightudata_mask(socket_pool_key));
+        lua_rawget(L, LUA_REGISTRYINDEX);
+        lua_pushstring(L, (char *) spool->key);
+        lua_rawget(L, -2);
+        spool = lua_touserdata(L, -1);
+        lua_pop(L, 1);
+
         if (spool == NULL) {
             return;
         }


spacewander:
@thibaultcha 
I agree that there is a connection counting bug related with the race condition when socket pool is created/destroyed.
But I am confused by the graph. Since `sock1:setkeepalive()` is called, and the internal connection is transferred to the pool, running `ngx_http_lua_socket_tcp_finalize` will not decrease the `spool->connection`. (`u->peer.connection` is `NULL` in this situation)

spacewander:
I have a patch to fix the counting bug, which is just written last night. It could also fix the `github issue #108: ngx.locaiton.capture + redis.set_keepalive` test case.

But I am not sure whether the PR is solving the same bug. Maybe there is another counting bug.

thibaultcha:
Unfortunately I had to step away from the computer. Maybe the graph isn’t completely accurate. In my rr record, we can see, it seems, that the cleaning up of the idle cosocket triggers the finalize call.
Regardless, it is wrong imho to check for the existence of `spool` solely based on the cosocket strict member, since the real existence of a pool is determined by itself being registered in the Lua Registry anyway. This patch may need some additional work, I have been wanting to double check a few more things including the pool allocation, but unfortunately hadn’t to call it a day. Tomorrow is another day though :)

spacewander:
@thibaultcha 
I have submitted my fix: https://github.com/openresty/lua-nginx-module/pull/1475

thibaultcha:
@spacewander I have just merged your fixes from #1475, they look good! This patch also makes the two test cases you added there pass. Regarding this patch in particular, I would say that imho, it is semantically incorrect to check for the existence of `spool` solely based on `u->socket_pool`, since `ngx_http_lua_socket_free_pool` acts on the Lua registry, but won't update those existing references. This patch on top of #1475 can help harden this code path and maybe prevent future issues. What do you say?

spacewander:
> since ngx_http_lua_socket_free_pool acts on the Lua registry, but won't update those existing references

If I am right, `ngx_http_lua_socket_free_pool` will be called only when the `spool->connections == 0`. So if the counting is correct, we will not touch `u->socket_pool` after `spool` is freed.

IMHO, maybe what you observed is the result of the bugs, but not the cause?

thibaultcha:
You are correct. My reasoning was to try to avoid another similar issue with the counters from happening again. But in non-debug builds, we have code to safe-guard us from that already, so let’s close this.

thibaultcha:
Oh, also, I have started a new EC2 cluster test run on my end Since I merged your fix, so we’ll probably see good improvements on that front now. There were other related failures (not just that redis bugs.t one).

PR #1478, opened 2019-02-20
feature: added new api function ngx.resolve which allows perform host…

slimboyfat:
… name's resolution from Lua code using Nginx core's dynamic resolver.

Let me propose a new API function for Lua - ngx.resolve. This function relies on Nginx core's dynamic resolver. This approach has several advantages e.g: common query cache for Nginx core & Lua code with ability to customize caching policy by DNS TTL/custom timeout.

Please, look at code snippet in README.markdown, I hope, it can uncover the key ideas and use-cases.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -7539,6 +7539,63 @@ This feature was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.resolve
+------------------
+**syntax:** *addr, err = ngx.resolve(host, options?)*
+
+**context:** *rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;*
+
+Resolves the given host name into IP address by using Nginx core's dynamic resolver. 
+
+An optional `options` table can be fed as the second argument, which supports the options:
+
+* `ipv4`
+    consider IPv4 addresses (A Record) - `true` by default
+* `ipv6`
+    consider IPv6 addresses (AAAA Record)
+    
+If multiple A/AAAA records were retrieved by Nginx core's dynamic resolver then this method will randomly choose and return only one address.
+
+It is required to configure the [resolver](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver) directive in the `nginx.conf`.
+
+```nginx
+ http {
+     resolver 8.8.8.8;
+
+     upstream backend {
+         server 0.0.0.0;
+
+         balancer_by_lua_block {
+             local balancer = require 'ngx.balancer'
+
+             local ok, err = balancer.set_current_peer(ngx.ctx.peer_addr, ngx.ctx.peer_port)
+             if not ok then
+                 ngx.log(ngx.ERR, "failed to set the peer: ", err)
+                 ngx.exit(500)
+             end
+         }
+     }
+
+     server {
+         listen 8080;
+
+         access_by_lua_block {
+             local addr, err = ngx.resolve('google.com', { ipv4 = true })
+             if addr then
+                 ngx.ctx.peer_addr = addr
+                 ngx.ctx.peer_port = 80
+             end
+         }
+
+         location / {
+             proxy_pass http://backend;
+         }
+     }
+ }
+```
+
+[Back to TOC](#nginx-api-for-lua)
+
 ngx.get_phase
 -------------
 **syntax:** *str = ngx.get_phase()*
--- a/src/ngx_http_lua_resolve.c
+++ b/src/ngx_http_lua_resolve.c
@@ -1,0 +1,483 @@
+
+/*
+ * Copyright (C) Sergey Kharkhardin (slimboyfat)
+ */
+
+
+#ifndef DDEBUG
+#define DDEBUG 0
+#endif
+#include "ddebug.h"
+
+
+#include "ngx_http_lua_util.h"
+#include "ngx_http_lua_resolve.h"
+#include "ngx_http_lua_contentby.h"
+
+
+typedef struct ngx_http_lua_resolve_ctx_s
+        ngx_http_lua_resolve_ctx_t;
+
+
+static int ngx_http_lua_ngx_resolve(lua_State *L);
+static void ngx_http_lua_resolve_handler(ngx_resolver_ctx_t *ctx);
+static void ngx_http_lua_resolve_cleanup(void *data);
+static void ngx_http_lua_resolve_empty_handler(ngx_resolver_ctx_t *ctx);
+static ngx_int_t ngx_http_lua_resolve_resume(ngx_http_request_t *r);
+static int ngx_http_lua_resolve_select_retval(ngx_http_lua_resolve_ctx_t *u,
+    lua_State *L);
+static int ngx_http_lua_resolve_get_retval(struct sockaddr *sockaddr,
+    socklen_t socklen, lua_State *L);
+static int ngx_http_lua_resolve_set_query_option(const char *name,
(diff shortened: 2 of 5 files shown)


src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+
+    return NGX_OK;
+}
+
+
+void
+ngx_http_lua_inject_resolve_api(lua_State *L)
+{
+    lua_pushcfunction(L, ngx_http_lua_ngx_resolve);
+    lua_setfield(L, -2, "resolve");
```
agentzh: We no longer add new API functions to the global `ngx` table. The current way is to add pure Lua modules to lua-resty-core for any new API functions. See `ngx.pipe` module in that lua-resty-core library for an example.

src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
+    if (ctx == NULL) {
+        return luaL_error(L, "no ctx found");
+    }
+
+    ngx_http_lua_check_context(L, ctx, NGX_HTTP_LUA_CONTEXT_REWRITE
+                                       | NGX_HTTP_LUA_CONTEXT_ACCESS
+                                       | NGX_HTTP_LUA_CONTEXT_CONTENT);
```
agentzh: The “timer" and "ssl cert" contexts should work as well, right?

src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+
+    if (url.addrs && url.addrs[0].sockaddr) {
+        ngx_log_debug0(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+            "the given host name is address");
+
+        return ngx_http_lua_resolve_get_retval(url.addrs[0].sockaddr,
+                                               url.addrs[0].socklen, L);
+    }
+
+    u = ngx_pcalloc(r->pool, sizeof(ngx_http_lua_resolve_ctx_t));
```
agentzh: It's not a good idea to allocate in the request pool. If the user calls this `resolve()` Lua function in a tight Lua loop, for example, then there will be a horrible memory leak. Providing an OO-like interface similar to `ngx.pipe` and cosockets might make more sense.

README.markdown
```diff
@@ -7539,6 +7539,63 @@ This feature was first introduced in the `v0.5.0rc1` release.
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.resolve
```
agentzh: This file is automatically generated. See the comment at the beginning of this file for more details. You should edit the `.wiki` file under `doc/` instead and use our toolchain to update this file accordingly.

README.markdown
```diff
@@ -7539,6 +7539,63 @@ This feature was first introduced in the `v0.5.0rc1` release.
...
+                 ngx.exit(500)
+             end
+         }
+     }
+
+     server {
+         listen 8080;
+
+         access_by_lua_block {
+             local addr, err = ngx.resolve('google.com', { ipv4 = true })
```
agentzh: Passing a boolean argument directly instead of using a Lua table is much more efficient.

README.markdown
```diff
@@ -7539,6 +7539,63 @@ This feature was first introduced in the `v0.5.0rc1` release.
...
+             if addr then
+                 ngx.ctx.peer_addr = addr
+                 ngx.ctx.peer_port = 80
```
agentzh: Calling `ngx.ctx` is expensive. Better cache the table in a Lua local variable in our own examples.

src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+{
+    ngx_uint_t                     i;
+    ngx_resolver_addr_t           *addr;
+    ngx_http_upstream_resolved_t  *ur = u->resolved;
+
+    for (i = 0; i < ur->naddrs; i++) {
+        addr = &ur->addrs[i];
+        switch (addr->sockaddr->sa_family) {
+#if (NGX_HAVE_INET6)
+            case AF_INET6:
```
agentzh: Bad indentation. See the C coding style guide for more details:

https://openresty.org/en/c-coding-style-guide.html

src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+                if (u->ipv6) {
+                    return ngx_http_lua_resolve_get_retval(addr->sockaddr,
+                                                           addr->socklen, L);
+                }
+                break;
+#endif
+            default: /* AF_INET */
```
agentzh: I'm not sure if we need to use `switch` for a 2-branch case. Seems overkill to me.

src/ngx_http_lua_resolve.c
```diff
@@ -0,0 +1,483 @@
...
+}
+
+
+static int
+ngx_http_lua_resolve_set_query_option(const char *name, unsigned *option, lua_State *L)
+{
+    lua_getfield(L, 2, name);
+
+    switch (lua_type(L, -1)) {
+        case LUA_TNIL:
```
agentzh: Ditto. Bad indentation.

src/ngx_http_lua_resolve.h
```diff
@@ -0,0 +1,18 @@
+
+/*
+ * Copyright (C) Sergey Kharkhardin (slimboyfat)
```
agentzh: You need to grant the copyright to the authors of this project before this can be merged. Thanks for your understanding.

src/ngx_http_lua_resolve.c
slimboyfat: Got it, I have completely rewritten my initial implementation.

src/ngx_http_lua_resolve.c
slimboyfat: I didn't cover these cases by tests, but, it seems it should work.

src/ngx_http_lua_resolve.c
slimboyfat: Thanks for that point. I completely reorganized memory management approach in my new implementation.

README.markdown
slimboyfat: Now, as I have reorganized my implementation, I put resolver.md right near resolver.lua in lua-resty-core module (just like pipe.md, hope this is right).

src/ngx_http_lua_resolve.h
slimboyfat: NP

src/ngx_http_lua_resolver.c
```diff
@@ -0,0 +1,367 @@
...
+    unsigned                         ipv6:1;
+} ngx_http_lua_resolver_ctx_t;
+
+
+static void ngx_http_lua_resolve_handler(ngx_resolver_ctx_t *rctx);
+static void ngx_http_lua_resolve_cleanup(void *data);
+static void ngx_http_lua_resolve_empty_handler(ngx_resolver_ctx_t *ctx);
+static ngx_int_t ngx_http_lua_resolve_resume(ngx_http_request_t *r);
+static int ngx_http_lua_resolve_retval(ngx_http_lua_resolver_ctx_t *ctx,
+                                       lua_State *L);
```
agentzh: Bad indentation. See the guide for details:

https://openresty.org/en/c-coding-style-guide.html

Also, make sure you run the `ngx-releng` tool mentioned in the guide as well. That tool should catch this.

PR #1479, opened 2019-02-21
feature: implemented the lindex method for list type share dict value

tokers:
The `lindex` method returns the element at index `index` in the shared
dict list. Just like the `LINDEX` of Redis.


src/ngx_http_lua_shdict.c
```diff
@@ -375,6 +376,9 @@ ngx_http_lua_inject_shdict_api(ngx_http_lua_main_conf_t *lmcf, lua_State *L)
         lua_pushcfunction(L, ngx_http_lua_shdict_llen);
         lua_setfield(L, -2, "llen");
 
+        lua_pushcfunction(L, ngx_http_lua_shdict_lindex);
```
agentzh: I think we should only do the pure FFI implementation (through lua-resty-core) for such new shdict methods. We're retiring the CFunction implementation very soon. They'll be gone.

tokers: Okay, I will rewrite this.

tokers:
@agentzh @membphis Rewrite it with the FFI method.

tokers:
The sibling PR is https://github.com/openresty/lua-resty-core/pull/231.

PR #1486, opened 2019-02-27
optimize: avoided using strings as code cache keys.

thibaultcha:
This optimization takes advantage of the `luaL_ref()` C API to generate
unique references for Lua buffers (inline and file). This avoids runtime
calls to `lj_str_new()` while invoking Lua handler.

By doing so, we observe reduced CPU utilization and increased throughput
(~2% to ~5% in preliminary tests) in a simplistic hello world handler.
Other benefits of this patch include the simplification of the code
cache keys generation and reduced memory consumption.

In this implementation, a `LUA_REFNIL` code cache reference means that the
associated chunk has not been cached yet. A `LUA_NOREF` code cache
reference applies to Lua file paths and means that this particular file
path has nginx variables and should thus generate its own cache key
(identical to the current implementation).

This patch also fixes an isue with `set_by_lua_file` handlers if those
were specified with nginx variables. We added a new test case for this
behavior which fails on the master branch of this repository (v0.10.14),
but passes with this patch.

### DIFF
--- a/src/ngx_http_lua_accessby.c
+++ b/src/ngx_http_lua_accessby.c
@@ -179,7 +179,7 @@ ngx_http_lua_access_handler_inline(ngx_http_request_t *r)
     rc = ngx_http_lua_cache_loadbuffer(r->connection->log, L,
                                        llcf->access_src.value.data,
                                        llcf->access_src.value.len,
-                                       llcf->access_src_key,
+                                       &llcf->access_src_ref,
                                        (const char *) llcf->access_chunkname);
 
     if (rc != NGX_OK) {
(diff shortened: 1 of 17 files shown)


agentzh:
@spacewander @dndx Will you please help review this PR? Thanks!

thibaultcha:
@spacewander @dndx Ping :) Will you help me review this patch? I gave it more review rounds lately and it looks all good on my side. Keep in mind the difference between `LUA_REFNIL` and `LUA_NOREF`. It's a very healthy change imho (much simpler, more efficient, and and even fixes a bug with `set_by_lua`). Cheers.

thibaultcha:
Note: an alternative approach to this PR (with a significant behavior difference) was opened at https://github.com/openresty/lua-nginx-module/pull/1543

thibaultcha:
Merged #1543 instead.

PR #1489, opened 2019-03-04
New API for sending ffi buffer in tcpsock without casting to string

dqminh:
This expose a way to set the tcp upstream socket's request buffer
manually, so we can call write later. The reason this is splitted into 2
functions is because i couldn't find a way to send cdata over in a
cfunction sanely.

Here is roughly how you can use them

```
local ffi = require "ffi"
local base = require "resty.core.base"

local C = ffi.C
local error = error
local errmsg = base.get_errmsg_ptr()
local FFI_OK = base.FFI_OK

ffi.cdef[[
int ngx_http_lua_socket_tcp_ffi_set_buf(ngx_http_request_t *r, void *u, void *cdata_ctx, size_t len, char **err);
]]

local function check_tcp(tcp)
  if not tcp or type(tcp) ~= "table" then
    return error("bad \"tcp\" argument")
  end
  tcp = tcp[1]
  if type(tcp) ~= "userdata" then
    return error("bad \"tcp\" argument")
  end
  return tcp
end

local function ffi_set_buffer(tcp, buffer, len)
  tcp = check_tcp(tcp)
  local r = base.get_request()
  if not r then
    return error("no request found")
  end
  local rc = C.ngx_http_lua_socket_tcp_ffi_set_buf(r, tcp, buffer, len, errmsg)
  if rc ~= FFI_OK then
    return nil, ffi.string(errmsg[0])
  end
  return true
end

local sock = ngx.socket.tcp()
sock:connect("unix:/tmp/go.sock")
sock:settimeout(1000)
ffi_set_buffer(sock, data, len)
sock:ffi_send()
```

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -18,6 +18,11 @@
 #include "ngx_http_lua_contentby.h"
 #include "ngx_http_lua_probe.h"
 
+#ifndef NGX_LUA_NO_FFI_API
+#ifndef LUA_TCDATA
+#define LUA_TCDATA 10
+#endif
+#endif
 
 static int ngx_http_lua_socket_tcp(lua_State *L);
 static int ngx_http_lua_socket_tcp_connect(lua_State *L);
@@ -27,6 +32,11 @@ static int ngx_http_lua_socket_tcp_sslhandshake(lua_State *L);
 static int ngx_http_lua_socket_tcp_receive(lua_State *L);
 static int ngx_http_lua_socket_tcp_receiveany(lua_State *L);
 static int ngx_http_lua_socket_tcp_send(lua_State *L);
+
+#ifndef NGX_LUA_NO_FFI_API
+static int ngx_http_lua_socket_tcp_ffi_send(lua_State *L);
+#endif
+
 static int ngx_http_lua_socket_tcp_close(lua_State *L);
 static int ngx_http_lua_socket_tcp_setoption(lua_State *L);
 static int ngx_http_lua_socket_tcp_settimeout(lua_State *L);
@@ -208,7 +218,6 @@ enum {
         return 2;                                                            \
     }
 
-
 static char ngx_http_lua_req_socket_metatable_key;
 static char ngx_http_lua_raw_req_socket_metatable_key;
 static char ngx_http_lua_tcp_socket_metatable_key;
@@ -295,6 +304,13 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
     lua_pushcfunction(L, ngx_http_lua_socket_tcp_settimeouts);
     lua_setfield(L, -2, "settimeouts"); /* ngx socket mt */
 
+#ifndef NGX_LUA_NO_FFI_API
+
+    lua_pushcfunction(L, ngx_http_lua_socket_tcp_ffi_send);
+    lua_setfield(L, -2, "ffi_send");
+
+#endif /* NGX_LUA_NO_FFI_API */
+
     lua_pushvalue(L, -1);
     lua_setfield(L, -2, "__index");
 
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_socket_tcp.c
```diff
@@ -27,6 +27,11 @@ static int ngx_http_lua_socket_tcp_sslhandshake(lua_State *L);
 static int ngx_http_lua_socket_tcp_receive(lua_State *L);
 static int ngx_http_lua_socket_tcp_receiveany(lua_State *L);
 static int ngx_http_lua_socket_tcp_send(lua_State *L);
+
+// send the exising buffer out non-blocking way.
+// buffer should be set using ngx_http_lua_socket_tcp_ffi_set_buf
```
agentzh: Do not use C++-style comments in this module. Please see the C coding style guide for more details:

https://openresty.org/en/c-coding-style-guide.html

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2963,6 +2981,225 @@ ngx_http_lua_socket_tcp_send(lua_State *L)
     return lua_yield(L, 0);
 }
```
agentzh: The nginx coding style requires the use of 2 blank lines to separate function definitions. See the guide mentioned above.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2963,6 +2981,225 @@ ngx_http_lua_socket_tcp_send(lua_State *L)
...
+        return NGX_ERROR;
+    }
+
+    if (u->write_waiting) {
+        *err = "socket busy writing";
+        return NGX_ERROR;
+    }
+
+    if (u->raw_downstream
+        && (r->connection->buffered & NGX_HTTP_LOWLEVEL_BUFFERED)) {
```
agentzh: When the if condition is in multiple lines, the `{` should be on a separate line by itself. See the guide mentioned above.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2963,6 +2981,225 @@ ngx_http_lua_socket_tcp_send(lua_State *L)
...
+        return NGX_ERROR;
+    }
+
+    b = cl->buf;
+    b->last = ngx_copy(b->last, cdata_ctx, len);
+    u->request_bufs = cl;
+    u->request_len = len;
+    return NGX_OK;
+}
+
```

src/ngx_http_lua_socket_tcp.c
```diff
@@ -295,6 +299,13 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
     lua_pushcfunction(L, ngx_http_lua_socket_tcp_settimeouts);
     lua_setfield(L, -2, "settimeouts"); /* ngx socket mt */
 
+#ifndef NGX_LUA_NO_FFI_API
+
```
agentzh: We'd better remove this blank line.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -295,6 +299,13 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
...
+    lua_pushcfunction(L, ngx_http_lua_socket_tcp_ffi_send);
+    lua_setfield(L, -2, "ffi_send");
+
```
agentzh: Ditto. Please address other similar places.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -295,6 +299,13 @@ ngx_http_lua_inject_socket_tcp_api(ngx_log_t *log, lua_State *L)
     lua_pushcfunction(L, ngx_http_lua_socket_tcp_settimeouts);
     lua_setfield(L, -2, "settimeouts"); /* ngx socket mt */
 
+#ifndef NGX_LUA_NO_FFI_API
+
+    lua_pushcfunction(L, ngx_http_lua_socket_tcp_ffi_send);
```
agentzh: As I mentioned, we should not implement this as a C function.

zhuizhuhaomeng:
Rewrite this PR in https://github.com/openresty/lua-nginx-module/pull/2294.

PR #1492, opened 2019-03-09
tests: removed the broken iscribblet.org from the test suite.

spacewander:
This change makes the CI pass again.


agentzh:
@spacewander Reducing the number of concurrent threads compromises the test itself. I've changed it to openresty.com instead in my own commit just pushed.

PR #1501, opened 2019-04-03
feat: implemented the lua_load_resty_core directive.

thibaultcha:

```
lua_load_resty_core [on|off]; # on by default
```

Sister PRs:
- https://github.com/openresty/lua-resty-core/pull/238
- https://github.com/openresty/meta-lua-nginx-module/pull/68
- https://github.com/openresty/stream-lua-nginx-module/pull/159

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1004,6 +1004,7 @@ See Also
 Directives
 ==========
 
+* [lua_load_resty_core](#lua_load_resty_core)
 * [lua_capture_error_log](#lua_capture_error_log)
 * [lua_use_default_type](#lua_use_default_type)
 * [lua_malloc_trim](#lua_malloc_trim)
@@ -1078,6 +1079,38 @@ how the result will be used. Below is a diagram showing the order in which direc
 ![Lua Nginx Modules Directives](https://cloud.githubusercontent.com/assets/2137369/15272097/77d1c09e-1a37-11e6-97ef-d9767035fc3e.png)
 
 [Back to TOC](#table-of-contents)
+
+lua_load_resty_core
+-------------------
+
+**syntax:** *lua_load_resty_core on|off*
+
+**default:** *lua_load_resty_core on*
+
+**context:** *http*
+
+Controls whether the `resty.core` module (from
+[lua-resty-core](https://github.com/openresty/lua-resty-core)) should be loaded
+or not. When enabled, this directive is equivalent to executing the following
+when the Lua VM is created:
+
+```lua
+
+ require "resty.core"
+```
+
+Note that usage of the `resty.core` module is recommended, as its
+FFI implementation is both faster, safer, and more complete than the Lua C API
+of the ngx_lua module.
+
+It must also be noted that the Lua C API of the ngx_lua module will eventually
+be removed, and usage of the FFI-based API (i.e. the `resty.core`
+module) will become mandatory. This directive only aims at providing a
+temporary backwards-compatibility mode in case of edge-cases.
+
+This directive was first introduced in the `v0.10.15` release.
+
+[Back to TOC](#directives)
 
 lua_capture_error_log
 ---------------------
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -147,6 +147,13 @@ static int ngx_http_lua_get_raw_phase_context(lua_State *L);
 #define LUA_PATH_SEP ";"
 #endif
 
+
+#if !defined(LUA_DEFAULT_PATH) && (NGX_DEBUG)
+#define LUA_DEFAULT_PATH "../lua-resty-core/lib/?.lua;"                      \
+                         "../lua-resty-lrucache/lib/?.lua"
+#endif
+
+
 #define AUX_MARK "\1"
 
 
@@ -3813,6 +3820,7 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
     ngx_pool_t *pool, ngx_http_lua_main_conf_t *lmcf, ngx_log_t *log,
     ngx_pool_cleanup_t **pcln)
 {
+    int                              rc;
     lua_State                       *L;
     ngx_uint_t                       i;
     ngx_pool_cleanup_t              *cln;
@@ -3878,6 +3886,21 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
         }
 
         lua_pop(L, 2);
+    }
+
+    if (lmcf->load_resty_core) {
+        lua_getglobal(L, "require");
+        lua_pushstring(L, "resty.core");
+
+        rc = lua_pcall(L, 1, 1, 0);
+        if (rc != 0) {
+            ngx_log_error(NGX_LOG_ERR, log, 0,
+                          "lua_load_resty_core failed to load the resty.core "
+                          "module from https://github.com/openresty/lua-resty"
+                          "-core; ensure you are using an OpenResty release "
+                          "from https://openresty.org/en/download.html "
+                          "(rc: %i, reason: %s)", rc, lua_tostring(L, -1));
+        }
     }
 
     return L;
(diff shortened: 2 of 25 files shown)


src/ngx_http_lua_util.c
```diff
@@ -3813,6 +3820,7 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
     ngx_pool_t *pool, ngx_http_lua_main_conf_t *lmcf, ngx_log_t *log,
     ngx_pool_cleanup_t **pcln)
 {
+    int                             rc;
```
agentzh: Alignment issue. Better use `ngx-releng` to check coding style, always.

src/ngx_http_lua_util.c
```diff
@@ -3880,6 +3888,18 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
...
 
+    if (lmcf->load_resty_core) {
+        lua_getglobal(L, "require");
+        lua_pushstring(L, "resty.core");
+
+        rc = lua_pcall(L, 1, 1, 0);
+        if (rc != 0) {
+            ngx_log_error(NGX_LOG_ERR, log, 0,
+                          "lua_load_resty_core failed to load resty.core "
+                          "(%i): %s", rc, lua_tostring(L, -1));
```
agentzh: Better give more details here like recommending the use of OpenResty (with link) and the URL of the lua-resty-core library.

src/ngx_http_lua_util.c
thibaultcha: Fixed - sometimes I forget :)

src/ngx_http_lua_util.c
thibaultcha: Precisions added in the warning log

README.markdown
```diff
@@ -1079,6 +1080,31 @@ how the result will be used. Below is a diagram showing the order in which direc
...
+should be loaded or not. When enabled, this directive is equivalent to executing the following when the Lua VM is
+created:
+
+```lua
+
+ require "resty.core"
+```
+
+Note that usage of the `resty.core` module is recommended, as its FFI implementation is both faster and
+more complete than the Lua C API of the ngx_lua module.
```
agentzh: And also "safer". Thinking about the "dead lock" risk in the shdict API based on CFunction.

Also, we can add a caveat that we will remove the old CFunction-based Lua API implementation in the very near future and the use of lua-resty-core will be mandated.

thibaultcha: Right, I added both notes to the documentation.

thibaultcha:
Manually merged

chobits:
Hi @thibaultcha ,

The default `on` value of `lua_load_resty_core` directive makes starting nginx failed without `lua-resty-core`.

See this issue: https://github.com/openresty/lua-nginx-module/issues/1509#issuecomment-485456796

thibaultcha:
@chobits Indeed. lua-resty-core is to become a dependency of this module. For now, you can still turn this directive off. In an upcoming release, all CFunction APIs will be removed and lua-resty-core will become mandatory.

chobits:
@thibaultcha OK, Let me try lua-resty-core!

src/ngx_http_lua_util.c
```diff
@@ -3880,6 +3888,21 @@ ngx_http_lua_init_vm(lua_State *parent_vm, ngx_cycle_t *cycle,
...
+        lua_pushstring(L, "resty.core");
+
+        rc = lua_pcall(L, 1, 1, 0);
+        if (rc != 0) {
+            ngx_log_error(NGX_LOG_ERR, log, 0,
+                          "lua_load_resty_core failed to load the resty.core "
+                          "module from https://github.com/openresty/lua-resty"
+                          "-core; ensure you are using an OpenResty release "
+                          "from https://openresty.org/en/download.html "
+                          "(rc: %i, reason: %s)", rc, lua_tostring(L, -1));
```
ElvinEfendi: Does this mean lua-nginx-module can not be used outside of OpenResty bundle anymore? In one of our projects we compile lua-nginx-module manually and also include lua-resty-core and set lua package path correctly. Since this PR Nginx fails when starting with this error. When I turn off `lua_load_resty_core` and do just `require("resty.core")` it works as expected.

agentzh: @ElvinEfendi It still can, just a little bit harder as you found out. If you want to add Lua library search paths to your own nginx build, then just override the value of the macro `LUA_DEFAULT_PATH` to include your own installation paths of lua-resty-core (and lua-resty-lrucache). The OpenResty distribution's build system does exactly that.

agentzh: Hmm, it should also work if the user specify the `lua_package_path` and `lua_package_cpath` directives. @thibaultcha The `lua_code_resty_core` directive does not run after these search path nginx directives?

ElvinEfendi: @agentzh the project I'm talking about is https://github.com/kubernetes/ingress-nginx/, and we are actually considering to use OpenResty as base image instead of Nginx.

--

I replicated CI image in this repository and renamed `lua-resty-core` and `lua-resty-lrucache` folders to something else. I then edited `t/161-load-resty-core.t` added `ONLY` to the first test and also added `lua_package_path "/lua-resty-corexxx/lib/?.lua;/lua-resty-lrucachexxx/lib/?.lua;;";` using `--- http_config`.
`/lua-resty-corexxx` and `/lua-resty-lrucachexxx` are what I renamed `lua-resty-core` and `lua-resty-lrucache` folders to. Then running test it succeeds. So that tells me it respects `lua_package_path` in this case. Not sure yet why it does not work in https://github.com/kubernetes/ingress-nginx though. The only difference between the Nginx we use in ingress-nginx and 1.15.8 (also the version I used in the above test: https://github.com/ElvinEfendi/lua-nginx-module/pull/1/files#diff-3254677a7917c6c01f55212f86c57fbfR7) that OpenResty uses is that we use Nginx `1.17.0`.

agentzh: @ElvinEfendi Then it's not something on our side? For OpenResty, then it already includes lua-resty-core or lua-resty-lrucache in its default search paths.

agentzh: @ElvinEfendi For us to support you, you need to prepare a minimal and self-contained example. We are not maintaining that ingress-nginx project anyway, not to mention your own setup.

ElvinEfendi: @agentzh all I wanted to clarify was "Does this mean lua-nginx-module can not be used outside of OpenResty bundle anymore?" :) Because the error message implies that.

I co-maintain ingress-nginx project - will look into this further on our side to see what's going on and will also potentially switch to OpenResty as base image for that project if the dependencies we need play well with OpenResty bundle.

Thanks for taking time to look into this though!

agentzh: Well, that's a recommendation. Not a requirement.

thibaultcha: @ElvinEfendi The `lua_load_resty_core` directive respects the `LUA_PATH` environment variable and the `lua_package_path` directive. In fact, I just pushed a new test asserting this behavior: https://github.com/openresty/lua-nginx-module/commit/4102b82cd0971c93663fe89eabd75b73741c4213

Make sure that you also clone [lua-resty-lrucache](https://github.com/openresty/lua-resty-lrucache) which is itself a dependency of lua-resty-core.
Or maybe you set your Lua path in Lua-land (e.g. `package.path = "...;" .. package.path`)? In which case, that'd be too late for this directive to honor the change.

Providing a minimally reproducible example with error logs may be useful here, but again as @agentzh said, this isn't really an issue this module.

This module can still be used independently but the margin for error is high for most users, hence the warnings and other logs (e.g. lua-resty-core, LuaJIT's fork, etc...) asking to prefer the formal OpenResty releases.

yjfvictor: Where can I set `package.path`?

PR #1514, opened 2019-04-25
feature: shared ngx.ctx among SSL_* phases and the following phases.

spacewander:
TODO: update documentation & support the non-FFI version of API.


spacewander:
This is the sister PR of https://github.com/openresty/lua-resty-core/pull/243

agentzh:
@spacewander We don't need the non-FFI version of the API. Actually we are retiring all the non-FFI variants of the existing API functions which already have an FFI version.

src/ngx_http_lua_ctx.c
```diff
@@ -145,19 +151,67 @@ ngx_http_lua_ffi_get_ctx_ref(ngx_http_request_t *r)
 }
 
 
+int
+ngx_http_lua_ffi_get_ssl_ctx_ref(ngx_http_request_t *r, int *in_ssl_phase)
```
dndx: I might be missing something, but it seems that we can change the existing `ngx_http_lua_ffi_get_ctx_ref` function to look for `ssl_ctx-> ssl_ctx` as well and return `int *need_clone` based on the current phase and where ref was located instead? That will save one FFI call and introducing new function.

spacewander:
Updated.

doc/HttpLuaModule.wiki
```diff
@@ -3069,6 +3069,12 @@ Then <code>GET /orig</code> will give
 
 rather than the original <code>"hello"</code> value.
 
+Because HTTP request is created after SSL handshake, the <code>ngx.ctx</code> created
+in [[#ssl_certificate_by_lua|ssl_certificate_by_lua*]], [[#ssl_session_store_by_lua|ssl_session_store_by_lua*]] and [[#ssl_session_fetch_by_lua|ssl_session_fetch_by_lua*]]
+is not available in the following phases like [[#rewrite_by_lua|rewrite_by_lua*]].
+Since <code>dev</code>, the <code>ngx.ctx</code> created during a SSL handshake
```
doujiang24, line 3075: style: break a new line may be better?

doc/HttpLuaModule.wiki
```diff
@@ -3069,6 +3069,12 @@ Then <code>GET /orig</code> will give
...
+will be inherited by the requests which share the same TCP connection established by the handshake.
```
doujiang24, line 3076: better clarify the inherited value from `ngx.ctx` is read-only.

doc/HttpLuaModule.wiki
spacewander: @doujiang24 
Updated.

doc/HttpLuaModule.wiki
```diff
@@ -3069,6 +3069,14 @@ Then <code>GET /orig</code> will give
 
 rather than the original <code>"hello"</code> value.
 
+Because HTTP request is created after SSL handshake, the <code>ngx.ctx</code> created
+in [[#ssl_certificate_by_lua|ssl_certificate_by_lua*]], [[#ssl_session_store_by_lua|ssl_session_store_by_lua*]] and [[#ssl_session_fetch_by_lua|ssl_session_fetch_by_lua*]]
+is not available in the following phases like [[#rewrite_by_lua|rewrite_by_lua*]].
+
+Since <code>dev</code>, the <code>ngx.ctx</code> created during a SSL handshake
+will be inherited by the requests which share the same TCP connection established by the handshake.
+Note that the inherited value from <code>ngx.ctx</code> is read-only.
```
doujiang24, line 3078:
```suggestion
Note that overwrite values in <code>ngx.ctx</code> in the http request phases (like `rewrite_by_lua*`) will only take affect in the current http request.
```

doujiang24: and add the following case will be much clearer.

for example:

```
ssl_certificate_by_lua_by_block {
    ngx.ctx.foo = 1  -- it will always be 1 in all of the http requests over this SSL connection.
}

location /foo-1 {
    rewrite_by_lua_block {
        ngx.ctx.foo = 2 -- only take affects in current http request
    }
}

location /foo-2 {
    rewrite_by_lua_block {
        ngx.say(ngx.ctx.foo) -- will alway be 1
    }
}
```

spacewander: @doujiang24 
Done. Test case is added in https://github.com/openresty/lua-resty-core/pull/243

src/ngx_http_lua_ctx.c
```diff
@@ -66,32 +70,71 @@ ngx_http_lua_ngx_set_ctx_helper(lua_State *L, ngx_http_request_t *r,
...
         return NGX_HTTP_LUA_FFI_NO_REQ_CTX;
     }
 
-    return ctx->ctx_ref;
+    if (ctx->ctx_ref >= 0 || in_ssl_phase == NULL) {
+        return ctx->ctx_ref;
+    }
+
+    *in_ssl_phase = ctx->context & (NGX_HTTP_LUA_CONTEXT_SSL_CERT
+                    | NGX_HTTP_LUA_CONTEXT_SSL_SESS_FETCH
```
agentzh, line 89: This has indentation problem caught by `ngx-releng`:

```
src/ngx_http_lua_ctx.c:
incorrect front spaces, unclosed bracket
89:                     | NGX_HTTP_LUA_CONTEXT_SSL_SESS_FETCH
```

spacewander: Solved.

PR #1525, opened 2019-05-10
Clarify ngx.req.set_header usage for proxies

haizaar:
It's not obvious that `ngx.req.set_header` acts similar `proxy_set_header` for `proxy_pass` scenarios. Clarifying that.

Discussion: https://groups.google.com/forum/#!topic/openresty-en/CDLGj69PDAI

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4788,6 +4788,8 @@ ngx.req.set_header
 
 Set the current request's request header named `header_name` to value `header_value`, overriding any existing ones.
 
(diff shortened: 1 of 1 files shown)


zhuizhuhaomeng:
merged with some modification.

```
It is not a Lua's equivalent of nginx `proxy_set_header` directive (same is true about [ngx.req.clear_header](#ngxreqclear_header)). `proxy_set_header` only affects the upstream request while `ngx.req.set_header` change the incoming request. Record the http headers in the access log file will show the difference. But you still can use it as an alternative of nginx `proxy_set_header` directive as long as you know the difference.
```

PR #1530, opened 2019-05-22
doc: fix thinko in tcpsocket:receiveany()

lukego:
This is just a really trivial doc change :-)

Originally I wanted to change the documentation of `tcpsocket:receiveany()` to specify that *some* data is always returned, even if none was available at the time the call is made. (Right?) But I punted on that because maybe this is supposed to be implied when saying that the call is *synchronous.* (It wasn't obvious to me so I had to peek at the C code fwiw.)

The one-liner is what was then left on that branch :grin: 

----

### DIFF
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -6178,7 +6178,7 @@ Timeout for the reading operation is controlled by the [[#lua_socket_read_timeou
 
 <geshi lang="lua">
     sock:settimeouts(1000, 1000, 1000)  -- one second timeout for connect/read/write
-    local data, err = sock:receiveany(10 * 1024 * 1024) -- read any data, at most 10K
+    local data, err = sock:receiveany(10 * 1024) -- read any data, at most 10K
     if not data then
         ngx.say("failed to read any data: ", err)
         return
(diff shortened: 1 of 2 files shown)


spacewander:
@lukego 
Will you update the README via `wiki2markdown.pl doc/HttpLuaModule.wiki >README.markdow`?
The `wiki2markdown.pl` can be found in https://github.com/openresty/openresty-devel-utils.

lukego:
Sure, done.

doc/HttpLuaModule.wiki
```diff
@@ -6178,7 +6178,7 @@ Timeout for the reading operation is controlled by the [[#lua_socket_read_timeou
 
 <geshi lang="lua">
     sock:settimeouts(1000, 1000, 1000)  -- one second timeout for connect/read/write
-    local data, err = sock:receiveany(10 * 1024 * 1024) -- read any data, at most 10K
+    local data, err = sock:receiveany(10 * 1024 * 1024) -- read any data, at most 10MB
```
thibaultcha: Imho, `10 * 1024` provides a more reasonable example for this API. I suggest that we update the bytes formula to `10 * 1024` or `10 * 2^10` to read 10KiB.

lukego: Hi @thibaultcha :). Sure, done. I took the liberty of a squash-rebase at the same time.

thibaultcha: @lukego Thanks!

thibaultcha:
Will merge after #1531 which fixes the CI build.

PR #1536, opened 2019-05-31
feature: logged function location when failed to run a timer.

spacewander:

### DIFF
--- a/src/ngx_http_lua_timer.c
+++ b/src/ngx_http_lua_timer.c
@@ -727,26 +727,26 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
 
 failed:
 
-    if (tctx.co_ref && tctx.co) {
-        /* co stack: func [args] */
-        lua_getinfo(tctx.co, ">Sf", &ar);
+    ngx_http_lua_assert(tctx.co_ref && tctx.co);
 
-        source = ar.source;
+    /* co stack: func [args] */
+    lua_getinfo(tctx.co, ">Sf", &ar);
 
-        if (source == NULL) {
-            source = "(unknown)";
-        }
+    source = ar.source;
 
-        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
-                      "lua run timer with function defined in %s:%d"
-                      " failed: %s", source, ar.linedefined, errmsg);
-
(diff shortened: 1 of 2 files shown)


src/ngx_http_lua_timer.c
```diff
@@ -559,6 +559,13 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
     ngx_http_lua_main_conf_t        *lmcf;
     ngx_http_core_loc_conf_t        *clcf;
 
+    lua_Debug                ar;
+    u_char                  *p;
+    u_char                   errbuf[64];
+    int                      errbuf_size = 64;
```
agentzh: Better avoid repeating this magical number 64, which can easily get out of sync. Better use a local anonymous `enum {}` to give this constant a name?

This way we can also save the `errbuf_size` auto variable.

Also, I don't think `64` is large enough. Better be 128 or even larger? Just to be safe.

src/ngx_http_lua_timer.c
```diff
@@ -579,17 +586,18 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
     }
 
     if (lmcf->running_timers >= lmcf->max_running_timers) {
-        ngx_log_error(NGX_LOG_ALERT, ngx_cycle->log, 0,
-                      "%i lua_max_running_timers are not enough",
-                      lmcf->max_running_timers);
+        p = ngx_snprintf(errbuf, errbuf_size,
+                         "%i lua_max_running_timers are not enough",
+                         lmcf->max_running_timers);
+        *p = '\0';
```
agentzh: This may overflow the buffer? Better pass `errbuf_size - 1` to the `ngx_snprintf()` call?

src/ngx_http_lua_timer.c
```diff
@@ -724,6 +725,19 @@ ngx_http_lua_timer_handler(ngx_event_t *ev)
 
 failed:
 
+    /* co stack: func [args] */
+    lua_getinfo(tctx.co, ">Sf", &ar);
```
agentzh: We should check if `tctx.co_ref` and `tctx.co` are defined, just like below on line 741. Just in case.

spacewander:
@agentzh 
Updated.

thibaultcha:
Merged with the appropriate fixes. Thanks!

PR #1537, opened 2019-05-31
bugfix: LuaJIT sets currentline to -1 if it is unavailable.

spacewander:

### DIFF
--- a/src/ngx_http_lua_log.c
+++ b/src/ngx_http_lua_log.c
@@ -194,7 +194,7 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
     *p++ = ':';
 
     p = ngx_snprintf(p, NGX_INT_T_LEN, "%d",
-                     ar.currentline ? ar.currentline : ar.linedefined);
+                     ar.currentline > 0 ? ar.currentline : ar.linedefined);
 
     *p++ = ':'; *p++ = ' ';
 


src/ngx_http_lua_log.c
```diff
@@ -194,7 +194,7 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
     *p++ = ':';
 
     p = ngx_snprintf(p, NGX_INT_T_LEN, "%d",
-                     ar.currentline ? ar.currentline : ar.linedefined);
+                     ar.currentline != -1 ? ar.currentline : ar.linedefined);
```
agentzh: Better be `>= 0` instead.

spacewander:
@agentzh 
Updated.

src/ngx_http_lua_log.c
```diff
@@ -194,7 +194,7 @@ log_wrapper(ngx_log_t *log, const char *ident, ngx_uint_t level,
     *p++ = ':';
 
     p = ngx_snprintf(p, NGX_INT_T_LEN, "%d",
-                     ar.currentline ? ar.currentline : ar.linedefined);
+                     ar.currentline >= 0 ? ar.currentline : ar.linedefined);
```
agentzh: Sorry, I meant 0 line number makes little sense, so we should use `>` here instead? Just like the previous behavior?

spacewander: @agentzh 
Updated again.

Issue #1539, opened 2019-06-06
ssl_session_fetch|store issues with TLSv1.3


yongjianchn:
@thibaultcha @spacewander  

with https://github.com/openresty/lua-nginx-module/pull/1503, I failed to reuse tls session with ssl_session_fetch\store_by_lua on TLSv1.3.

I reviewed the code, and I think SSL_client_hello_get0_session_id() which called in ngx_http_lua_ssl_client_hello_handler() will get the wrong session id on TLSv1.3, while the real session id is stored in PSK and ngx_http_lua_ssl_sess_fetch_handler() can get.

yongjianchn:
-- ssl_session_fetch_by_lua_file

local ssl_sess = require "ngx.ssl.session"

local local_dict = ngx.shared.ssl_session

local function ssl_session_fetch()
    local sess_id, err = ssl_sess.get_session_id()
    if not sess_id then 
        ngx.log(ngx.ERR, "failed to get session ID: ", err) 
        return
    end  

    local sess,err = local_dict:get(sess_id)
    if sess then 
        ngx.log(ngx.ERR, sess_id, ", get sess from local dict")
    else 
        ngx.log(ngx.ERR, sess_id, ", can not get sess from local dict")
        return
    end  

    local ok, err = ssl_sess.set_serialized_session(sess)
    if not ok then 
        ngx.log(ngx.ERR, "failed to set SSL session: ", err) 
        return
    end  
end

ssl_session_fetch()
-- ssl_session_store_by_lua_file
local ssl_sess = require "ngx.ssl.session"

local local_dict = ngx.shared.ssl_session

local function ssl_session_store()
    local sess_id, err = ssl_sess.get_session_id()
    if not sess_id then 
        ngx.log(ngx.ERR, "failed to get session ID: ", err) 
        return
    end  

    local sess, err = ssl_sess.get_serialized_session()
    if not sess then         ngx.log(ngx.ERR, "failed to get SSL session: ", err) 
        return
    end  

    local ok, err = local_dict:set(sess_id, sess)
    if not ok then 
        ngx.log(ngx.ERR, sess_id, ", failed to store sess into local dict")
    else 
        ngx.log(ngx.ERR, sess_id, ", store sess into local dict")
    end  
end

ssl_session_store()

spacewander:
Session ID is obsolete in TLS 1.3.

[...]

PR #1540, opened 2019-06-10
bugfix: skipped ssl_session_fetchby_by_lua* if session ticket is present

spacewander:
…is present.

Thanks Yongjian Xu for the bug report.

### DIFF
--- a/src/ngx_http_lua_ssl_session_fetchby.c
+++ b/src/ngx_http_lua_ssl_session_fetchby.c
@@ -370,17 +370,40 @@ int
 ngx_http_lua_ssl_client_hello_handler(ngx_ssl_conn_t *ssl_conn,
     int *al, void *arg)
 {
-    int                len;
-    ngx_int_t          rc;
-    const u_char      *id;
+    size_t                       len;
+    ngx_int_t                    rc;
+    const u_char                *id;
+    ngx_connection_t            *c;
+    ngx_http_lua_ssl_ctx_t      *cctx;
 
-    len = SSL_client_hello_get0_session_id(ssl_conn, &id);
+    c = ngx_ssl_get_connection(ssl_conn);
 
-    if (len <= 0) {
-        return SSL_CLIENT_HELLO_SUCCESS;
+    cctx = ngx_http_lua_ssl_get_ctx(c->ssl->connection);
+
+    if (cctx && cctx->entered_sess_fetch_handler) {
+        /* not the first time */
+        rc = ngx_http_lua_ssl_sess_fetch_helper(ssl_conn, NULL, 0);
+
+    } else {
+        if (SSL_client_hello_get0_legacy_version(ssl_conn) > SSL3_VERSION) {
+            if (!(SSL_get_options(ssl_conn) & SSL_OP_NO_TICKET)
+                && SSL_client_hello_get0_ext(ssl_conn,
+                                             TLSEXT_TYPE_session_ticket,
+                                             NULL, &len)
+                && len > 0)
+            {
+                return SSL_CLIENT_HELLO_SUCCESS;
+            }
+        }
+
+        len = SSL_client_hello_get0_session_id(ssl_conn, &id);
+
+        if (len <= 0) {
+            return SSL_CLIENT_HELLO_SUCCESS;
+        }
+
+        rc = ngx_http_lua_ssl_sess_fetch_helper(ssl_conn, id, len);
     }
-
-    rc = ngx_http_lua_ssl_sess_fetch_helper(ssl_conn, id, len);
 
     if (rc == NGX_AGAIN) {
         return SSL_CLIENT_HELLO_RETRY;
--- a/t/143-ssl-session-fetch.t
+++ b/t/143-ssl-session-fetch.t
@@ -7,7 +7,7 @@ use File::Basename;
 
 repeat_each(3);
 
-plan tests => repeat_each() * (blocks() * 6);
+plan tests => repeat_each() * (blocks() * 5 + 15);
 
 $ENV{TEST_NGINX_HTML_DIR} ||= html_dir();
 
@@ -1215,6 +1215,151 @@ GET /t
 --- grep_error_log eval: qr/old (foo|bar): \d+/
 --- grep_error_log_out eval
 ["", "old foo: 1\n", "old bar: 1\nold foo: 2\n"]
+--- no_error_log
+[error]
+[alert]
+[emerg]
+
+
+
+=== TEST 15: don't run when session ticket is provided
+--- http_config
+    ssl_session_fetch_by_lua_block { ngx.log(ngx.ERR, "ssl fetch sess by lua is running!") }
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        server_name   test.com;
+        ssl_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
+        ssl_certificate_key $TEST_NGINX_CERT_DIR/cert/test.key;
+        server_tokens off;
+    }
+--- config
+    server_tokens off;
+    resolver $TEST_NGINX_RESOLVER ipv6=off;
+    lua_ssl_trusted_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
+
+    location /t {
+        set $port $TEST_NGINX_MEMCACHED_PORT;
+
+        content_by_lua_block {
+            do
+                local sock = ngx.socket.tcp()
+
+                sock:settimeout(5000)
+
+                local ok, err = sock:connect("unix:$TEST_NGINX_HTML_DIR/nginx.sock")
+                if not ok then
+                    ngx.say("failed to connect: ", err)
+                    return
+                end
+
+                ngx.say("connected: ", ok)
+
+                local sess, err = sock:sslhandshake(package.loaded.session, "test.com", true)
+                if not sess then
+                    ngx.say("failed to do SSL handshake: ", err)
+                    return
+                end
+
+                ngx.say("ssl handshake: ", type(sess))
+
+                package.loaded.session = sess
+
+                local ok, err = sock:close()
+                ngx.say("close: ", ok, " ", err)
+            end  -- do
+            -- collectgarbage()
+        }
+    }
+
+--- request
+GET /t
+--- response_body
+connected: 1
+ssl handshake: userdata
+close: 1 nil
+
+--- no_error_log
+[error]
+[alert]
+[emerg]
+
+
+
+=== TEST 16: always run when SSLv3 is used (SSLv3 doesn't support ticket)
+--- http_config
+    ssl_session_fetch_by_lua_block { print("ssl fetch sess by lua is running!") }
+    server {
+        listen unix:$TEST_NGINX_HTML_DIR/nginx.sock ssl;
+        server_name   test.com;
+        ssl_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
+        ssl_certificate_key $TEST_NGINX_CERT_DIR/cert/test.key;
+        ssl_protocols SSLv3;
+        server_tokens off;
+    }
+--- config
+    server_tokens off;
+    resolver $TEST_NGINX_RESOLVER ipv6=off;
+    lua_ssl_trusted_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
+    lua_ssl_protocols SSLv3;
+
+    location /t {
+        set $port $TEST_NGINX_MEMCACHED_PORT;
+
+        content_by_lua_block {
+            do
+                local sock = ngx.socket.tcp()
+
+                sock:settimeout(5000)
+
+                local ok, err = sock:connect("unix:$TEST_NGINX_HTML_DIR/nginx.sock")
+                if not ok then
+                    ngx.say("failed to connect: ", err)
+                    return
+                end
+
+                ngx.say("connected: ", ok)
+
+                local sess, err = sock:sslhandshake(package.loaded.session, "test.com", true)
+                if not sess then
+                    ngx.say("failed to do SSL handshake: ", err)
+                    return
+                end
+
+                ngx.say("ssl handshake: ", type(sess))
+
+                package.loaded.session = sess
+
+                local ok, err = sock:close()
+                ngx.say("close: ", ok, " ", err)
+            end  -- do
+            -- collectgarbage()
+        }
+    }
+
+--- request
+GET /t
+--- response_body
+connected: 1
+ssl handshake: userdata
+close: 1 nil
+
+--- grep_error_log eval: qr/ssl_session_fetch_by_lua_block:.*?,|\bssl session fetch: connection reusable: \d+|\breusable connection: \d+/
+
+--- grep_error_log_out eval
+[
+qr/\A(?:reusable connection: [01]\n)+\z/s,
+qr/^reusable connection: 1
+ssl session fetch: connection reusable: 1
+reusable connection: 0
+ssl_session_fetch_by_lua_block:1: ssl fetch sess by lua is running!,
+/m,
+qr/^reusable connection: 1
+ssl session fetch: connection reusable: 1
+reusable connection: 0
+ssl_session_fetch_by_lua_block:1: ssl fetch sess by lua is running!,
+/m,
+]
+
 --- no_error_log
 [error]
 [alert]


yongjianchn:
Maybe we should also disable ssl_session_store_by_lua* when the new session is TLSv1.3, because the stored session has no chance to be used.

spacewander:
@yongjianchn @thibaultcha 
I removed the PSK check in this PR, so that it can be merged ASAP without any temporary hacks.

thibaultcha:
@spacewander @yongjianchn Thank you for working on this - I have been following the discussion but have been busy for the last few days. I will have some time again later this week or early next week to look at this. We are not close to a release yet, so I see no urgency to fix this in the next few days, personally. But definitely before our next formal release, of course :)

src/ngx_http_lua_ssl_session_fetchby.c
```diff
@@ -370,17 +370,40 @@ int
...
 
-    rc = ngx_http_lua_ssl_sess_fetch_helper(ssl_conn, id, len);
+    if (cctx && cctx->entered_sess_fetch_handler) {
+        /* not the first time */
+        rc = ngx_http_lua_ssl_sess_fetch_helper(ssl_conn, NULL, 0);
+
+    } else {
+        if (SSL_client_hello_get0_legacy_version(ssl_conn) > SSL3_VERSION) {
+            if (!(SSL_get_options(ssl_conn) & SSL_OP_NO_TICKET)
+                && SSL_client_hello_get0_ext(ssl_conn,
```
crasyangel: Handshake Protocol: Client Hello
    Handshake Type: Client Hello (1)
    Length: 227
    Version: TLS 1.2 (0x0303)
    Random: a19ee48afa8bbb92977d2d0758209301bfef9036789d5ccb...
    Session ID Length: 32
    Session ID: f30e59969ebabeb836ad262f9e4fef379da0c1e425bc9b28...
    Cipher Suites Length: 8
    Cipher Suites (4 suites)
    Compression Methods Length: 1
    Compression Methods (1 method)
    Extensions Length: 146
    Extension: server_name (len=17)
    Extension: ec_point_formats (len=4)
    Extension: supported_groups (len=12)
    Extension: SessionTicket TLS (len=0)
    Extension: encrypt_then_mac (len=0)
    Extension: extended_master_secret (len=0)
    Extension: signature_algorithms (len=30)
    Extension: supported_versions (len=3)
    Extension: psk_key_exchange_modes (len=2)
    Extension: key_share (len=38)

TLS 1.3 send random session id (useless), and zero-length SessionTicket, it uses key_share for negotiation and session reuse

It is not a good idea that only check TLSEXT_TYPE_session_ticket, perhaps it is better that check TLSEXT_TYPE_supported_versions to identify whether TLS 1.3 or not

spacewander: @crasyangel
Thanks for the idea for checking `TLSEXT_TYPE_supported_versions`. Although I think it is enough to detect TLS 1.3 resumption via checking the PSK.

This pull request doesn't plan to make `ssl_session_fetch_by_lua*` work (or not work) in TLS 1.3.
To avoid shipping a hack to the master branch, I have removed the PSK check in the previous commit. This pull request only try to solve the problem that `ssl_session_fetch_by_lua*` is incorrectly invoked when the session can be resumed via session ticket.

We will solve the TLS 1.3 problem in another pull request.

crasyangel: The PSK check is not enough for TLS 1.3, it would send random session id even when full handshake, and no PSK ext.

yongjianchn: > The PSK check is not enough for TLS 1.3, it would send random session id even when full handshake, and no PSK ext.

+1

> It is not a good idea that only check TLSEXT_TYPE_session_ticket, perhaps it is better that check TLSEXT_TYPE_supported_versions to identify whether TLS 1.3 or not

I think this would not work either. It seems that TLSEXT_TYPE_supported_versions is parsed after client_hello_cb() called.

crasyangel: Yes, TLSEXT_TYPE_supported_versions is parsed after client_hello_cb() called. But we could use  SSL_client_hello_get0_ext to get TLSEXT_TYPE_supported_versions, Client Hello has been processed in tls_process_client_hello(), although openssl assigns ssl member in tls_early_post_process_client_hello(), such as s->version, which executes after client_hello_cb()

And, TLSEXT_TYPE_supported_versions would be a raw string which has packet prefix before client_hello_cb(), but perhaps it is enough to verify if supported_versions exists, as:
`
suppversions = &s->clienthello->pre_proc_exts[TLSEXT_IDX_supported_versions];
if (suppversions->present)
    return;
`

Maybe TLSEXT_TYPE_key_share is a better choice

yongjianchn: You mean `if (client_support_tls1.3) skip;` or `if (client_support_tls1.3 and server_support_tls1.3) skip;`?
The presence of TLSEXT_TYPE_supported_versions or TLSEXT_TYPE_key_share just indicates the client support TLS1.3, the server still can choose not to use TLS1.3. And a TLS1.3 client may carry a real session id, from a previous TLS1.2 server for example, and the server should get a chance to reuse it.

spacewander: @crasyangel @yongjianchn
For `TLSEXT_TYPE_supported_versions`, We can check if the server supports TLS 1.3 or not.

For the previous session id, maybe the client will also populate it in the PSK extension? I can not find OpenSSL does  resumption work with the legacy client id in TLS 1.3 yet. (Please correct me if I am wrong).

This pull request is not relative with TLS 1.3. Please move to issue https://github.com/openresty/lua-nginx-module/issues/1539 for further discussion.

yongjianchn: If a TLS1.3 client connected to a TLS1.2 server, for the next connection, it will carry the previous session id in TLS1.2 style. If not, how can the TLS1.2 server do session resumption? But yes, if we also check the server supports TLS1.3 or not, the problem would not exist.

spacewander: @yongjianchn

> And a TLS1.3 client may carry a real session id, from a previous TLS1.2 server for example, and the server should get a chance to reuse it.

If we know the TLS 1.3 won't be selected, we already get a change to reuse it. The `ssl_session_fetch_by_lua*` still works under TLS 1.2.

crasyangel: Indeed, you are right, session id would be useful when server does not support TLS 1.3

yongjianchn:
This PR will cause new problems for TLS1~TLS1.2.
Now the logic is `if (client_has_ticket and server support ticket) skip;`. With that logic, when the client carries a session ticket, and the server supports session ticket but refuses to reuse it(tickets_key changed or something), the session id would have no chance to be reused.
For example, we have 2 servers. They all support session ticket but have different ticket key, and they share a session cache(Redis by Lua). If a client supporting session ticket and session ID alternately accesses them, TLS session will never be reused.

This means that we have to know whether the session ticket can be reused, which is hard.

spacewander:
@yongjianchn 
You are correct. Looks like we have to take the OpenSSL patch back. Will you submit a pull request to update the OpenSSL 1.1.0d patch to make it work with OpenSSL 1.1.1?
https://github.com/openresty/openresty/blob/master/patches/openssl-1.1.0d-sess_set_get_cb_yield.patch

Also need to update the Nginx patch:
https://github.com/openresty/openresty/blob/master/patches/nginx-1.15.8-ssl_sess_cb_yield.patch

I am too busy to write the code by myself, but I will squeeze some time to review your pull request. Feel free to ping me if you have any questions.

I have tried to upgrade the patch to adapt OpenSSL 1.1.1 before: https://github.com/spacewander/openssl/commit/fcb00fe3d8e43c1afff180417bca7906903e859d.
Note that this progress is not finished. But you can take it a look.

crasyangel:
@spacewander 
Perhaps it is acceptable that looking session id in TLS 1.3, and we can control it in fetch lua?

And, this is one version for openssl-1.1.1 patch:
`
diff --git a/openssl-1.1.1b/include/openssl/bio.h b/openssl-1.1.1b/include/openssl/bio.h
index 2888b42..b1210db 100644
--- a/openssl-1.1.1b/include/openssl/bio.h
+++ b/openssl-1.1.1b/include/openssl/bio.h
@@ -219,6 +219,8 @@ void BIO_clear_flags(BIO *b, int flags);
 /* Returned from the accept BIO when an accept would have blocked */
 # define BIO_RR_ACCEPT                   0x03

+# define BIO_RR_SSL_SESSION_LOOKUP       0x09
+
 /* These are passed by the BIO callback */
 # define BIO_CB_FREE     0x01
 # define BIO_CB_READ     0x02
diff --git a/openssl-1.1.1b/include/openssl/ssl.h b/openssl-1.1.1b/include/openssl/ssl.h
index 48e1152..06d3e4c 100644
--- a/openssl-1.1.1b/include/openssl/ssl.h
+++ b/openssl-1.1.1b/include/openssl/ssl.h
@@ -896,6 +896,7 @@ __owur int SSL_extension_supported(unsigned int ext_type);
 # define SSL_ASYNC_PAUSED       5
 # define SSL_ASYNC_NO_JOBS      6
 # define SSL_CLIENT_HELLO_CB    7
+# define SSL_SESS_LOOKUP        99

 /* These will only be used when doing non-blocking IO */
 # define SSL_want_nothing(s)         (SSL_want(s) == SSL_NOTHING)
@@ -905,6 +906,7 @@ __owur int SSL_extension_supported(unsigned int ext_type);
 # define SSL_want_async(s)           (SSL_want(s) == SSL_ASYNC_PAUSED)
 # define SSL_want_async_job(s)       (SSL_want(s) == SSL_ASYNC_NO_JOBS)
 # define SSL_want_client_hello_cb(s) (SSL_want(s) == SSL_CLIENT_HELLO_CB)
+# define SSL_want_sess_lookup(s)     (SSL_want(s) == SSL_SESS_LOOKUP)

 # define SSL_MAC_FLAG_READ_MAC_STREAM 1
 # define SSL_MAC_FLAG_WRITE_MAC_STREAM 2
@@ -1190,6 +1192,8 @@ DECLARE_PEM_rw(SSL_SESSION, SSL_SESSION)
 # define SSL_ERROR_WANT_ASYNC            9
 # define SSL_ERROR_WANT_ASYNC_JOB       10
 # define SSL_ERROR_WANT_CLIENT_HELLO_CB 11
+# define SSL_ERROR_WANT_SESSION_LOOKUP  99
+# define SSL_ERROR_PENDING_SESSION      99 /* BoringSSL compatibility */
 # define SSL_CTRL_SET_TMP_DH                     3
 # define SSL_CTRL_SET_TMP_ECDH                   4
 # define SSL_CTRL_SET_TMP_DH_CB                  6
@@ -1662,6 +1667,7 @@ int SSL_SESSION_print(BIO *fp, const SSL_SESSION *ses);
 int SSL_SESSION_print_keylog(BIO *bp, const SSL_SESSION *x);
 int SSL_SESSION_up_ref(SSL_SESSION *ses);
 void SSL_SESSION_free(SSL_SESSION *ses);
+SSL_SESSION *SSL_magic_pending_session_ptr(void);
 __owur int i2d_SSL_SESSION(SSL_SESSION *in, unsigned char **pp);
 __owur int SSL_set_session(SSL *to, SSL_SESSION *session);
 int SSL_CTX_add_session(SSL_CTX *ctx, SSL_SESSION *session);
diff --git a/openssl-1.1.1b/ssl/bio_ssl.c b/openssl-1.1.1b/ssl/bio_ssl.c
index d1876d8..c59bb2a 100644
--- a/openssl-1.1.1b/ssl/bio_ssl.c
+++ b/openssl-1.1.1b/ssl/bio_ssl.c
@@ -139,6 +139,10 @@ static int ssl_read(BIO *b, char *buf, size_t size, size_t *readbytes)
         BIO_set_retry_special(b);
         retry_reason = BIO_RR_SSL_X509_LOOKUP;
         break;
+    case SSL_ERROR_WANT_SESSION_LOOKUP:
+        BIO_set_retry_special(b);
+        retry_reason = BIO_RR_SSL_SESSION_LOOKUP;
+        break;
     case SSL_ERROR_WANT_ACCEPT:
         BIO_set_retry_special(b);
         retry_reason = BIO_RR_ACCEPT;
@@ -207,6 +211,10 @@ static int ssl_write(BIO *b, const char *buf, size_t size, size_t *written)
         BIO_set_retry_special(b);
         retry_reason = BIO_RR_SSL_X509_LOOKUP;
         break;
+    case SSL_ERROR_WANT_SESSION_LOOKUP:
+        BIO_set_retry_special(b);
+        retry_reason = BIO_RR_SSL_SESSION_LOOKUP;
+        break;
     case SSL_ERROR_WANT_CONNECT:
         BIO_set_retry_special(b);
         retry_reason = BIO_RR_CONNECT;
@@ -361,6 +369,10 @@ static long ssl_ctrl(BIO *b, int cmd, long num, void *ptr)
             BIO_set_retry_special(b);
             BIO_set_retry_reason(b, BIO_RR_SSL_X509_LOOKUP);
             break;
+        case SSL_ERROR_WANT_SESSION_LOOKUP:
+            BIO_set_retry_special(b);
+            BIO_set_retry_reason(b, BIO_RR_SSL_SESSION_LOOKUP);
+            break;
         default:
             break;
         }
diff --git a/openssl-1.1.1b/ssl/ssl_lib.c b/openssl-1.1.1b/ssl/ssl_lib.c
index 4440a9f..47b83a0 100644
--- a/openssl-1.1.1b/ssl/ssl_lib.c
+++ b/openssl-1.1.1b/ssl/ssl_lib.c
@@ -3556,6 +3556,8 @@ int SSL_get_error(const SSL *s, int i)
         return SSL_ERROR_WANT_ASYNC_JOB;
     if (SSL_want_client_hello_cb(s))
         return SSL_ERROR_WANT_CLIENT_HELLO_CB;
+    if (SSL_want_sess_lookup(s))
+        return SSL_ERROR_WANT_SESSION_LOOKUP;

     if ((s->shutdown & SSL_RECEIVED_SHUTDOWN) &&
         (s->s3->warn_alert == SSL_AD_CLOSE_NOTIFY))
 int SSL_CTX_use_PrivateKey_file(SSL_CTX *ctx, const char *file, int type)
diff --git a/openssl-1.1.1b/ssl/ssl_sess.c b/openssl-1.1.1b/ssl/ssl_sess.c
index 5ad2792..01b0bf0 100644
--- a/openssl-1.1.1b/ssl/ssl_sess.c
+++ b/openssl-1.1.1b/ssl/ssl_sess.c
@@ -16,6 +16,8 @@
 #include "ssl_locl.h"
 #include "statem/statem_locl.h"

+static const char g_pending_session_magic = 0;
+
 static void SSL_SESSION_list_remove(SSL_CTX *ctx, SSL_SESSION *s);
 static void SSL_SESSION_list_add(SSL_CTX *ctx, SSL_SESSION *s);
 static int remove_session_lock(SSL_CTX *ctx, SSL_SESSION *c, int lck);
@@ -476,6 +478,10 @@ SSL_SESSION *lookup_sess_in_cache(SSL *s, const unsigned char *sess_id,

         ret = s->session_ctx->get_session_cb(s, sess_id, sess_id_len, &copy);

+        if (ret == SSL_magic_pending_session_ptr()) {
+            return ret; /* Retry later */
+        }
+
         if (ret != NULL) {
             tsan_counter(&s->session_ctx->stats.sess_cb_hit);

@@ -564,6 +570,10 @@ int ssl_get_prev_session(SSL *s, CLIENTHELLO_MSG *hello)
                 try_session_cache = 1;
                 ret = lookup_sess_in_cache(s, hello->session_id,
                                            hello->session_id_len);
+
+                if (ret == SSL_magic_pending_session_ptr()) {
+                    return -2; /* Retry later */
+                }
             }
             break;
         case SSL_TICKET_NO_DECRYPT:
@@ -989,6 +999,11 @@ X509 *SSL_SESSION_get0_peer(SSL_SESSION *s)
     return s->peer;
 }

+SSL_SESSION *SSL_magic_pending_session_ptr(void)
+{
+    return (SSL_SESSION *) &g_pending_session_magic;
+}
+
 int SSL_SESSION_set1_id_context(SSL_SESSION *s, const unsigned char *sid_ctx,
                                 unsigned int sid_ctx_len)
 {
diff --git a/openssl-1.1.1b/ssl/statem/statem_srvr.c b/openssl-1.1.1b/ssl/statem/statem_srvr.c
index 6b8aae6..9f16c07 100644
--- a/openssl-1.1.1b/ssl/statem/statem_srvr.c
+++ b/openssl-1.1.1b/ssl/statem/statem_srvr.c
@@ -1595,6 +1595,7 @@ static int tls_early_post_process_client_hello(SSL *s)
 #ifndef OPENSSL_NO_COMP
     SSL_COMP *comp = NULL;
 #endif
+    PACKET ciphersuites;
     const SSL_CIPHER *c;
     STACK_OF(SSL_CIPHER) *ciphers = NULL;
     STACK_OF(SSL_CIPHER) *scsvs = NULL;
@@ -1707,6 +1708,7 @@ static int tls_early_post_process_client_hello(SSL *s)
     }

     s->hit = 0;
+    ciphersuites = clienthello->ciphersuites;

     if (!ssl_cache_cipherlist(s, &clienthello->ciphersuites,
                               clienthello->isv2) ||
@@ -1812,6 +1814,10 @@ static int tls_early_post_process_client_hello(SSL *s)
         } else if (i == -1) {
             /* SSLfatal() already called */
             goto err;
+        } else if (i == -2) {
+            clienthello->ciphersuites = ciphersuites; // reset ciphersuites
+            s->rwstate = SSL_SESS_LOOKUP;
+            return -1;
         } else {
             /* i == 0 */
             if (!ssl_get_new_session(s, 1)) {
@@ -1819,6 +1825,7 @@ static int tls_early_post_process_client_hello(SSL *s)
                 goto err;
             }
         }
+        s->rwstate = SSL_NOTHING;
     }

     if (SSL_IS_TLS13(s)) {
diff --git a/openssl-1.1.1b/util/libssl.num b/openssl-1.1.1b/util/libssl.num
index 297522c..abb2a44 100644
--- a/openssl-1.1.1b/util/libssl.num
+++ b/openssl-1.1.1b/util/libssl.num
@@ -498,3 +498,4 @@ SSL_CTX_get_recv_max_early_data         498 1_1_1   EXIST::FUNCTION:
 SSL_CTX_set_recv_max_early_data         499    1_1_1   EXIST::FUNCTION:
 SSL_CTX_set_post_handshake_auth         500    1_1_1   EXIST::FUNCTION:
 SSL_get_signature_type_nid              501    1_1_1a  EXIST::FUNCTION:
+SSL_magic_pending_session_ptr           999    1_1_0   EXIST::FUNCTION:
`

yongjianchn:
@spacewander 
From next week, I will start traveling for three weeks. And after that, I can only use my spare time to do this work. So this could be delayed by one to two months if you wait for me. 
@crasyangel 
Can you work on this for the next few days? Seems you already did a lot of work on TLS1.3.

spacewander:
@yongjianchn @crasyangel 
I decide to do the stuff by myself, so that it could be review ASAP. Don't want to wait too long.

PR #1543, opened 2019-06-15
optimize: improved code cache lookup performance.

thibaultcha:

### Summary

This PR aims at reducing the code cache lookup cost by avoiding the use of `lj_str_new()`. Performance gains will be measured in the coming days, but #1486 is a similar implementation and showed preliminary improvements of ~2% to ~5%.

### #1486 vs. this PR

The difference between the two is that #1486 completely removes code cache strings in favor of `luaL_ref()` references, while this PR preserves code cache strings, and adds `luaL_ref()` as an optimization only. This makes for an important difference in behavior, explained below.

- This PR: removing string cache keys (#1486) creates a distinct `clfactory` closure for each `*_by_lua` directive, even if their Lua chunks are identical.
- #1486: keeping the string cache key (this PR) allows for `*_by_lua` identical Lua chunks (as per `ngx_http_lua_digest_hex()`) to share the same`clfactory` closure, as per the current behavior (non-breaking).

See the behavior change in action [at this line of the test suite](https://github.com/openresty/lua-nginx-module/pull/1486/files#diff-ffea124057bc9c9adc35e4fe4ca6b9d0R1371) (different closures for the same Lua chunk with `ref = 3`), compared to [this line of the test suite](https://github.com/thibaultcha/lua-nginx-module/commit/a2d6f17bb4cfbf69a7eee22d523ae24e17bc5800#diff-ffea124057bc9c9adc35e4fe4ca6b9d0R1316) in this PR (same closure for identical Lua chunks with `ref = 2`).

### Other changes in this PR

This PR **intentionally** contains several commits that should not be squashed together:

- tests: added more complete test cases for NGINX variables in `*_by_lua_file` directives to prevent future regressions (this was added before any changes were made to the underlying implementation).
- tests: completed a test case to ensure that different `*_by_lua` directives with identical Lua chunks do not produce the same closure, as per the current behavior and to prevent future regressions.
- optimize: improved code cache lookup performance. This is the main commit of this PR.
- bugfix: fixed an issue with `set_by_lua_file` code caching being invalid when using NGINX variables (**this fix is also included in #1486**). 
- refactor: reduce cache key size by removing unnecessary components.
- refactor: greatly simplified and DRY out cache key generation code.

Besides the performance gains, however modest they may be (I still have to run more scientific benchmarks), this also greatly simplifies code cache key maintainability, readability, and test coverage. **Most of the added lines of codes in this PR are due to new test cases previously non-existing or incomplete**.

### Notes to reviewers

It will be easiest to review commits individually and in order, as tests were always added prior any refactor or optimization, which were added iteratively.


src/ngx_http_lua_module.c
```diff
@@ -972,20 +972,29 @@ ngx_http_lua_create_srv_conf(ngx_conf_t *cf)
...
      *
      *      lscf->srv.ssl_session_fetch_handler = NULL;
      *      lscf->srv.ssl_session_fetch_src = { 0, NULL };
      *      lscf->srv.ssl_session_fetch_src_key = NULL;
+     *      lscf->srv.ssl_session_fetch_src_ref = 0;
      *
      *      lscf->balancer.handler = NULL;
      *      lscf->balancer.src = { 0, NULL };
      *      lscf->balancer.src_key = NULL;
+     *      lscf->balancer.src_ref = 0;
```
dndx: A really minor styling detail: We generally only comment about the `set by ngx_pcalloc` if it is not initialized by this function. Since those fields are initialized to `LUA_REFNIL` below, no need to add these comments.

See http://lxr.nginx.org/source/src/http/ngx_http_core_module.c#3375 as an example.

thibaultcha: Correct and done, thanks!

PR #1544, opened 2019-06-15
bugfix: supported OpenSSL 1.1.1 by upgrading the OpenSSL patch.

spacewander:
Previously, we used ClientHello callback to do ssl session fetching
non-blockingly. However, this way can not handle an edge case: the ssl
session resumption via session ticket might fail, and the client falls
back to session ID resumption. The ClientHello callback is run too early
to know if the client will fall back to use session ID resumption.

Therefore, we have to take back the OpenSSL sess_set_get_cb_yield patch and
upgrade it to adapt OpenSSL 1.1.1.

Thanks @yongjianchn and @crasyangel for their help.

Current implementation skips `ssl_session_fetch|store_by_lua*` for TLS 1.3 client. It is possible to support PSK with session ID in TLS 1.3. But it need to modify a number of functions to pass the result up, which will make the patch too complex to maintain. Since PSK with session ticket is supported, supporting PSK with session ID is not so profitable. If someone needs this feature, he/she can submit another pull request.

I have tested the change of `ssl_session_fetch_by_lua*` locally with this test case:

```
=== TEST 17: skip ssl_session_fetch in TLSv1.3
--- http_config
    lua_shared_dict done 16k;
    ssl_session_fetch_by_lua_block { error("should not run") }
    server {
        listen $TEST_NGINX_SERVER_SSL_PORT ssl;
        server_name   test.com;
        ssl_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;
        ssl_certificate_key $TEST_NGINX_CERT_DIR/cert/test.key;
        ssl_protocols TLSv1.3;
        ssl_session_tickets off;
        server_tokens off;

        location / {
            content_by_lua_block {
                ngx.shared.done:set("handshake", true)
            }
        }
    }
--- config
    server_tokens off;
    lua_ssl_trusted_certificate $TEST_NGINX_CERT_DIR/cert/test.crt;

    location /t {
        content_by_lua_block {
            local req = "'GET / HTTP/1.0\r\nHost: test.com\r\nConnection: close\r\n\r\n'"
            local id = ("1"):rep(32)
            local ngx_pipe = require "ngx.pipe"
            local cmd = "echo -n " .. req .. " | " ..
                "/home/lzx/openssl-1.1.1b/apps/openssl s_client -connect 127.0.0.1:$TEST_NGINX_SERVER_SSL_PORT -tls1_3 -psk_test_id " .. id
            local proc, err = ngx_pipe.spawn({"sh", "-c", cmd})
            if not proc then
                ngx.say(err)
                return
            end

            local step = 0.001
            while step < 2 do
                ngx.sleep(step)
                step = step * 2

                if ngx.shared.done:get("handshake") then
                    local out = proc:stderr_read_all()
                    ngx.log(ngx.INFO, out)
                    ngx.say("ok")
                    return
                end
            end

            ngx.log(ngx.ERR, "openssl client handshake timeout")
        }
    }

--- request
GET /t
--- response_body
ok
--- error_log
ssl session fetch: skipped: TLS version 304
O = OpenResty, OU = OpenResty
--- no_error_log
[error]
[alert]
[emerg]
```

Since `openssl s_client` doesn't support PSK with session ID, I have to modify its source code and add a way to specify a fake session ID in the PSK. Maybe I need to submit this change to `openresty/openssl` as a test dependency? `curl` doesn't support PSK at all.

TODO: update documentation.

### DIFF
--- a/src/ngx_http_lua_ssl_session_fetchby.c
+++ b/src/ngx_http_lua_ssl_session_fetchby.c
@@ -181,6 +181,9 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn,
 #endif
     u_char *id, int len, int *copy)
 {
+#if defined(TLS1_3_VERSION)
+    int                              version;
+#endif
     lua_State                       *L;
     ngx_int_t                        rc;
     ngx_connection_t                *c, *fc = NULL;
@@ -197,6 +200,16 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn,
     *copy = 0;
 
     c = ngx_ssl_get_connection(ssl_conn);
+
+#if defined(TLS1_3_VERSION)
+    version = SSL_version(ssl_conn);
+
+    if (version >= TLS1_3_VERSION) {
+        ngx_log_debug1(NGX_LOG_DEBUG_HTTP, c->log, 0,
+                       "ssl session fetch: skipped: TLS version %xd", version);
+        return 0;
+    }
+#endif
 
     ngx_log_debug1(NGX_LOG_DEBUG_HTTP, c->log, 0,
                    "ssl session fetch: connection reusable: %ud", c->reusable);
--- a/src/ngx_http_lua_ssl_session_fetchby.c
+++ b/src/ngx_http_lua_ssl_session_fetchby.c
@@ -173,12 +173,16 @@ ngx_http_lua_ssl_sess_fetch_by_lua(ngx_conf_t *cf, ngx_command_t *cmd,
 }
 
 
-static ngx_int_t
-ngx_http_lua_ssl_sess_fetch_helper(ngx_ssl_conn_t *ssl_conn,
-    const u_char *id, int len)
+/* cached session fetching callback to be set with SSL_CTX_sess_set_get_cb */
+ngx_ssl_session_t *
+ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn,
+#if OPENSSL_VERSION_NUMBER >= 0x10100003L
+    const
+#endif
+    u_char *id, int len, int *copy)
 {
     lua_State                       *L;
-    ngx_int_t                        rc, res = NGX_ERROR;
+    ngx_int_t                        rc;
     ngx_connection_t                *c, *fc = NULL;
     ngx_http_request_t              *r = NULL;
     ngx_pool_cleanup_t              *cln;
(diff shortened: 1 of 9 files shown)


spacewander:
Relative pull request: https://github.com/openresty/openresty/pull/497

spacewander:
Also depends on https://github.com/openresty/test-nginx/pull/93

src/ngx_http_lua_ssl_session_fetchby.c
```diff
@@ -181,6 +181,9 @@ ngx_http_lua_ssl_sess_fetch_handler(ngx_ssl_conn_t *ssl_conn,
 #endif
     u_char *id, int len, int *copy)
 {
+#if defined(TLS1_3_VERSION)
```
thibaultcha: `#ifdef TLS_1_3_VERSION` is enough here?

Or actually maybe `#if defined(NGX_SSL_TLSv1_3) && defined(TLS1_3_VERSION)` for older Nginx versions linked with newer OpenSSL versions? Not sure from the top of my head which Nginx core those directives were first supported with.

spacewander: @thibaultcha
Checking `NGX_SSL_TLSv1_3` is fine, since if the Nginx doesn't support TLS 1.3 will not use TLS 1.3.

However, I think checking `TLS1_3_VERSION` is enough.
1. `SSL_version()` is inexpensive operation.
2. If you are using an old Nginx, you are probably also using an old lua-nginx-module/OpenResty which doesn't have this commit. And it is quite rare to build an old Nginx with a new OpenSSL.
3. personally speaking, if we can avoid it, it is not a good idea to add additional code to make legacy system happy. If people want to build Nginx in a new Linux distribution, they should build a new one or provide old version of OpenSSL themselves. An old Nginx with a new OpenSSL is not tested well and should be avoided.

thibaultcha: Yeah, i just meant to switch to the `#ifdef` notation. `NGX_SSL_TLSv1_3` isn’t very helpful anyway, albeit still safer I would argue.

spacewander: @thibaultcha 
Done via the new commit.

thibaultcha:
@spacewander Just an FYI: I reverted the previous 1.1.1 patches in various repositories. 

I will be be out until Monday, but will then pull your commits in and make everything work fine on 1.17.1. Everything is already planned, I don’t think there is anything you have to do for now 👍🏻 That’s all :)

thibaultcha:
Merged, thanks!

PR #1546, opened 2019-06-17
fix(set_more_tries): fix infinite retry when no available servers and…

### LINKED ISSUE #1545: Bug set_more_tries  cause infinite retrying when no valid servers and proxy_next_proxy_tries=0
* when nginx has no valid servers and proxy_next_upstream_tries=0 will cause infinite retrying. this can be reproduce use next configuaration
```
worker_processes  1;
pid /usr/local/nginx/logs/nginx.pid;
daemon off;
master_process off;
error_log  /usr/local/nginx/logs/error.log debug;
events {
    worker_connections  1024;
}
http {
    lua_package_cpath "/path/to/openresty-lualib/?.so;;";
    lua_package_path "/path/to/openresty-lualib/?.lua;;";
    server {
        listen       80;
        server_name  localhost;
        location / {
          proxy_pass http://test;
       }
    }
    upstream test {
      server 127.0.0.1:30909;
      balancer_by_lua_block {
  local ngx_balancer = require("ngx.balancer")
  local ok, err = ngx_balancer.set_current_peer("127.0.0.1", "30909")
  if not ok then
     ngx.log(ngx.ERR, string.format("error while setting current upstream peer %s: %s", peer, err))
    return ngx.exit(-3)
  end
  ok, err = ngx_balancer.set_more_tries(1)
  if not ok then
   ngx.log(ngx.ERR, "set_more_tries failed, ", err)
    return
  end
      }
    }
}
```  
* my nginx  configuration:
```
nginx version: nginx/1.14.2
built by clang 10.0.0 (clang-1000.10.44.4)
built with OpenSSL 1.1.1b  26 Feb 2019
TLS SNI support enabled
configure arguments: --prefix=/usr/local/nginx --with-ld-opt=-Wl,-rpath,/usr/local/Cellar/openresty/1.13.6.2/luajit/lib --add-module=../lua-nginx-module-0.10.13 --add-module=../lua-upstream-nginx-module-0.07 --add-module=../nginx_upstream_check_module --with-http_ssl_module --with-http_stub_status_module --with-http_v2_module --with-pcre=../pcre-8.43 --with-openssl=../openssl-1.1.1b --with-debug
```
* lua-nginx-module version: 0.10.13

* reason
i find this is because in set_more_tries function has not deal with the situation when proxy_next_upstream_tries = 0, then it cause nginx use invalid server to infinite retrying  cause the peer.tries never decrease to 0.
```
int ngx_http_lua_ffi_balancer_set_more_tries(ngx_http_request_t *r, int count, char **err) {
...
#if (nginx_version >= 1007005)
    max_tries = r->upstream->conf->next_upstream_tries;
    total = bp->total_tries + r->upstream->peer.tries - 1;

    if (max_tries && total + count > max_tries) {
        count = max_tries - total;
        *err = "reduced tries due to limit";

    } else {
        *err = NULL;
    }
#else
    *err = NULL;
#endif

    bp->more_tries = count;
    return NGX_OK
}
``` 

* ngx_http_upstream_next
```
static void
ngx_http_upstream_next(ngx_http_request_t *r, ngx_http_upstream_t *u,
    ngx_uint_t ft_type)
{
    ....
   // will never enter this if body 
    if (u->peer.tries == 0
        || ((u->conf->next_upstream & ft_type) != ft_type)
        || (u->request_sent && r->request_body_no_buffering)
        || (timeout && ngx_current_msec - u->peer.start_time >= timeout))
    {
#if (NGX_HTTP_CACHE)

        if (u->cache_status == NGX_HTTP_CACHE_EXPIRED
            && ((u->conf->cache_use_stale & ft_type) || r->cache->stale_error))
        {
            ngx_int_t  rc;

            rc = u->reinit_request(r);

            if (rc != NGX_OK) {
                ngx_http_upstream_finalize_request(r, u, rc);
                return;
            }

            u->cache_status = NGX_HTTP_CACHE_STALE;
            rc = ngx_http_upstream_cache_send(r, u);

            if (rc == NGX_DONE) {
                return;
            }

            if (rc == NGX_HTTP_UPSTREAM_INVALID_HEADER) {
                rc = NGX_HTTP_INTERNAL_SERVER_ERROR;
            }

            ngx_http_upstream_finalize_request(r, u, rc);
            return;
        }
#endif

        ngx_http_upstream_finalize_request(r, u, status);
        return;
    }
  ....  
    ngx_http_upstream_connect(r, u);
}
```

* fix
i think this can fix by add a default value to proxy_next_upstream_tries when it equals 0,  it is better to set to the number of the upstream peers, but we can't get this value from this method
```
int ngx_http_lua_ffi_balancer_set_more_tries(ngx_http_request_t *r, int count, char **err) {
...
#if (nginx_version >= 1007005)
    max_tries = r->upstream->conf->next_upstream_tries;
    // set a default tries to avoid infinite tries when all peers are not available
    max_tries = max_tries == 0 ? 3 : max_tries; 
    total = bp->total_tries + r->upstream->peer.tries - 1;

    if (max_tries && total + count > max_tries) {
        count = max_tries - total;
        *err = "reduced tries due to limit";

    } else {
        *err = NULL;
    }
#else
    *err = NULL;
#endif

    bp->more_tries = count;
    return NGX_OK
}
```
* pull request:

anicca-sa:
set default proxy_next_proxy_tries to 3 to avoid infinite retry when proxy_next_proxy_tries=0 and no
valid servers

fix #1545

related issues：https://github.com/openresty/lua-nginx-module/issues/1545

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -692,8 +692,9 @@ ngx_http_lua_ffi_balancer_set_more_tries(ngx_http_request_t *r,
 
 #if (nginx_version >= 1007005)
     max_tries = r->upstream->conf->next_upstream_tries;
-    // set a default tries to avoid infinite tries when all peers are not available
-    max_tries = max_tries == 0 ? 3 : max_tries; 
+    // set a default tries to avoid infinite retrying
+    // when all peers are invalid with default proxy_next_upstream_tries
+    max_tries = max_tries == 0 ? 3 : max_tries;
     total = bp->total_tries + r->upstream->peer.tries - 1;
 
     if (max_tries && total + count > max_tries) {
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -692,6 +692,8 @@ ngx_http_lua_ffi_balancer_set_more_tries(ngx_http_request_t *r,
 
 #if (nginx_version >= 1007005)
     max_tries = r->upstream->conf->next_upstream_tries;
+    // set a default tries to avoid infinite tries when all peers are not available
+    max_tries = max_tries == 0 ? 3 : max_tries; 
     total = bp->total_tries + r->upstream->peer.tries - 1;
 
     if (max_tries && total + count > max_tries) {
--- a/t/138-balancer.t
+++ b/t/138-balancer.t
@@ -522,4 +522,3 @@ http next upstream, 2
 http next upstream, 2
 --- error_log
 failed to set more tries: reduced tries due to limit
-[alert]
--- a/t/138-balancer.t
+++ b/t/138-balancer.t
@@ -522,3 +522,4 @@ http next upstream, 2
 http next upstream, 2
 --- error_log
 failed to set more tries: reduced tries due to limit
+
--- a/t/138-balancer.t
+++ b/t/138-balancer.t
@@ -520,7 +520,6 @@ set more tries: reduced tries due to limit
 http next upstream, 2
 http next upstream, 2
 http next upstream, 2
-http next upstream, 2
---- no_error_log
+--- error_log
 failed to set more tries: reduced tries due to limit
 [alert]


Crack:
Correct me if i am wrong, but are you basically forcing a cap on the value of `proxy_next_upstream_tries` to 3? `proxy_next_upstream_tries: 0` is meant to cause infinite retries, and `balancer.set_current_peer` gives nginx an address for current try. So an infinite loop is a result of programmer's error, not OpenResty's.

thibaultcha:
@anicca-sa Indeed, @Crack is correct here. See the documentation on [balancer.set_more_tries()](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#set_more_tries):

> Please note that, the total number of tries in a single downstream request cannot exceed the hard limit configured by directives like proxy_next_upstream_tries

Having `proxy_next_upstream_tries 0;` is not only acceptable, it is the default value for this directive (see http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream_tries). This not only breaks a behavior used by many OpenResty applications in the wild, it also limits the power and flexibility of `balancer.set_more_tries()`.

@Crack is also correct when pointing out that this is your own programming error with regards to `balancer.set_more_tries()` usage.

Closing this PR now, thank you.

PR #1560, opened 2019-07-13
bugfix: response header field should not contains `\r` or `\n`.

doujiang24:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4189,7 +4189,7 @@ to be returned when reading `ngx.header.Foo`.
 
 Note that `ngx.header` is not a normal Lua table and as such, it is not possible to iterate through it using the Lua `ipairs` function.
 
-Notice: both the `HEADER` and the `VALUE` should not contain `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, including the `\r` or `\n` themself.
+Notice: both the `HEADER` and the `VALUE` should not contain `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, including the `\r` or `\n` bytes themself.
 
 For reading *request* headers, use the [ngx.req.get_headers](#ngxreqget_headers) function instead.
 
@@ -5156,7 +5156,7 @@ ngx.redirect
 
 Issue an `HTTP 301` or `302` redirection to `uri`.
 
-Notice: the `uri` should not contains `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, including the `\r` or `\n` themself.
+Notice: the `uri` should not contains `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, including the `\r` or `\n` bytes themself.
 
 The optional `status` parameter specifies the HTTP status code to be used. The following status codes are supported right now:
 
--- a/README.markdown
+++ b/README.markdown
@@ -4189,6 +4189,8 @@ to be returned when reading `ngx.header.Foo`.
 
 Note that `ngx.header` is not a normal Lua table and as such, it is not possible to iterate through it using the Lua `ipairs` function.
 
+Notice: both the `HEADER` and the `VALUE` should not contains `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, include the `\r` or `\n` themself.
+
 For reading *request* headers, use the [ngx.req.get_headers](#ngxreqget_headers) function instead.
 
 [Back to TOC](#nginx-api-for-lua)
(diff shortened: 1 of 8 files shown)


README.markdown
```diff
@@ -4189,6 +4189,8 @@ to be returned when reading `ngx.header.Foo`.
 
 Note that `ngx.header` is not a normal Lua table and as such, it is not possible to iterate through it using the Lua `ipairs` function.
 
+Notice: both the `HEADER` and the `VALUE` should not contains `\r` or `\n`, otherwise, the characters after `\r` or `\n` will be truncated, include the `\r` or `\n` themself.
```
agentzh: Grammar: "include" => "including".

"should not contains" => "should not contain".

src/ngx_http_lua_control.c
```diff
@@ -248,6 +248,8 @@ ngx_http_lua_ngx_redirect(lua_State *L)
                           "the headers");
     }
 
+    len = ngx_http_lua_safe_header_filed_len(p, len);
```
agentzh: I have problem understanding the meaning of the word "filed" used here. Seems like `value` is a better word here for `filed`?

src/ngx_http_lua_util.c
```diff
@@ -4266,4 +4266,25 @@ ngx_http_lua_set_sa_restart(ngx_log_t *log)
 #endif
 
 
+size_t
+ngx_http_lua_safe_header_filed_len(u_char *str, size_t len)
```
agentzh: Better make this an inline function defined in `util.h` instead to avoid C function call overhead.

src/ngx_http_lua_util.c
```diff
@@ -4266,4 +4266,25 @@ ngx_http_lua_set_sa_restart(ngx_log_t *log)
...
+    p = (u_char *) memchr(str, 13, len);
+    if (p != NULL) {
+        len = p - str;
+    }
+
+    /* search \n */
+    p = (u_char *) memchr(str, 10, len);
+    if (p != NULL) {
+        len = p - str;
+    }
```
agentzh: Since header values are usually quite short, maybe it is even faster to write a single tight loop to scan for `\r` and `\n` than 2 `memchr()` calls. We can do some micro benchmark if you have any doubts :)

doujiang24: `memchr` is faster than `loop scan` in such case: https://gist.github.com/doujiang24/cbf071ad5b40d2bdf67b372732644676

1.2s vs 1.6s on my side.

src/ngx_http_lua_control.c
doujiang24: sorry, it's a typo, it's `field` since we want to take care of both the header name and the header value.

doujiang24:
@agentzh All comments addressed, thanks :)

src/ngx_http_lua_util.c
doujiang24: for the short strings, like the length shorter than 18, `loop` run faster, changed to use `loop` instead of `memchr`.

src/ngx_http_lua_util.h
```diff
@@ -260,6 +260,23 @@ void ngx_http_lua_set_sa_restart(ngx_log_t *log);
     }
 
 
+static ngx_inline size_t
+ngx_http_lua_safe_header_field_len(u_char *str, size_t len)
+{
+    size_t i;
+
+    for (i = 0; i < len; i++) {
+        if (*str == 13 || *str == 10) {
```
agentzh: I think it's more readable to write `\r` and `\n` instead. The performance is exactly the same.

src/ngx_http_lua_util.h
```diff
@@ -260,6 +260,23 @@ void ngx_http_lua_set_sa_restart(ngx_log_t *log);
     }
 
 
+static ngx_inline size_t
+ngx_http_lua_safe_header_field_len(u_char *str, size_t len)
+{
+    size_t i;
+
+    for (i = 0; i < len; i++) {
```
agentzh: We could put `str++` here, right after `i++`, separated by a comma.

src/ngx_http_lua_util.h
```diff
@@ -260,6 +260,21 @@ void ngx_http_lua_set_sa_restart(ngx_log_t *log);
     }
 
 
+static ngx_inline size_t
+ngx_http_lua_safe_header_value_len(u_char *str, size_t len)
+{
+    size_t i;
+
+    for (i = 0; i < len; i++, str++) {
+        if (*str == '\r' || *str == '\n') {
```
thibaultcha: Should we maybe use the `LF` and `CR` macros here instead?

thibaultcha: Will merge without this suggestion for now if we think it is more readable. We can always update later if we change our mind.

thibaultcha:
Merged, thanks!

src/ngx_http_lua_util.h
agentzh: Oh yeah, CR and LF are better.

PR #1574, opened 2019-08-07
Add ngx_http_lua_ffi_shdict_store_when for set_when and safe_set_when

hnakamur:
This pull request adds `ngx_http_lua_ffi_shdict_store_when`
which is used to define `set_when` and `safe_set_when` methods in `ngx.shared.DICT`.

These methods are useful for avoiding race conditions between successive calls of `get` and `set`.
You pass `old_value` as well as `value` to `set_when`, and it changes the value only when the value for the key is still the same as `old_value`.
Adding this method is suggested in the [comment](https://github.com/openresty/lua-resty-limit-traffic/pull/48#issuecomment-515670492) by @doujiang24.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3261,6 +3261,8 @@ Nginx API for Lua
 * [ngx.shared.DICT.get_stale](#ngxshareddictget_stale)
 * [ngx.shared.DICT.set](#ngxshareddictset)
 * [ngx.shared.DICT.safe_set](#ngxshareddictsafe_set)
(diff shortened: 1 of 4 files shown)


hnakamur:
Test with the following pull request branches passes at https://travis-ci.org/hnakamur/lua-nginx-module/builds/569371351

* https://github.com/openresty/lua-resty-core/pull/268
* https://github.com/openresty/stream-lua-nginx-module/pull/171

PR #1581, opened 2019-08-23
When building from RPM auto/feature lacks LD_LIBRARY_PATH to be able to run the test.

arcivanov:
Basically, the issue is when building as part of an RPM the luajit SO is in temp location and is not rpath'ed but the the temp location is not in ld config so test doesn't run as it can't find luajit SO.

This is a hack to demo how to remediate it for RPM build, but it's obviously insufficient for all possible build paths and should probably not applied as is.


thibaultcha:
Hi,

This is fixed in #1577, but considering the disrupting nature of the PR, I just opened #1582 which ports the fix for this (as well as another one for static libpcre builds). Would you mind giving it a try? The fix in this PR would not be sufficient for other platforms.

PR #1584, opened 2019-08-28
feature: we now ignore 'lua_regex_*' directives when NGINX is compiled without PCRE support.

thibaultcha:
Similarly as other directives and options (e.g. our patch to support `resolver ... ipv6=...` on builds without IPv6 support), `lua_regex_cache_max_entries` and `lua_regex_match_limit` are now supported when NGINX is built without PCRE support. The directives are simply ignored.

Concretely, this allows for `resty` to be used even when NGINX lacks PCRE support, since [it hard-codes the lua_regex_cache_max_entries directive](https://github.com/openresty/resty-cli/blob/master/bin/resty#L734).

When users attempt to use PCRE APIs (e.g. `ngx.re`), lua-resty-core already produces friendly errors describing the lack of support for PCRE in the current NGINX build since recently.

Sister PR:
- https://github.com/openresty/lua-resty-core/pull/270

No tests for this change at the moment, as our test matrix does not include a build without PCRE support, which is something I will work on in the future, after revisiting the CI setup.

### DIFF
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -232,6 +232,30 @@ ngx_http_lua_package_path(ngx_conf_t *cf, ngx_command_t *cmd, void *conf)
 }
 
 
+char *
+ngx_http_lua_regex_cache_max_entries(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf)
+{
+#if (NGX_PCRE)
+    return ngx_conf_set_num_slot(cf, cmd, conf);
+#else
+    return NGX_CONF_OK;
+#endif
+}
+
+
+char *
+ngx_http_lua_regex_match_limit(ngx_conf_t *cf, ngx_command_t *cmd,
+    void *conf)
+{
+#if (NGX_PCRE)
+    return ngx_conf_set_num_slot(cf, cmd, conf);
+#else
+    return NGX_CONF_OK;
+#endif
+}
+
+
 #if defined(NDK) && NDK
 char *
 ngx_http_lua_set_by_lua_block(ngx_conf_t *cf, ngx_command_t *cmd,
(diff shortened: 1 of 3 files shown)


agentzh:
Silently ignoring configurations can be confusing for the user. I think a better practice is to emit warnings to the error log file at least.

thibaultcha:
@agentzh This was indeed my original intention, but it isn't viable either, because `resty` defines its default logging level to `warn`, so adding a warning defeats the purpose of what this PR is trying to do.

This PR's goal is to make the of `resty` on a non-PCRE enabled OpenResty build viable. However, adding a warning log would mean that using `resty` on such non-PCRE enabled builds would always output warning logs, which both give the perception to users that `resty` can only be used with PCRE enabled, _and_ would make the parsing of `resty`'s stderr logs quite annoying as well:

```
 $ resty -e 'print("hello")'  
nginx: [warn] NO SUPPORT NO PCRE in /tmp/resty_nrvpvFSLyk/conf/nginx.conf:107
hello
```

The goal that I have been pursing is to allow for `resty` to be used with non-PCRE enabled builds, and only fail when the user attempts to use PCRE in such cases. As of recently, this is what `resty` will print in such cases already:

```
$ resty -e 'require "ngx.re"'
ERROR: (command line -e):1: no support for 'ngx.re' module: OpenResty was compiled without PCRE support
stack traceback:
        ../lua-resty-core/lib/ngx/re.lua:12: in main chunk
        [C]: in function 'require'
        (command line -e):1: in function 'inline_gen'
        init_worker_by_lua:45: in function <init_worker_by_lua:44>
        [C]: in function 'xpcall'
        init_worker_by_lua:53: in function <init_worker_by_lua:51>
```

The idea here is **indeed** to **not** do anything on those directives on non-enabled PCRE builds. It doesn't matter whether they are specified or not, so long as the user does **not** use `ngx.re` or other PCRE APIs.

PR #1599, opened 2019-09-14
feature: cosocket client certificate support (client mutual TLS handshake)

dndx:

This PR aims to provide a method to support client TLS certificate for cosocket API.

An existing PR https://github.com/openresty/lua-nginx-module/pull/997 tried to solve the same problem but has not been accepted by the maintainers, primarily due to runtime safety concern and easiness of use issues. 

This PR's approach is similar to the existing `ssl_certificate_by*` APIs by reusing the parser provided `ngx.ssl` module. Only `cdata` pointer is passed which is harder to mess up (compared to full `SSL_CTX` swap). In addition, this design means `cdata` can still be cached and reused like we do in `ssl_certificate_by*` today.

The introduction of options table means we can add more options like those (such as cipher suite) into the API without having to add more positional arguments, which is already getting long.

Depends on: https://github.com/openresty/luajit2/pull/74

If `OPENRESTY_LUAJIT` is not detected, attempting to set client certificate using the option table throws an Lua error.


dndx:
@agentzh @thibaultcha would you please take a look at this PR and let me know what you thinks? Thanks very much!

t/129-ssl-socket.t
```diff
@@ -2508,6 +2545,41 @@ SSL reused session
...
+
+            local ok, err = sock:connect("openresty.org", 443)
+            if not ok then
+                ngx.say("failed to connect: ", err)
+                return
+            end
+
+            ngx.say("connected: ", ok)
+
+            local session, err = sock:sslhandshake(1, 2, 3, 4, 5, 6)
```
agentzh: This test case assumes that the number of arg check is performed before arg type check, which assumes too much about the underlying implementation.

t/129-ssl-socket.t
```diff
@@ -2528,7 +2600,318 @@ SSL reused session
...
+            f = assert(io.open('$::HtmlDir/mtls_client.key'))
+            local key_data = f:read('*a')
+            f:close()
+
+            local ssl = require('ngx.ssl')
+
+            local chain = assert(ssl.parse_pem_cert(cert_data))
+            local priv = assert(ssl.parse_pem_priv_key(key_data))
+
+            assert(sock:sslhandshake(nil, nil, nil, nil, { client_cert = chain, client_priv_key = priv, }))
```
agentzh: I think we should introduce a separate method instead to avoid the requirement of an extra table here. Also it makes connection pool reuse easier.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1695,6 +1701,100 @@ ngx_http_lua_socket_tcp_sslhandshake(lua_State *L)
...
                     }
                 }
+
+                if (n >= 6) {
+                    if (lua_type(L, 6) == LUA_TTABLE) {
+                        lua_getfield(L, 6, "client_cert");
+
+                        if (!lua_isnil(L, -1)) {
+#ifdef OPENRESTY_LUAJIT
+                            chain = luaL_checkcdataptr(L, -1);
```
agentzh: I think maybe it's the time to rewrite this `sslhandshake` method with FFI.

agentzh: Oh, if we introduce a separate method to set the client certificates and keys, then we don't need to touch this method at all and just need to introduce a new FFI-based method instead.

t/129-ssl-socket.t
dndx: Do you mean to implement something that is the superset of `sock:sslhandshake` just like `settimeout` and `settimeouts`? For example:

```
sock:tlshandshake({
    client_cert = chain,
    client_priv_key = priv,
    server_name = "example.com",
    ssl_verify = true,
})
```

Personally I like this design, but wasn't sure at first so went for the more conservative approach.

src/ngx_http_lua_socket_tcp.c
dndx: Yes if we are decided on adding a brand new API then I'm all in for implementing it using FFI.

t/129-ssl-socket.t
dndx: Second thought. I think we can get rid of the `sock:sslhandshake` C function altogether and instead implement it (for backward compatibility) in lua-resty-core and simply calls `sock:tlshandshake`. This way we do not need to maintain two sets of logic for the same thing.

dndx:
Superseded by https://github.com/openresty/lua-nginx-module/pull/1602.

PR #1600, opened 2019-09-17
feature: implemented dynamic keepalive connections pooling in 'balancer_by_lua*'. 

thibaultcha:
#### Summary

This PR implements support for dynamic connection pools in NGINX's `upstream {}` blocks, similarly to NGINX's [keepalive](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) directive, controlled by Lua APIs.

#### Background

This need for this feature was originally driven by NGINX limitations around upstream keepalive pools: when using the ngx_http_upstream_module's `keepalive` directive, connections are reused solely based on their remote IP and remote port attributes, ignoring, for example, TLS attributes. The following tickets describe those limitations in more details, with example configurations to reproduce them:

* https://trac.nginx.org/nginx/ticket/1340
* https://trac.nginx.org/nginx/ticket/1593

These limitations can be addressed by defining several `upstream {}` blocks, or by disabling upstream keepalive connections pooling altogether.

A PoC patch was developed last year to address this issue in Kong (https://github.com/Kong/openresty-patches/blob/be8b79bd3a97cd8d1d9fce2da70a2be0a57af7d6/patches/1.13.6.1/nginx-1.13.6_01-upstream-keepalive.patch) and has proven itself successful in fixing the limitations: https://github.com/Kong/openresty-patches/pull/40.

A subsequent and more complete patch was recently submitted to the nginx-devel mailing list, implementing a way to dynamically allocate upstream connection pools: http://mailman.nginx.org/pipermail/nginx-devel/2019-August/012580.html. Alas (and unsurprisingly), the patch was not welcomed.

#### Proposed API

Since it is typical for an OpenResty application to define a limited number of `upstream {}` blocks that are shared between upstream services, this limitation should be overcome in OpenResty itself. The decision of implementing the connection pooling in this module has a few major benefits:

1. It avoids having to write and maintain another NGINX core patch.
2. It allows for dynamic and granular control over the upstream connection pools, which was not possible with the `keepalive` directive.
3. It makes the Lua balancer module better integrate with other stock NGINX upstream modules.
4. It allows for specifying an indefinite idle timeout and max requests unlike the NGINX upstream keepalive module since 1.15.3.

Example usage:

```lua
balancer_by_lua_block {

    local ok, err = balancer.set_current_peer("127.0.0.2", 8080)
    if not ok then
        ngx.log(ngx.ERR, "failed to set current peer: ", err)
        return
    end

    -- default pool will be "host:port"
    -- default pool_size will be 30
    ok, err = balancer.enable_keepalive()
    if not ok then
        ngx.log(ngx.ERR, "failed to set keepalive: ", err)
        return
    end
}
```

#### Documentation

The documentation is available in the lua-resty-core sister PR: https://github.com/openresty/lua-resty-core/pull/276

#### Additional commits

This PR also contains a couple of other improvements:

* It removes the need for defining a stub `server` directive in the `upstream {}` block.
* It refactors the stashing of the balancer peer data structure to be more efficient and more intuitive (i.e. the pattern to retrieve the peer data is identical in the upstream keepalive module interface and in the FFI interface: `r->upstream->peer.data`).

#### Sister PRs

* https://github.com/openresty/lua-resty-core/pull/276

---

### DIFF
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -16,46 +16,102 @@
 #include "ngx_http_lua_directive.h"
 
 
+typedef struct {
+    ngx_uint_t                               size;
+    ngx_uint_t                               connections;
+
+    uint32_t                                 crc32;
+
+    lua_State                               *lua_vm;
+
+    ngx_queue_t                              cache;
+    ngx_queue_t                              free;
+} ngx_http_lua_balancer_keepalive_pool_t;
+
+
+typedef struct {
+    ngx_queue_t                              queue;
+    ngx_connection_t                        *connection;
+
+    ngx_http_lua_balancer_keepalive_pool_t  *cpool;
+} ngx_http_lua_balancer_keepalive_item_t;
+
+
 struct ngx_http_lua_balancer_peer_data_s {
-    /* the round robin data must be first */
-    ngx_http_upstream_rr_peer_data_t    rrp;
+    ngx_uint_t                              cpool_size;
+    ngx_uint_t                              keepalive_requests;
+    ngx_msec_t                              keepalive_timeout;
 
-    ngx_http_lua_srv_conf_t            *conf;
-    ngx_http_request_t                 *request;
+    ngx_uint_t                              more_tries;
+    ngx_uint_t                              total_tries;
 
-    ngx_uint_t                          more_tries;
-    ngx_uint_t                          total_tries;
+    int                                     last_peer_state;
 
-    struct sockaddr                    *sockaddr;
-    socklen_t                           socklen;
+    uint32_t                                cpool_crc32;
 
-    ngx_str_t                          *host;
-    in_port_t                           port;
+    void                                   *data;
 
-    int                                 last_peer_state;
+    ngx_event_get_peer_pt                   original_get_peer;
+    ngx_event_free_peer_pt                  original_free_peer;
+
+#if (NGX_HTTP_SSL)
+    ngx_event_set_peer_session_pt           original_set_session;
+    ngx_event_save_peer_session_pt          original_save_session;
+#endif
+
+    ngx_http_request_t                     *request;
+    ngx_http_lua_srv_conf_t                *conf;
+    ngx_http_lua_balancer_keepalive_pool_t *cpool;
+
+    ngx_str_t                              *host;
+
+    struct sockaddr                        *sockaddr;
+    socklen_t                               socklen;
+
+    unsigned                                keepalive:1;
 
 #if !(HAVE_NGX_UPSTREAM_TIMEOUT_FIELDS)
-    unsigned                            cloned_upstream_conf;  /* :1 */
+    unsigned                                cloned_upstream_conf:1;
 #endif
 };
 
 
-#if (NGX_HTTP_SSL)
-static ngx_int_t ngx_http_lua_balancer_set_session(ngx_peer_connection_t *pc,
-    void *data);
-static void ngx_http_lua_balancer_save_session(ngx_peer_connection_t *pc,
-    void *data);
-#endif
+static ngx_int_t ngx_http_lua_balancer_by_chunk(lua_State *L,
+    ngx_http_request_t *r);
 static ngx_int_t ngx_http_lua_balancer_init(ngx_conf_t *cf,
     ngx_http_upstream_srv_conf_t *us);
 static ngx_int_t ngx_http_lua_balancer_init_peer(ngx_http_request_t *r,
     ngx_http_upstream_srv_conf_t *us);
 static ngx_int_t ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc,
     void *data);
-static ngx_int_t ngx_http_lua_balancer_by_chunk(lua_State *L,
-    ngx_http_request_t *r);
 static void ngx_http_lua_balancer_free_peer(ngx_peer_connection_t *pc,
     void *data, ngx_uint_t state);
+static ngx_int_t ngx_http_lua_balancer_create_keepalive_pool(lua_State *L,
+    ngx_log_t *log, uint32_t cpool_crc32, ngx_uint_t cpool_size,
+    ngx_http_lua_balancer_keepalive_pool_t **cpool);
+static void ngx_http_lua_balancer_get_keepalive_pool(lua_State *L,
+    uint32_t cpool_crc32, ngx_http_lua_balancer_keepalive_pool_t **cpool);
+static void ngx_http_lua_balancer_free_keepalive_pool(ngx_log_t *log,
+    ngx_http_lua_balancer_keepalive_pool_t *cpool);
+static void ngx_http_lua_balancer_close(ngx_connection_t *c);
+static void ngx_http_lua_balancer_dummy_handler(ngx_event_t *ev);
+static void ngx_http_lua_balancer_close_handler(ngx_event_t *ev);
+#if (NGX_HTTP_SSL)
+static ngx_int_t ngx_http_lua_balancer_set_session(ngx_peer_connection_t *pc,
+    void *data);
+static void ngx_http_lua_balancer_save_session(ngx_peer_connection_t *pc,
+    void *data);
+#endif
+
+
+#define ngx_http_lua_balancer_keepalive_is_enabled(bp)                       \
+    (bp->keepalive)
+
+#define ngx_http_lua_balancer_peer_set(bp)                                   \
+    (bp->sockaddr && bp->socklen)
+
+
+static char ngx_http_lua_balancer_keepalive_pools_table_key;
 
 
 ngx_int_t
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_balancer.c
```diff
@@ -286,21 +440,24 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
 
     ctx->context = NGX_HTTP_LUA_CONTEXT_BALANCER;
 
+    bp->kpool = NULL;
```
agentzh: "kpool" may not be a good name. How about use "mpool" for "memory pool" and "cpool" for "connection pool" consistently? The latter names can be used NOT just for this module.

src/ngx_http_lua_balancer.c
```diff
@@ -322,105 +479,426 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
-        pc->name = bp->host;
-
-        bp->rrp.peers->single = 0;
 
         if (bp->more_tries) {
             r->upstream->peer.tries += bp->more_tries;
         }
 
-        dd("tries: %d", (int) r->upstream->peer.tries);
+        if (ngx_http_lua_balancer_keepalive_enabled(bp)) {
```
agentzh: Better use `_is_enabled` instead? It'll be clearer.

src/ngx_http_lua_balancer.c
```diff
@@ -322,105 +479,426 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+            ngx_http_lua_balancer_get_keepalive_pool(L, bp->kpool_crc32,
+                                                     &bp->kpool);
 
-        return NGX_OK;
-    }
+            if (bp->kpool == NULL
+                && ngx_http_lua_balancer_create_keepalive_pool(L, pc->log,
```
agentzh: How do you limit the number of pools created? I'm worried about too many connection pools getting created without bound. The variants of crc32 can be huge. How about doing a search with crc32 just with the existing keepalive upstream module?

src/ngx_http_lua_common.h
```diff
@@ -298,10 +291,13 @@ union ngx_http_lua_srv_conf_u {
...
 
     struct {
-        ngx_http_lua_srv_conf_handler_pt  handler;
-        ngx_str_t           src;
-        u_char             *src_key;
-        int                 src_ref;
+        ngx_http_lua_srv_conf_handler_pt     handler;
+        ngx_str_t                            src;
+        u_char                              *src_key;
+        int                                  src_ref;
```
agentzh: Be careful about struct field memory layout. This `int` creates a 4-byte whole in the middle of this struct.

src/ngx_http_lua_balancer.c
```diff
@@ -16,46 +16,103 @@
 #include "ngx_http_lua_directive.h"
 
 
+typedef struct {
+    lua_State                               *lua_vm;
+
+    uint32_t                                 crc32;
```
agentzh: This creates a 4-byte whole in the struct layout. Please check other similar places.

src/ngx_http_lua_common.h
agentzh: We usually sort the fields by size by convention.

thibaultcha: I am aware, but this simply follows the layout of the above struct.

thibaultcha: From my observations, many structs have similar layout flaws, so I assumed we stopped paying attention to those. Easy to update back :+1:

src/ngx_http_lua_balancer.c
thibaultcha: > How do you limit the number of pools created? 

In future PRs, I was planning on adding ways to retrieve metrics from the pools (e.g. number of active pools, total memory allocated to existing pools, shutting down existing pools, retrieve the number of connections by pool, `getreusedtimes()` per connection, etc...). It is then up to users to manage their upstream keepalive pools as desired, according to their own constraints.
Limiting the max number of pools via a simple number feels very limiting in comparison. We could introduce a new NGINX directive, but those are also extremely limiting since they greatly limit the potential for dynamicity in OpenResty-based applications. If anything, such a "limit" should be applicable dynamically imho (i.e. via a Lua API). But again, letting users interact with existing pools and decide for themselves how many pools they should maintain is much better imho.     

> How about doing a search with crc32 just with the existing keepalive upstream module?

I'm not sure I get what you mean here. We don't have access to the inner keepalive upstream module without an NGINX patch, which is what this PR aims at avoiding. Besides, the NGINX keepalive upstream module has many other limitations which resulted in this PR being created in the first place.

thibaultcha: It is worth noting that even our cosocket connection pooling does not implement such a limit on the total number of pools.

agentzh: I'm afraid it would still be tricky to do LRU by the user herself unless we also provide an API to return the least recently used pool name. Besides, doing a lot of metrics bookkeeping is extra overhead at runtime and extra complexity in the code. I also have mixed feeling about that.

agentzh: The OpenResty way of doing things is to impose default limits to avoid shooting oneself on the foot. We should at least have something.

thibaultcha: Granted, but how do today's users of the cosocket connection pooling API deal with this? They don't...

rainingmaster:
@agentzh @thibaultcha Hello, How is this PR going? I think this function is great, allowing us to use only one upstream block, but manage our own connection pool according to the requirements. And implement the scenario of multiple upstream blocks in Nginx.

If we are worried about the total connection pool or the number of connections, can we make a configuration limit like [`lua_max_pending_timers`](https://github.com/openresty/lua-nginx-module#lua_max_pending_timers)?

In addition, I think that shutting down a pool by name and adjusting the size of the connection pool are also very important tools, let user manage all connection pool.

src/ngx_http_lua_balancer.c
agentzh: @thibaultcha currently it's a single pool so there's no issue of pool leaking.

agentzh: For the first version, what's the current strategy (or reality) of pool lifetime management?

agentzh: Because it's common to add unbounded number of dynamic peers in the ngx_lua world. An unbounded number of pools created is a big concern.

agentzh:
@rainingmaster We don't need multiple `upstream {}` blocks for long. That's the very reason to introduce `balancer_by_lua*` in the first place. The biggest limitation right now is we only respect socket addresses when reusing a connection, which can be addressed relatively easily by patching or forking the existing `ngx_http_upstream_keepalive` module. This PR tries to do a lot more, which is a bit overwhelming (at least to me).

@thibaultcha Shall we take smaller steps so that it's easier to review and also safer to incorporate? I appreciate your work here. Just want to be cautious and conservative for such relatively big things.

src/ngx_http_lua_balancer.c
```diff
@@ -16,46 +16,103 @@
...
+    ngx_str_t                              *host;
+
+    struct sockaddr                        *sockaddr;
+    socklen_t                               socklen;
+
+    unsigned                                keepalive:1;
 
 #if !(HAVE_NGX_UPSTREAM_TIMEOUT_FIELDS)
-    unsigned                            cloned_upstream_conf;  /* :1 */
+    unsigned                                cloned_upstream_conf;  /* :1 */
```
agentzh, line 74: We should use bit fields here since they are after the bit field above (`keepalive`).

src/ngx_http_lua_balancer.c
thibaultcha: What I meant is that today, there is no `lua_socket_max_pools` directive imposing an upper bound on cosocket connection pools either, and, unless I am mistaken, we have not heard of prominent concerns in that regard. That said, I could add something to this PR if that would alleviate your concerns.

agentzh:
OK, I had a closer look. It looks good generally and we can have it if those remaining concerns are addressed.

I have the following questions or concerns:

1. we need some kind of upper limit for the total number of connection pools.
2. will the connection pools get automatically freed when there's no idle connections in them?
3. what happens when there is a CRC32 collision? Is it possible that we misuse a connection? It's better if we also compare the socket address and etc to make sure. CRC32 is easy to get a collision.
4. I don't think infinite max idle time is a good idea...
5. Is the default pool size (30) configurable via some nginx directive?

thibaultcha:
@agentzh Ah, thanks! I split the commits up in different chunks, separated by features, so reviewing them individually will be easier. Now this PR is quite a bit old so I need to re-familiarize myself with it, but I recall reviewing it very extensively and feeling like it would overall greatly improve the realm of possibilities for dynamic load balancing in OpenResty rather than keep relying on the Nginx keepalive module, hence why the decision to implement our own keepalive module. It gives us extra flexibility for future use-cases of connection pools management, monitoring, etc...

> 1. we need some kind of upper limit for the total number of connection pools

I can certainly add this, but I then wonder how come the cosocket API not suffer the same flaw then? `ngx_http_lua_socket_tcp_create_socket_pool` is not bound to any upper limit.

> 2. will the connection pools get automatically freed when there's no idle connections in them?

Yes, see https://github.com/openresty/lua-nginx-module/pull/1600/files#diff-998ae59f12e2d4f8f745561e5c824912R900-R902 and the test cases, e.g. this one: https://github.com/openresty/lua-resty-core/pull/276/files#diff-de1e53807cbda85df1ae20d983bd76efR805

> 3. what happens when there is a CRC32 collision?

Currently the pools are stored in a Lua table stored in the Lua registry. I believe this is because I followed the approach taken in the cosocket implementation (see https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L360).

However, a better implementation of this would be using the Lua registry [reference system](https://www.lua.org/pil/27.3.2.html) instead (we have already been using it successfully since refactoring the Lua code cache system in https://github.com/openresty/lua-nginx-module/commit/39c2ead3cb7efae4456494ea27a348fa1eca6c5c). A Lua-land table stored as an upvalue in `lib/ngx/balancer.lua` can then be a table whose keys are the pool names (as strings), and values are their associated reference to their Lua registry C pointers counterparts. This way, we even avoid having to compute _any_ CRC32 hashes in the first place too.

> 4. I don't think infinite max idle time is a good idea...

An indefinite max idle time has been the default and the widely used value for a long time, until the release of Nginx 1.15.3 in 2018. [This commit](https://github.com/nginx/nginx/commit/7de808990b8e49e2e18c2a013e4cca82a092fdbc) whose intent is to fix a tangentially related issue, caused a feature regression here (and as I recall, Nginx was never actively closing its upstream connections before). While having a default upstream keepalive timeout of 60s is a good idea, the fix should also not prevent any potentially previously developed systems that may have been taking advantage of Nginx never actively closing upstream connections.
As such, this PR stills proposes a default timeout of 60s, a max number of requests of 100 (https://github.com/openresty/lua-resty-core/pull/276/files#diff-973c4ef5939470d0a1c529e4e692db8eR106-R108), but allows for respecting the pre- Nginx 1.15.3 behavior of not actively closing connections.

Besides, the [tcpsock:setkeepalive API](https://github.com/openresty/lua-nginx-module#tcpsocksetkeepalive) also allows for an indefinite idle timeout; this PR intentionally unifies the user experience of the balancer and cosocket network APIs. 

> 5. Is the default pool size (30) configurable via some nginx directive?

It currently is a hard-coded Lua-land constant: https://github.com/openresty/lua-resty-core/pull/276/files#diff-973c4ef5939470d0a1c529e4e692db8eR106. Since the accent was put on making OpenResty more _dynamic_ in this PR, I have been reticent at introducing new Nginx directives...

The reasoning being that either:
- one does not use the `balancer.enable_keepalive()` API, and isn't concerned with this default.
- one uses the `balancer.enable_keepalive()` API, and if so may change the defaults by simply adding a new `opts` table to their `balancer.set_current_peer()` calls, in an area of their codebase related to the balancing code.
Overall, I believe that configuring Nginx should be done in `nginx.conf`, and configuring OpenResty should be mostly done in Lua code, thus reducing the "sovereignty" that Nginx patterns have over OpenResty, and allowing developers to rely less and less over a static `nginx.conf` overtime... But if that is not the decision you wish to take going forward, I _can_ add a directive here, it is a short and easy task indeed. My concern is more philosophical :)

rainingmaster:
@doujiang24 @spacewander hihi, could you help to have a look together. And I think we can support the connection limit to upstream(like backlog in ngx.tcp.socket) base on this feature in future, which in not support in open source nginx.

src/ngx_http_lua_balancer.c
```diff
@@ -322,79 +480,87 @@ ngx_http_lua_balancer_get_peer(ngx_peer_connection_t *pc, void *data)
...
+            {
+                return NGX_ERROR;
+            }
 
-    return ngx_http_upstream_get_round_robin_peer(pc, &bp->rrp);
-}
+            ngx_http_lua_assert(bp->cpool);
 
+            if (!ngx_queue_empty(&bp->cpool->cache)) {
+                q = ngx_queue_head(&bp->cpool->cache);
```
spacewander, line 511: Look like we need to get the cached from the tail? Since we always insert the object to the head, the one from the head is the one just inserted.

src/ngx_http_lua_balancer.c
```diff
@@ -178,11 +292,42 @@ ngx_http_lua_balancer_by_lua(ngx_conf_t *cf, ngx_command_t *cmd,
...
+    if (uscf->servers->nelts == 0) {
+        us = ngx_array_push(uscf->servers);
+        if (us == NULL) {
+            return NGX_CONF_ERROR;
+        }
+
+        ngx_memzero(us, sizeof(ngx_http_upstream_server_t));
+        ngx_memzero(&url, sizeof(ngx_url_t));
+
+        ngx_str_set(&url.url, "0.0.0.1");
```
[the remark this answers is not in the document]
thibaultcha: Because it is the address used here: https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/balancer.md#http-subsystem.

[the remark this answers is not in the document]
spacewander: @thibaultcha 
Aha, I see. We need to add a comment to explain it in code.

gung-rgb:
> > 3. what happens when there is a CRC32 collision?
> 
> Currently the pools are stored in a Lua table stored in the Lua registry. I believe this is because I followed the approach taken in the cosocket implementation (see https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_socket_tcp.c#L360).
> 
> However, a better implementation of this would be using the Lua registry [reference system](https://www.lua.org/pil/27.3.2.html) instead (we have already been using it successfully since refactoring the Lua code cache system in [39c2ead](https://github.com/openresty/lua-nginx-module/commit/39c2ead3cb7efae4456494ea27a348fa1eca6c5c)). A Lua-land table stored as an upvalue in `lib/ngx/balancer.lua` can then be a table whose keys are the pool names (as strings), and values are their associated reference to their Lua registry C pointers counterparts. This way, we even avoid having to compute _any_ CRC32 hashes in the first place too.

Hi, @thibaultcha, Is that any change mentions here have apply to kong's openresty patchs? I use kong 2.2.2, suffer a CRC32 collision in my envirentment.  Details are [here](https://github.com/Kong/kong/issues/9582).

fularac:
This can probably be closed now as it was merged with #1908

zhuizhuhaomeng:
Thanks for your reminder.

Issue #1620, opened 2019-11-19
Feature: Polling on multiple cosockets


GuyLewin:
I believe the ability to poll on multiple cosockets (using the Nginx event loop obviously) while providing a timeout (similar API to socket.select()) would be very useful for creating full-duplex connections (such as WebSockets) or raw TCP proxying.

Is there any substantial reason this API shouldn't be implemented? I'm asking to make sure before I try to implement it myself.

PR #1630, opened 2019-12-29
Feature/tcp socket select

GuyLewin:

Implements #1620


GuyLewin:
@spacewander do you mind looking at this too? This PR is a lot more complex than the other, but I believe it's a *very* important feature for us to have (for example in proxying web sockets or other TCP sockets).

spacewander:
Does this new method have any advantage over running multiple coroutines (created by `ngx.thread.spawn`, each one contains their own sockets) and sync with `ngx.semaphore`? The latter one is already battle-tested.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2927,6 +3086,83 @@ ngx_http_lua_socket_tcp_send_retval_handler(ngx_http_request_t *r,
...
+        }
+
+        lua_pushnumber(L, matched_index + 1);
+        (void) ngx_http_lua_socket_read_error_retval_handler(r, u, L);
+        lua_remove(L, -2);
+        return 2;
+    }
+
+    lua_pushnumber(L, matched_index + 1);
+    return 1;
```
spacewander: I am wonder if returning only one readable socket is OK... Please correct me if I am wrong:
At least with epoll event module, the read event is triggered only once at the edge (search EPOLLET as the keyword). If multiple events happen at the same time, and only one event is handled, the other shots will be lost.

GuyLewin:
@spacewander thanks for going over this! Can you give an example for Lua code you'd suggest to replace this functionality? Obviously I'd want this code to run without extra threads, but all based on the epoll() (or any module)'s multiplexing logic.

src/ngx_http_lua_socket_tcp.c
GuyLewin: You're definitely right. I'm not sure how conceptually I could do that (combine all the events from the same epoll into 1 return code) but I'll look into that.
I believe that even if a solution of returning multiple FDs doesn't work out - this select() API is very relevant, as you can call select() again after you just called it and get the remaining FD(s).
I'll look into it and update.

spacewander:
@GuyLewin 
You can use the example in `ngx.semaphore` docs as a start:
https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/semaphore.md

Personally speaking, I think the coroutines could be cheap enough. And implementing a new layer of I/O multiplexing (in the Lua code) above the other one (in the Nginx) may be too complex and inflexible. Somebody has used epoll to reduce memory usage in Go: https://www.freecodecamp.org/news/million-websockets-and-go-cc58418460bb, but they called the epoll directly so this is a different story.

zhuizhuhaomeng:
I don't want to merge this feature. I believe the existing features are already sufficient to meet the requirements.

PR #1635, opened 2020-01-29
feature: implement keepalive socket context

bjne:
A new option is added to the tcp.connect options_table:

ctx = boolean, where a "true" value would change the value returned
from connect from current ok, err to:
```lua
local socket_ctx, reused = sock:connect('127.0.0.1', 5432, {ctx = true})
if not socket_ctx then
    return nil, reused
end

if reused == 0 then
    setmetatable(socket_ctx, some_mt)
end

socket_ctx.sock = sock

socket_ctx:execute("select $1, $2", 1, 2)

if reused < 100 then
    sock:setkeepalive()
end
```

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -529,6 +529,49 @@ ngx_http_lua_socket_tcp_create_socket_pool(lua_State *L, ngx_http_request_t *r,
 }
 
 
+static void
+ngx_http_lua_socket_tcp_ctx_unref(lua_State *L, int *ctx_ref)
+{
+    if (*ctx_ref != LUA_NOREF && *ctx_ref != LUA_REFNIL) {
+        ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
+                    "lua unref socket ctx");
+
+        lua_pushliteral(L, ngx_http_lua_tcp_ctx_tables_key);
+        lua_rawget(L, LUA_REGISTRYINDEX);
+        luaL_unref(L, -1, *ctx_ref);
+        lua_pop(L, 1);
+        *ctx_ref = LUA_NOREF;
+    }
+}
+
+
+static int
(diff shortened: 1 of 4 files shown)


bjne:
this is the basic idea for usage:
```lua
local pg = {
        login = function() return true end,
	prepare = function(self, query)
		local portal = (self.nportals or 0) + 1

		local ok, err = ngx.say("prepare ", portal, " ", query)
		if not ok then
			return nil, err
		end

		self.nportals, self[query] = portal, portal

		return portal
	end,
	execute = function(self, query, ...)
		if not query then
			return nil, "no query defined"
		end

		local portal, err = self[query]

		if not portal then
			portal, err = self:prepare(query)
			if not portal then
				return nil, err
			end
		end

		ngx.say("execute ", portal, " ", table.concat({...}, ", "))
	end
}

local pg_mt = { __index = pg }

local conn_opts = { ctx = true }

return {
	go = function()
		local query = ngx.unescape_uri(ngx.var.arg_query)
		if not query then
			return nil, "query must be spesified"
		end

		local sock = ngx.socket.tcp()
		local sock_ctx, reused = sock:connect('127.0.0.1', 1234, conn_opts)
		if not sock_ctx then
			return nil, reused
		end

		if reused == 0 then
			setmetatable(sock_ctx, pg_mt)
                        if not sock_ctx:login() then
                            return nil, "login failed"
                        end
		end

		sock_ctx.sock = sock

		sock_ctx:execute(query,  reused, sock_ctx.nportals)

		sock:setkeepalive()
	end
}
```

spacewander:
Three questions:
1. I am not sure if it's a good idea to change the return values from `ok, err` to `ctx, reused|err`.
2. Maybe we need a new API to create ctx on demand, just like `ngx.ctx`.
3. More test cases, please.

@thibaultcha, what is your opinion?

zhuizhuhaomeng:
You should use the sock:getreusedtimes() API instead of changing the return values of the API.

PR #1654, opened 2020-02-24
bugfix: ensured arguments of APIs mutating uri or request/response headers do not contain unsafe characters.

doujiang24:

### DIFF
--- a/src/ngx_http_lua_headers_in.c
+++ b/src/ngx_http_lua_headers_in.c
@@ -658,6 +658,12 @@ ngx_http_lua_set_input_header(ngx_http_request_t *r, ngx_str_t key,
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
+    if (ngx_http_lua_check_header_safe(r, key.data, key.len) != NGX_OK
+        || ngx_http_lua_check_header_safe(r, value.data, value.len) != NGX_OK)
+    {
+        return NGX_ERROR;
+    }
+
     hv.hash = ngx_hash_key_lc(key.data, key.len);
     hv.key = key;
 
--- a/src/ngx_http_lua_uri.c
+++ b/src/ngx_http_lua_uri.c
@@ -16,6 +16,8 @@
 
 
 static int ngx_http_lua_ngx_req_set_uri(lua_State *L);
+static ngx_int_t ngx_http_lua_check_uri_safe(ngx_http_request_t *r,
+    u_char *str, size_t len);
 
 
 void
@@ -53,6 +55,10 @@ ngx_http_lua_ngx_req_set_uri(lua_State *L)
 
     if (len == 0) {
         return luaL_error(L, "attempt to use zero-length uri");
+    }
+
+    if (ngx_http_lua_check_uri_safe(r, p, len) != NGX_OK) {
+        return luaL_error(L, "attempt to use unsafe uri");
     }
 
     if (n == 2) {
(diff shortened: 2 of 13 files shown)


src/ngx_http_lua_uri.c
```diff
@@ -107,4 +110,38 @@ ngx_http_lua_ngx_req_set_uri(lua_State *L)
...
+
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000  /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+    };
+
+    for (i = 0; i < len; i++, str++) {
+        if (unsafe[*str >> 5] & (1 << (*str & 0x1f))) {
+            return i;
```
agentzh: I think it's safer to return `(size_t) -1` here to indicate an error instead of do blind truncation. Blind truncation may still have security implications.

src/ngx_http_lua_headers_in.c
```diff
@@ -658,6 +658,10 @@ ngx_http_lua_set_input_header(ngx_http_request_t *r, ngx_str_t key,
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
+    key.len = ngx_http_lua_safe_header_value_len(key.data, key.len);
+
+    value.len = ngx_http_lua_safe_header_value_len(value.data, value.len);
```
agentzh: Let's check if the return value is `(size_t) -1`. If yes, then return `NGX_ERROR` in this function.

src/ngx_http_lua_uri.c
```diff
@@ -55,6 +56,8 @@ ngx_http_lua_ngx_req_set_uri(lua_State *L)
         return luaL_error(L, "attempt to use zero-length uri");
     }
 
+    len = ngx_http_lua_safe_uri_len(p, len);
```
agentzh: Ditto. Let's check the return value if there is unsafe bytes, we should throw out a Lua exception.

src/ngx_http_lua_uri.c
```diff
@@ -107,4 +110,38 @@ ngx_http_lua_ngx_req_set_uri(lua_State *L)
...
+        0x80000000, /* 1000 0000 0000 0000  0000 0000 0000 0000 */
+
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000  /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+    };
+
+    for (i = 0; i < len; i++, str++) {
+        if (unsafe[*str >> 5] & (1 << (*str & 0x1f))) {
```
agentzh: For the best performance, let's cache the value of the expression `*str` into a local C variable like `c` here.

src/ngx_http_lua_util.h
```diff
@@ -247,8 +247,25 @@ ngx_http_lua_safe_header_value_len(u_char *str, size_t len)
...
+
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000  /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+    };
+
     for (i = 0; i < len; i++, str++) {
-        if (*str == '\r' || *str == '\n') {
+        if (unsafe[*str >> 5] & (1 << (*str & 0x1f))) {
```
agentzh: Ditto. Let's cache the value of `*str`.

src/ngx_http_lua_uri.c
agentzh: Also need debug logging here to show the offensive byte.

src/ngx_http_lua_util.h
```diff
@@ -241,20 +241,9 @@ void ngx_http_lua_cleanup_free(ngx_http_request_t *r,
...
-    for (i = 0; i < len; i++, str++) {
-        if (*str == '\r' || *str == '\n') {
-            return i;
-        }
-    }
-
-    return len;
-}
+size_t ngx_http_lua_log_escape(u_char *dst, u_char *src, size_t size);
+ngx_int_t ngx_http_lua_check_header_safe(ngx_http_request_t *r, u_char *str,
```
thibaultcha: Should we ensure this function is inlined just like its predecessor?

src/ngx_http_lua_util.c
```diff
@@ -4261,4 +4261,121 @@ ngx_http_lua_set_sa_restart(ngx_log_t *log)
...
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
+        0x80000000, /* 1000 0000 0000 0000  0000 0000 0000 0000 */
+
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000  /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+    };
```
thibaultcha: There are more unsafe characters for headers that aren't marked as such here (ditto for the set of unsafe URI characters). Are we only concerned about the control characters marked as unsafe here or do we want to add more?

thibaultcha:
(Rebased on top of master due to a newly failing tests caused by changes in openresty.org's accepter ciphers)

src/ngx_http_lua_util.h
doujiang24: I'm not quite sure here, I think the new function is a bit long for `inline`.

thibaultcha: It isn't longer than `ngx_http_lua_check_uri_safe`, which is marked inlined itself though.

thibaultcha:
Just pushed an additional commit with more fixes.

src/ngx_http_lua_util.c
```diff
@@ -4324,7 +4324,7 @@ ngx_http_lua_log_escape(u_char *dst, u_char *src, size_t size)
         size--;
     }
 
-    return NGX_OK;
+    return 0;
```
thibaultcha: imho, this is a more correct return value given the `size_t` return type of this function.

src/ngx_http_lua_util.h
agentzh: Marking `inline` is fine.

PR #1666, opened 2020-03-13
feature: add FFI interface to verify SSL client certificate

ArchangelSDY:
This PR adds a FFI interface that allows users to configure SSL client certificate verification dynamically. For example, Nginx can now read SNI first and then determine whether to request a client certificate during handshake. Optionally, caller can pass in a CA list used for verification.

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -1311,6 +1311,7 @@ int
 ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
     int depth, char **err)
 {
+    ngx_http_lua_ctx_t          *ctx;
     ngx_ssl_conn_t              *ssl_conn;
     ngx_http_ssl_srv_conf_t     *sscf;
     STACK_OF(X509)              *chain = ca_certs;
@@ -1323,6 +1324,17 @@ ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
 #else
     int                         i;
 #endif
+
+    ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
+    if (ctx == NULL) {
+        *err = "no request ctx found";
+        return NGX_ERROR;
+    }
+
+    if (!(ctx->context & NGX_HTTP_LUA_CONTEXT_SSL_CERT)) {
+        *err = "API disabled in the current context";
+        return NGX_ERROR;
+    }
 
     if (r->connection == NULL || r->connection->ssl == NULL) {
         *err = "bad request";
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -1294,4 +1294,104 @@ failed:
 }
 
 
+static int
+ngx_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store)
+{
+    /*
+     * we never terminate handshake here and user can later use
+     * $ssl_client_verify to check verification result.
+     *
+     * this is consistent with Nginx behavior.
+     */
+    return 1;
+}
+
+
+int
+ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r,
+    int depth,
+    void *cdata, char **err)
+{
+    ngx_ssl_conn_t         *ssl_conn;
+    STACK_OF(X509)         *chain = cdata;
+    STACK_OF(X509_NAME)    *name_chain = NULL;
+    X509                   *x509 = NULL;
+    X509_NAME              *subject = NULL;
+    X509_STORE             *ca_store = NULL;
+#ifdef OPENSSL_IS_BORINGSSL
+    size_t                 i;
+#else
+    int                    i;
+#endif
+
+    if (r->connection == NULL || r->connection->ssl == NULL) {
+        *err = "bad request";
+        return NGX_ERROR;
+    }
+
+    ssl_conn = r->connection->ssl->connection;
+    if (ssl_conn == NULL) {
+        *err = "bad ssl conn";
+        return NGX_ERROR;
+    }
+
+    ca_store = SSL_CTX_get_cert_store(SSL_get_SSL_CTX(ssl_conn));
+    if (ca_store == NULL) {
+        *err = "SSL_CTX_get_cert_store() failed";
+        return NGX_ERROR;
+    }
+
+    SSL_set_verify(ssl_conn, SSL_VERIFY_PEER, ngx_ssl_verify_callback);
+
+    SSL_set_verify_depth(ssl_conn, depth);
+
+    if (chain != NULL) {
+        /* construct name chain */
+
+        name_chain = sk_X509_NAME_new_null();
+        if (name_chain == NULL) {
+            *err = "sk_X509_NAME_new_null() failed";
+            return NGX_ERROR;
+        }
+
+        for (i = 0; i < sk_X509_num(chain); i++) {
+            x509 = sk_X509_value(chain, i);
+            if (x509 == NULL) {
+                *err = "sk_X509_value() failed";
+                goto failed;
+            }
+
+            /* add subject to name chain, which will be sent to client */
+            subject = X509_NAME_dup(X509_get_subject_name(x509));
+            if (subject == NULL) {
+                *err = "X509_get_subject_name() failed";
+                goto failed;
+            }
+
+            if (!sk_X509_NAME_push(name_chain, subject)) {
+                *err = "sk_X509_NAME_push() failed";
+                X509_NAME_free(subject);
+                goto failed;
+            }
+
+            /* add to trusted CA store */
+            if (X509_STORE_add_cert(ca_store, x509) == 0) {
+                *err = "X509_STORE_add_cert() failed";
+                goto failed;
+            }
+        }
+
+        SSL_set_client_CA_list(ssl_conn, name_chain);
+    }
+
+    return NGX_OK;
+
+failed:
+
+    sk_X509_NAME_free(name_chain);
+
+    return NGX_ERROR;
+}
+
+
 #endif /* NGX_HTTP_SSL */
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -1295,7 +1295,7 @@ failed:
 
 
 static int
-ngx_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store)
+ngx_http_lua_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store)
 {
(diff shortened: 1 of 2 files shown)


doujiang24:
@ArchangelSDY will you create another PR in `lua-resty-core` to expose Lua API?

ArchangelSDY:
@doujiang24  Sure if you are fine with the API change in this PR. Here they are:

https://github.com/openresty/lua-resty-core/pull/289
https://github.com/openresty/meta-lua-nginx-module/pull/75
https://github.com/openresty/stream-lua-nginx-module/pull/190

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,104 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+        for (i = 0; i < sk_X509_num(chain); i++) {
+            x509 = sk_X509_value(chain, i);
+            if (x509 == NULL) {
+                *err = "sk_X509_value() failed";
+                goto failed;
+            }
+
+            /* add subject to name chain, which will be sent to client */
+            if ((subject = X509_NAME_dup(
+                    X509_get_subject_name(x509))) == NULL) {
```
doujiang24: style: we usually use two lines here.
```suggestion
    subject = X509_NAME_dup(X509_get_subject_name(x509))
    if (subject == NULL) {
```

t/140-ssl-c-api.t
```diff
@@ -12,7 +12,7 @@ if ($openssl_version =~ m/built with OpenSSL (0|1\.0\.(?:0|1[^\d]|2[a-d]).*)/) {
     plan(skip_all => "too old OpenSSL, need 1.0.2e, was $1");
 
 } else {
-    plan tests => repeat_each() * (blocks() * 5 + 1);
+    plan tests => repeat_each() * 38;
```
doujiang24: style: we'd better keep this style: `block() * 5 + N`.

ArchangelSDY: Sure. Reverted back. It took me a while to figure out what does `5` and `N` mean :)

doujiang24:
@ArchangelSDY we'd better use new commit for the new changes, force push is not friendly for the reviewer to check the new changes only, thanks :)

ArchangelSDY:
I see. Will follow that for new changes.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,104 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+     *
+     * this is consistent with Nginx behavior.
+     */
+    return 1;
+}
+
+
+int
+ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r,
+    int depth,
```
doujiang24, line 1312: style: this line can move to the suffix of the previous line.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,104 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+    void *cdata, char **err)
```
doujiang24, line 1313: `cdata` may not be a good name here, better use `ca_chain` instead?

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,116 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+
+    /* enable verify */
+
+    SSL_set_verify(ssl_conn, SSL_VERIFY_PEER, ngx_ssl_verify_callback);
+
+    /* set depth */
+
+    if (depth < 0) {
+        sscf = ngx_http_get_module_srv_conf(r, ngx_http_ssl_module);
+        if (sscf != NULL) {
```
doujiang24, line 1346: we'd better set the `depth` to 1 when `sscf` is `NULL`, and a comment: `same as the default value of ssl_verify_depth`.

ArchangelSDY: sure. updated.

rainingmaster:
@ArchangelSDY 
Hello, I see the [default cert](https://github.com/openresty/lua-nginx-module/pull/1666/files#diff-91e07e8def01efe511547241b8afaea8R861) is same with the [cert loaded in ffi api](https://github.com/openresty/lua-nginx-module/pull/1666/files#diff-91e07e8def01efe511547241b8afaea8R842), maybe we should change another cert as new CA cert? I also comment [here](https://github.com/openresty/lua-resty-core/pull/289#issuecomment-616292311)

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,119 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+     * $ssl_client_verify to check verification result.
+     *
+     * this is consistent with Nginx behavior.
+     */
+    return 1;
+}
+
+
+int
+ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
```
zhuizhuhaomeng, line 1311: if arg ca_certs is path of the cert, then the code would be must easizer.  just like ngx_ssl_client_certificate
why not use path ?

ArchangelSDY: Since `ssl.set_cert` accepts a pointer rather than a path, I think it should better to keep them consistent. Besides, it provides some flexibility that these certificates do not have to be on disk. For example, in [Kubernetes ingress controller](https://github.com/kubernetes/ingress-nginx) use case, certificates are kept in Nginx memory and are update via a REST API.

ArchangelSDY:
@rainingmaster  I see your concern. Then I will use a different cert at server side. Now both certs are self signed so the CA lists should not interfere with each other.

rainingmaster:
> @rainingmaster I see your concern. Then I will use a different cert at server side. Now both certs are self signed so the CA lists should not interfere with each other.

@ArchangelSDY Good job, and I think `Test 8` compare to `Test 6` should only difference in *client provides no certificate*, but should provide CA cert [here](https://github.com/openresty/lua-nginx-module/pull/1666/files#diff-91e07e8def01efe511547241b8afaea8R983), like `Test 7` should only different with `Test 6` in *CA certificates*?

Beside, `server_tokens` and `lua_ssl_trusted_certificate` should be useless in new test cases;

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1294,4 +1294,119 @@ ngx_http_lua_ffi_set_priv_key(ngx_http_request_t *r,
...
+    size_t                      i;
+#else
+    int                         i;
+#endif
+
+    if (r->connection == NULL || r->connection->ssl == NULL) {
+        *err = "bad request";
+        return NGX_ERROR;
+    }
+
```
rainingmaster, line 1331: As [the document describe](https://github.com/openresty/lua-resty-core/pull/289/files#diff-a69ae86a1de8bc9594d4d4f6dca89709R482), here should check the phase like [ngx.exit](https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_control.c#L363). And now the function can be run other phase such as `content_by_lua_block`.

ArchangelSDY: I see other SSL APIs don't have this check either. Is this simply because of some historical reasons?

ArchangelSDY:
> > @rainingmaster I see your concern. Then I will use a different cert at server side. Now both certs are self signed so the CA lists should not interfere with each other.
> 
> @ArchangelSDY Good job, and I think `Test 8` compare to `Test 6` should only difference in _client provides no certificate_, but should provide CA cert [here](https://github.com/openresty/lua-nginx-module/pull/1666/files#diff-91e07e8def01efe511547241b8afaea8R983), like `Test 7` should only different with `Test 6` in _CA certificates_?
> 
> Beside, `server_tokens` and `lua_ssl_trusted_certificate` should be useless in new test cases;

Make sense. Updated `Test 8` and removed those useless parts.

src/ngx_http_lua_ssl_certby.c
ArchangelSDY: Check added.

rainingmaster: > I see other SSL APIs don't have this check either. Is this simply because of some historical reasons?

Yes, I agree, but I think we should fix these inconsistencies(such as [set_der_cert](https://github.com/openresty/lua-resty-core/blob/master/lib/ngx/ssl.md#set_der_cert)) with the documentation and add some test cases, right? @doujiang24

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1367,7 +1367,7 @@ ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
     /* set CA chain */
 
     if (chain != NULL) {
-        ca_store = SSL_CTX_get_cert_store(SSL_get_SSL_CTX(ssl_conn));
+        ca_store = X509_STORE_new();
         if (ca_store == NULL) {
             *err = "SSL_CTX_get_cert_store() failed";
             return NGX_ERROR;
```
spacewander, line 1373: Remember to `goto failed` when failed to create `name_chain`:
https://github.com/openresty/lua-nginx-module/blob/6b58760fe4da41c895ab8dc4ddbd8c5feccc75f2/src/ngx_http_lua_ssl_certby.c#L1378-L1381

ArchangelSDY: Fixed.

rainingmaster:
@agentzh @doujiang24 hihi, since 1.17.8 is released, could we merge this PR now?

t/140-ssl-c-api.t
```diff
@@ -812,3 +816,206 @@ lua ssl server name: "test.com"
...
+            if r == nil then
+                ngx.log(ngx.ERR, "no request found")
+                return
+            end
+
+            local f = assert(io.open("t/cert/test.crt", "rb"))
+            local cert_data = f:read("*all")
+            f:close()
+
+            local cert = ffi.C.ngx_http_lua_ffi_parse_pem_cert(cert_data, #cert_data, errmsg)
```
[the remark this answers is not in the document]
ArchangelSDY: I see. Just fixed them. For the second leak I think it should be caused by SSL session reuse. After I set `proxy_ssl_session_reuse` to `off`, it looks good now.

doujiang24:
@ArchangelSDY Thanks for your contribution again.
The new OpenResty 1.19.3.1 release is coming, and we are preparing an online open meeting to introduce the new release.
Do you have time to have a talk about this new feature?
My email is doujiang24#gmail.com and my WeChat ID is doujiang24, welcome to contact me if you are interested in it.

PR #1668, opened 2020-03-18
fix FFI check when LuaJIT is linked as .so

fancy-rabbit:

----
otherwise the configure script cannot found libluajit-5.1.a and aborted.


thibaultcha:
This should now be fixed in master. Thanks!

PR #1680, opened 2020-04-01
travis-ci: bumped OpenSSL to 1.1.1f and remove older OpenSSL versions.

thibaultcha:
Older OpenSSL versions are not supported anymore
(see https://www.openssl.org/source/).


spacewander:
Although OpenSSL doesn't support older versions any more, OpenResty users may be still using them. Most of our machines are running on CentOS7, which use OpenSSL 1.0.2 by default. I think we should keep them in the test suite to prevent regression.

thibaultcha:
Fine with me. In which case we need to find a mirror because those versions are not available at openssl.org anymore.

PR #1693, opened 2020-04-17
doc: added description of memory fragmentation to the shdict

wanghuizzz:

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -6548,6 +6548,8 @@ The optional `flags` argument specifies a user flags value associated with the e
 
 When it fails to allocate memory for the current key-value item, then `set` will try removing existing items in the storage according to the Least-Recently Used (LRU) algorithm. Note that, LRU takes priority over expiration time here. If up to tens of existing items have been removed and the storage left is still insufficient (either due to the total capacity limit specified by [lua_shared_dict](#lua_shared_dict) or memory segmentation), then the `err` return value will be `no memory` and `success` will be `false`.
 
+If the sizes of items in the dictionary are not approximate, it is eaisier to encounter `no memory` error because of memory fragmentation. It is recommended to use different dictionaries for different sizes of items. When you encounter `no memory` error, you can also remove more existing items by retrying more times to make the operation successful.
+
 If this method succeeds in storing the current item by forcibly removing other not-yet-expired items in the dictionary via LRU, the `forcible` return value will be `true`. If it stores the item without forcibly removing other valid items, then the return value `forcible` will be `false`.
 
 The first argument to this method must be the dictionary object itself, for example,
--- a/README.markdown
+++ b/README.markdown
@@ -6550,7 +6550,7 @@ When it fails to allocate memory for the current key-value item, then `set` will
 
 If the sizes of items in the dictionary are not multiples or even powers of a certain value (like 2), it is easier to encounter `no memory` error because of memory fragmentation. It is recommended to use different dictionaries for different sizes of items.
 
-When you encounter `no memory` error, you can also remove more existing items by retrying more times to make the operation successful.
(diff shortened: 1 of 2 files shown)


README.markdown
```diff
@@ -6548,6 +6548,8 @@ The optional `flags` argument specifies a user flags value associated with the e
 
 When it fails to allocate memory for the current key-value item, then `set` will try removing existing items in the storage according to the Least-Recently Used (LRU) algorithm. Note that, LRU takes priority over expiration time here. If up to tens of existing items have been removed and the storage left is still insufficient (either due to the total capacity limit specified by [lua_shared_dict](#lua_shared_dict) or memory segmentation), then the `err` return value will be `no memory` and `success` will be `false`.
 
+If the sizes of values in the dictionary are not approximate, it is eaisier to encounter `no memory` error because of memory fragmentation. It is recommended to use different dictionaries for different size values. When you encounter `no memory` error, you can also retry a few times to make sure setting successfully.
```
doujiang24, line 6551: If the sizes of values => If the sizes of items.
different size values => different sizes of items.

doujiang24: you can also retry a few times to make sure setting successfully
=>
you can also remove more existing items by retrying more times to make the operation successful.

README.markdown
```diff
@@ -6548,6 +6548,8 @@ The optional `flags` argument specifies a user flags value associated with the e
 
 When it fails to allocate memory for the current key-value item, then `set` will try removing existing items in the storage according to the Least-Recently Used (LRU) algorithm. Note that, LRU takes priority over expiration time here. If up to tens of existing items have been removed and the storage left is still insufficient (either due to the total capacity limit specified by [lua_shared_dict](#lua_shared_dict) or memory segmentation), then the `err` return value will be `no memory` and `success` will be `false`.
 
+If the sizes of items in the dictionary are not approximate, it is eaisier to encounter `no memory` error because of memory fragmentation. It is recommended to use different dictionaries for different sizes of items. When you encounter `no memory` error, you can also remove more existing items by retrying more times to make the operation successful.
```
agentzh, line 6551: What do you mean by "sizes are not approximate"? You mean they are not the same here right? Actually, if the sizes are all multiples or even powers of a certain value (like 2), it will also help too though there will still be a chance for fragmentation.

Also, this discussion also applies to other storing operations like `add()` and etc. We should add a separate section instead of repeating this in every storing operation's doc?

Also, better to use shorter lines here to help diffs in the future.

wanghuizzz: Other storing operations already have pointed out that they are similar to the `set` method, so I think adding here is ok.

README.markdown
```diff
@@ -6548,6 +6548,10 @@ The optional `flags` argument specifies a user flags value associated with the e
 
 When it fails to allocate memory for the current key-value item, then `set` will try removing existing items in the storage according to the Least-Recently Used (LRU) algorithm. Note that, LRU takes priority over expiration time here. If up to tens of existing items have been removed and the storage left is still insufficient (either due to the total capacity limit specified by [lua_shared_dict](#lua_shared_dict) or memory segmentation), then the `err` return value will be `no memory` and `success` will be `false`.
 
+If the sizes of items in the dictionary are not multiples or even powers of a certain value (like 2), it is easier to encounter `no memory` error because of memory fragmentation. It is recommended to use different dictionaries for different sizes of items.
+
+When you encounter `no memory` error, you can also remove more existing items by retrying more times to make the operation successful.
```
agentzh, line 6553: I think "evict" is a better word than "remove" here. Also "more existing items" should be "more least-recently-used items".

"by retrying" => "by retrying this method call"

"to make the operation successful" => "to make room for the current item". It's not necessarily possible to succeed if we keep retrying (like when the item is larger than the whole dictionary).

agentzh:
Merged. Thanks.

PR #1694, opened 2020-04-18
validate and expose nextUpdate field in OCSP response

ElvinEfendi:

Required by https://github.com/openresty/lua-resty-core/pull/296

This is an alternative solution to https://github.com/openresty/lua-nginx-module/pull/1041. I like the approach in this PR better because it adds minimal amount of code and it makes sense for validation API to return this data.

The code has been [fork lifted from Nginx source](https://github.com/nginx/nginx/blob/3ba88365b5acef17f01671cd969c909dee5e2cde/src/event/ngx_event_openssl_stapling.c#L700) and adjusted here.

In addition to exposing `nextUpdate` field, this way we will also be validating `nextUpdate` field.

### DIFF
--- a/src/ngx_http_lua_ssl_ocsp.c
+++ b/src/ngx_http_lua_ssl_ocsp.c
@@ -20,7 +20,7 @@
 static int ngx_http_lua_ssl_empty_status_callback(ngx_ssl_conn_t *ssl_conn,
     void *data);
 
-static time_t ngx_http_lua_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time);
+static long ngx_http_lua_ssl_stapling_time(ASN1_GENERALIZEDTIME *asn1time);
 #endif
 
 int
@@ -263,7 +263,7 @@ failed:
 int
 ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
     size_t resp_len, const char *chain_data, size_t chain_len,
-    u_char *errbuf, size_t *errbuf_size, time_t *valid)
+    u_char *errbuf, size_t *errbuf_size, long *valid)
 {
 #ifndef NGX_HTTP_LUA_USE_OCSP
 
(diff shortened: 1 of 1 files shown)


ElvinEfendi:
For tests see https://github.com/openresty/lua-resty-core/pull/296#issuecomment-615883139

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -383,6 +384,16 @@ ngx_http_lua_ffi_ssl_validate_ocsp_response(const u_char *resp,
         goto error;
     }
 
+    if (nextupdate) {
```
agentzh, line 387: Shall we explicitly test if `nextupdate` is equal to zero instead of testing it as a boolean value?

src/ngx_http_lua_ssl_ocsp.c
```diff
@@ -437,6 +448,39 @@ ngx_http_lua_ssl_empty_status_callback(ngx_ssl_conn_t *ssl_conn, void *data)
...
+     */
+
+    bio = BIO_new(BIO_s_mem());
+    if (bio == NULL) {
+        return NGX_ERROR;
+    }
+
+    /* fake weekday prepended to match C asctime() format */
+
+    BIO_write(bio, "Tue ", sizeof("Tue ") - 1);
```
agentzh, line 474: Shall we check the return values of `BIO_write()` for any errors here?

src/ngx_http_lua_ssl_ocsp.c
ElvinEfendi: It's a double pointer to `ASN1_GENERALIZEDTIME` type, I don't know if we can easily compare it to zero. Also NGINX follows this approach: https://github.com/nginx/nginx/blob/7547581bbcb7b737710f9141260d822a08685b83/src/event/ngx_event_openssl_stapling.c#L2392

src/ngx_http_lua_ssl_ocsp.c
ElvinEfendi: Since the input is hardcoded, there's probably not much value in testing that. But if you feel strongly about it I can check the return value too.

zhuizhuhaomeng:
@ElvinEfendi  I would like to merge this PR.
But I found the t/ocsp.t  TEST 19 can not pass.
got the following error other than that in the test case.
Would you please help add some doc about how to generate the revoke ocsp. 
```
2022/01/08 13:54:59 [error] 589#0: *8 [lua] ssl_certificate_by_lua:22: failed to validate OCSP response: certificate status not found in the OCSP response. next_update: nil, context: ssl_certificate_by_lua*, client: unix:, server: unix:/home/ljl/code/openresty/lua-resty-core/t/servroot/html/nginx.sock
```

zhuizhuhaomeng (approved):

PR #1715, opened 2020-06-06
test: add test cases about http response header location 

zhuizhuhaomeng:
testcast: add tow cases about http response header location in ngx_http_static_module
    
    see https://github.com/openresty/openresty/pull/611

### DIFF
--- a/t/162-static-module-location.t
+++ b/t/162-static-module-location.t
@@ -93,4 +93,3 @@ GET /t/%E4%B8%AD%E6%96%87?q=name
 Location: http://localhost:1984/t/%E4%B8%AD%E6%96%87/?q=name
 --- response_body_like
 .*301 Moved Permanently.*
-
--- a/t/162-static-module-location.t
+++ b/t/162-static-module-location.t
@@ -1,0 +1,96 @@
+use Test::Nginx::Socket::Lua;
+
+repeat_each(2);
+
+plan tests => repeat_each() * (blocks() * 3);
+
+our $HtmlDir = html_dir;
+
+no_long_string();
+run_tests();
+
+__DATA__
+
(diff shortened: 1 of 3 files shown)


t/162-static-module-location.t
```diff
@@ -0,0 +1,130 @@
...
+    server_tokens off;
+--- user_files
+>>> t/中文/foo.txt
+Hello, world
+--- request
+GET /t/%E4%B8%AD%E6%96%87?q=name
+--- error_code: 301
+--- response_headers
+Location: http://localhost:1984/t/%E4%B8%AD%E6%96%87/?q=name
+--- response_body eval
```
agentzh, line 122: Why use a regex pattern here when the pattern is essentially a literal string? Am I overlooking anything here? The same applies things above. I suggest only test for a sub-string like "301 Moved Permanently" instead of the full HTML content. The same applies to other test cases here.

t/162-static-module-location.t
```diff
@@ -0,0 +1,96 @@
...
+>>> t/中文/foo.txt
+Hello, world
+--- request
+GET /t/%E4%B8%AD%E6%96%87?q=name
+--- error_code: 301
+--- response_headers
+Location: http://localhost:1984/t/%E4%B8%AD%E6%96%87/?q=name
+--- response_body_like
+.*301 Moved Permanently.*
+
```
agentzh, line 96: You forgot to use the `reindex` script to re-format your `.t`files? This extra blank line should not exist.

agentzh:
@zhuizhuhaomeng Please use the `tests: ` prefix instead of `test: ` or `testcast: ` (the latter also looks like a typo).

agentzh:
Merged after fixing the commit log message. Thanks.

PR #1716, opened 2020-06-06
bugfix: ngx.req.set_uri_args with string argument can contain invalid…

zhuizhuhaomeng:
bugfix: ngx.req.set_uri_args with string argument can contain invalid chacters.

throw an Lua Exception when argument contain \r or  \n or \0

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4552,7 +4552,8 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```
 
-where in the latter case, this method will escape argument keys and values according to the URI escaping rule.
+In the former case, this method will throw out a Lua exception if `args` contain '\r' or '\n' or '\0'.
+In the latter case, this method will escape argument keys and values according to the URI escaping rule.
 
 Multi-value arguments are also supported:
 
--- a/README.markdown
+++ b/README.markdown
@@ -4552,7 +4552,8 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```
 
-In the former case, this method will throw out a Lua exception if `args` contain '\r' or '\n' or '\0'.
+In the former case, the input Lua string should be escaped. Otherwise, this
+method will escape ASCII code 0x00~0x32 and 0x7F in the Lua string.
 In the latter case, this method will escape argument keys and values according to the URI escaping rule.
 
 Multi-value arguments are also supported:
--- a/src/ngx_http_lua_args.c
+++ b/src/ngx_http_lua_args.c
@@ -17,20 +17,6 @@
 
 static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
-/*
-https://tools.ietf.org/html/rfc3986*
-URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]
-pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
-query         = *( pchar / "/" / "?" )
-fragment      = *( pchar / "/" / "?" )
-pct-encoded   = "%" HEXDIG HEXDIG
-unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
-reserved      = gen-delims / sub-delims
-sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
-              / "*" / "+" / "," / ";" / "="
-
-gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
- */
 
 static ngx_inline ngx_int_t
 ngx_http_lua_is_escaped_query_string(ngx_http_request_t *r, u_char *str,
--- a/src/ngx_http_lua_args.c
+++ b/src/ngx_http_lua_args.c
@@ -19,9 +19,14 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
 
 
-static ngx_inline int
-ngx_http_lua_valid_args(const u_char *src, size_t size)
+static ngx_inline ngx_uint_t
+ngx_http_lua_is_escaped_query_string(ngx_http_request_t *r, u_char *str,
+    size_t len)
 {
+    size_t           i, buf_len;
+    u_char           c;
+    u_char          *buf, *src = str;
+
     static uint32_t   map[] = {
         0x00002401, /* 0000 0000 0000 0000  0010 0100 0000 0001 */
 
@@ -40,14 +45,27 @@ ngx_http_lua_valid_args(const u_char *src, size_t size)
         0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
     };
 
-    while (size) {
-        if (map[*src >> 5] & (1U << (*src & 0x1f))) {
+    for (i = 0; i < len; i++, str++) {
+        c = *str;
+        if (map[c >> 5] & (1 << (c & 0x1f))) {
+            buf_len = ngx_http_lua_escape_log(NULL, src, len);
+            buf = ngx_palloc(r->pool, buf_len);
+            if (buf == NULL) {
+                return NGX_ERROR;
+            }
+
+            ngx_http_lua_escape_log(buf, src, len);
+
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "unsafe byte \"0x%uxd\" in query string \"%*s\"",
+                          (unsigned) c, buf_len, buf);
+
+            ngx_pfree(r->pool, buf);
+
             return 0;
         }
-
-        src++;
-        size--;
     }
+
     return 1;
 }
 
@@ -59,6 +77,7 @@ ngx_http_lua_ngx_req_set_uri_args(lua_State *L)
     const char                  *msg;
     size_t                       len;
     u_char                      *p;
+    u_char                       c;
     int                          type;
 
     if (lua_gettop(L) != 1) {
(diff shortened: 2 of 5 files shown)


src/ngx_http_lua_args.c
```diff
@@ -40,10 +73,14 @@ ngx_http_lua_ngx_req_set_uri_args(lua_State *L)
...
     ngx_http_lua_check_fake_request(L, r);
 
-    switch (lua_type(L, 1)) {
+    type = lua_type(L, 1);
+    switch (type) {
     case LUA_TNUMBER:
     case LUA_TSTRING:
         p = (u_char *) lua_tolstring(L, 1, &len);
+        if (type == LUA_TSTRING && !ngx_http_lua_valid_args(p, len)) {
+            return luaL_argerror(L, 1, "args need to be escaped");
```
agentzh, line 82: We should print out offending byte otherwise it would be very difficult to debug such errors. Also, error messages should be saying what is wrong instead of only what to do. So a message like "query-string contains unescaped byte 0x09" would be much better here.

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,38 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
...
+        if (map[*src >> 5] & (1U << (*src & 0x1f))) {
+            return 0;
+        }
+
+        src++;
+        size--;
+    }
+    return 1;
+}
+
```
agentzh, line 53: Forgot to run `ngx-releng` to check this?

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,38 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
...
+    while (size) {
+        if (map[*src >> 5] & (1U << (*src & 0x1f))) {
+            return 0;
+        }
+
+        src++;
+        size--;
+    }
+    return 1;
```
agentzh, line 51: A blank line should be put after code blocks.

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,38 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
 
 
+static ngx_inline int
+ngx_http_lua_valid_args(const u_char *src, size_t size)
```
agentzh, line 23: Use of `const` looks wrong here since you actually change the value of the `src` pointer below.

src/ngx_http_lua_args.c
```diff
@@ -40,10 +73,14 @@ ngx_http_lua_ngx_req_set_uri_args(lua_State *L)
 
     ngx_http_lua_check_fake_request(L, r);
 
-    switch (lua_type(L, 1)) {
+    type = lua_type(L, 1);
+    switch (type) {
     case LUA_TNUMBER:
     case LUA_TSTRING:
         p = (u_char *) lua_tolstring(L, 1, &len);
+        if (type == LUA_TSTRING && !ngx_http_lua_valid_args(p, len)) {
```
agentzh, line 81: I suggest you no longer combine the `LUA_TNUMBER` and `LUA_TSTRING` cases together otherwise you are testing against the `LUA_TSTRING` twice here...

src/ngx_http_lua_args.c
agentzh: I don't like this C function name. `args` is confusing when being used alone. How about `ngx_http_lua_is_escaped_query_string()`?

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,56 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
 
 
+static ngx_inline ngx_uint_t
```
agentzh, line 22: Please note that in nginx's code base, it's usually not using `ngx_uint_t` for boolean types. Maybe you should just use `ngx_int_t` and then returning `NGX_OK` and `NGX_ERROR` instead.

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,56 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
...
+ngx_http_lua_is_escaped_query_string(ngx_http_request_t *r, u_char *str,
+    size_t len)
+{
+    size_t           i, buf_len;
+    u_char           c;
+    u_char          *buf, *src = str;
```
agentzh, line 28: Please do not declare multiple variables in a single declarator. We usually don't do this in the nginx code base.

src/ngx_http_lua_args.c
```diff
@@ -19,6 +19,58 @@ static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
...
+                return NGX_ERROR;
+            }
+
+            ngx_http_lua_escape_log(buf, src, len);
+
+            ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                          "unsafe byte \"0x%uxd\" in query string \"%*s\"",
+                          (unsigned) c, buf_len, buf);
+
+            ngx_pfree(r->pool, buf);
```
agentzh, line 65: This looks very wasteful. Why checking if the query-string contains invalid bytes requires memory allocations and data copying?

README.markdown
```diff
@@ -4552,7 +4552,8 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```
 
-where in the latter case, this method will escape argument keys and values according to the URI escaping rule.
+In the former case, this method will throw out a Lua exception if `args` contain '\r' or '\n' or '\0'.
```
agentzh, line 4555: These are not enough. Should cover `\t` and other control and non-printable ASCII codes as well.

src/ngx_http_lua_args.c
```diff
@@ -42,8 +80,37 @@ ngx_http_lua_ngx_req_set_uri_args(lua_State *L)
...
+
+            ngx_http_lua_escape_log(buf, p, len);
+            /* lua_pushfstring should be zero terminated */
+            buf[buf_len] = '\0';
+            msg = lua_pushfstring(L, "unescaped byte 0x%x in query string "
+                                  "\"%s\"", b, buf);
+
+            ngx_pfree(r->pool, buf);
+
+            return luaL_argerror(L, 1, msg);
```
agentzh, line 112: Sorry, after thinking more about this, I believe a better approach is to simply escape all these control and non-printable ASCII bytes automatically according to the URI escaping rules. This is also the way with minimal impact to the existing user code.

src/ngx_http_lua_args.c
```diff
@@ -18,6 +18,65 @@
 static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
 
+uintptr_t
+ngx_http_lua_escape_args(u_char *dst, u_char *src, size_t size)
+{
+    ngx_uint_t      n;
+    static u_char   hex[] = "0123456789ABCDEF";
+
+
```
agentzh, line 27: We should only use a single blank line here.

README.markdown
```diff
@@ -4552,7 +4552,9 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```
 
-where in the latter case, this method will escape argument keys and values according to the URI escaping rule.
+In the former case, the input Lua string should be escaped. Otherwise, this
+method will escape ASCII code 0x00~0x32 and 0x7F in the Lua string.
```
agentzh, line 4556: Your implementation is 0 ~ 0x20 but your doc states "0x00~0x32". Typo?

src/ngx_http_lua_args.c
agentzh: It's still a single blank line? Seems like `ngx-releng` misses this issue for some reasons I don't know. We always need 2 blank lines to separate C function definitions. See https://openresty.org/en/c-coding-style-guide.html

src/ngx_http_lua_args.c
```diff
@@ -18,6 +18,65 @@
 static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);
```
agentzh, line 20: Also, two blank lines needed here.

agentzh:
Merged after applying the following patch:

````diff
diff --git a/README.markdown b/README.markdown
index e86d94bf..c822b3d4 100644
--- a/README.markdown
+++ b/README.markdown
@@ -4552,8 +4552,11 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```

-In the former case, the input Lua string should be escaped. Otherwise, this
-method will escape ASCII code 0x00~0x32 and 0x7F in the Lua string.
+In the former case, i.e., when the whole query-strng is provided directly,
+the input Lua string should already be well-formed with the URI encoding.
+For security considerations, his method will autoamticaly escape any control and
+whitespace characters (ASCII code 0x00 ~ 0x32 and 0x7F) in the Lua string.
+
 In the latter case, this method will escape argument keys and values according to the URI escaping rule.

 Multi-value arguments are also supported:
index 7066f58f..6f997016 100644
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -3791,8 +3791,11 @@ or a Lua table holding the query arguments' key-value pairs, as in
     ngx.req.set_uri_args({ a = 3, b = "hello world" })
 </geshi>

-In the former case, the input Lua string should be escaped. Otherwise, this
-method will escape ASCII code 0x00~0x32 and 0x7F in the Lua string.
+In the former case, i.e., when the whole query-strng is provided directly,
+the input Lua string should already be well-formed with the URI encoding.
+For security considerations, his method will autoamticaly escape any control and
+whitespace characters (ASCII code 0x00 ~ 0x32 and 0x7F) in the Lua string.
+
 In the latter case, this method will escape argument keys and values according to the URI escaping rule.

 Multi-value arguments are also supported:
diff --git a/src/ngx_http_lua_args.c b/src/ngx_http_lua_args.c
index 4e7c40ef..7b7d5eae 100644
--- a/src/ngx_http_lua_args.c
+++ b/src/ngx_http_lua_args.c
@@ -18,13 +18,13 @@
 static int ngx_http_lua_ngx_req_set_uri_args(lua_State *L);
 static int ngx_http_lua_ngx_req_get_post_args(lua_State *L);

+
 uintptr_t
 ngx_http_lua_escape_args(u_char *dst, u_char *src, size_t size)
 {
     ngx_uint_t      n;
     static u_char   hex[] = "0123456789ABCDEF";

-
                     /* %00-%20 %7F*/

     static uint32_t   escape[] = {
@@ -78,6 +78,7 @@ ngx_http_lua_escape_args(u_char *dst, u_char *src, size_t size)
     return (uintptr_t) dst;
 }

+
 static int
 ngx_http_lua_ngx_req_set_uri_args(lua_State *L)
 {
````

PR #1717, opened 2020-06-07
bugfix: escape http header

zhuizhuhaomeng:
bugfix:
    1. escape \r\n\0 in http header value, escape CTLs or separators in http header name
    2. allow \t in http uri
    
    goto rfc7230 https://tools.ietf.org/html/rfc7230#section-3.2 for detail

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -4552,9 +4552,9 @@ or a Lua table holding the query arguments' key-value pairs, as in
  ngx.req.set_uri_args({ a = 3, b = "hello world" })
 ```
 
-In the former case, i.e., when the whole query-strng is provided directly,
+In the former case, i.e., when the whole query-string is provided directly,
 the input Lua string should already be well-formed with the URI encoding.
-For security considerations, his method will autoamticaly escape any control and
+For security considerations, this method will automatically escape any control and
 whitespace characters (ASCII code 0x00 ~ 0x32 and 0x7F) in the Lua string.
 
 In the latter case, this method will escape argument keys and values according to the URI escaping rule.
@@ -4882,6 +4882,11 @@ ngx.req.set_header
 **context:** *set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;*
 
 Set the current request's request header named `header_name` to value `header_value`, overriding any existing ones.
+
+The input Lua string `header_name` and `header_value` should already be well-formed with the URI encoding.
+For security considerations, this method will automatically escape " ", """, "(", ")", ",", "/", ":", ";", "?",
+"<", "=", ">", "?", "@", "[", "]", "\", "{", "}", 0x00-0x1F, 0x7F-0xFF in `header_name` and automatically escape
+"0x00-0x08, 0x0A-0x0F, 0x7F in `header_value`.
 
 By default, all the subrequests subsequently initiated by [ngx.location.capture](#ngxlocationcapture) and [ngx.location.capture_multi](#ngxlocationcapture_multi) will inherit the new header.
 
--- a/src/ngx_http_lua_headers_in.c
+++ b/src/ngx_http_lua_headers_in.c
@@ -653,16 +653,18 @@ ngx_http_lua_set_input_header(ngx_http_request_t *r, ngx_str_t key,
 {
     ngx_http_lua_header_val_t         hv;
     ngx_http_lua_set_header_t        *handlers = ngx_http_lua_set_handlers;
-
+    ngx_int_t                         rc;
     ngx_uint_t                        i;
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    if (ngx_http_lua_check_unsafe_string(r, key.data, key.len,
-                                         "header name") != NGX_OK
-        || ngx_http_lua_check_unsafe_string(r, value.data, value.len,
-                                            "header value") != NGX_OK)
-    {
+    rc = ngx_http_lua_copy_escaped_header(r, &key, 1);
+    if (rc != NGX_OK) {
+        return NGX_ERROR;
+    }
+
+    rc = ngx_http_lua_copy_escaped_header(r, &value, 0);
+    if (rc != NGX_OK) {
         return NGX_ERROR;
     }
 
--- a/src/ngx_http_lua_headers_out.c
+++ b/src/ngx_http_lua_headers_out.c
@@ -491,11 +491,11 @@ ngx_http_lua_set_output_header(ngx_http_request_t *r, ngx_http_lua_ctx_t *ctx,
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    if (ngx_http_lua_check_unsafe_string(r, key.data, key.len,
-                                         "header name") != NGX_OK
-        || ngx_http_lua_check_unsafe_string(r, value.data, value.len,
-                                            "header value") != NGX_OK)
-    {
+    if (ngx_http_lua_copy_escaped_header(r, &key, 1) != NGX_OK) {
+        return NGX_ERROR;
+    }
+
+    if (ngx_http_lua_copy_escaped_header(r, &value, 0) != NGX_OK) {
         return NGX_ERROR;
     }
 
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -2008,8 +2008,52 @@ ngx_http_lua_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
                     /* mail_auth is the same as memcached */
 
+                    /* " ", """, "(", ")", ",", "/", ":", ";", "?",
+                     * "<", "=", ">", "?", "@", "[", "]", "\", "{",
+                     * "}", %00-%1F, %7F-%FF
+                     */
+
+    static uint32_t   header_name[] = {
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+
+                    /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
+        0xfc009305, /* 1111 1100 0000 0000  1001 0011 0000 0101 */
+
+                    /* _^]\ [ZYX WVUT SRQP  ONML KJIH GFED CBA@ */
+        0x38000001, /* 0011 1000 0000 0000  0000 0000 0000 0001 */
+
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
+        0xa8000000, /* 1010 1000 0000 0000  0000 0000 0000 0000 */
+
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+    };
+
+                    /* "%00-%08, %0A-%0F, %7F */
+
+    static uint32_t   header_value[] = {
+        0xfffffdff, /* 1111 1111 1111 1111  1111 1101 1111 1111 */
+
+                    /* ?>=< ;:98 7654 3210  /.-, +*)( '&%$ #"!  */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+
+                    /* _^]\ [ZYX WVUT SRQP  ONML KJIH GFED CBA@ */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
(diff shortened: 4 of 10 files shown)


src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,36 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
     return 0;
 }
 
+ngx_int_t ngx_http_lua_copy_escape_header(ngx_http_request_t *r,
```
agentzh, line 4367: Bad coding style. Forgot to run the `ngx-releng` tool?

src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,36 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
...
+    ngx_str_t *dst, ngx_int_t header_name)
```
agentzh, line 4368: What does this `header_name` argument mean? If it's a boolean value, then we should use the type `unsigned` and parameter name `is_name` instead.

src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,36 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
...
+        dst->data = ngx_palloc(r->pool, len + 2 * escape + 1);
+        if (dst->data == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_http_lua_escape_uri(dst->data, data, len, type);
+        dst->len = len + 2 * escape;
+        dst->data[dst->len] = '\0';
+    }
+    return NGX_OK;
```
agentzh, line 4396: Please always use a blank line after `if` statement code blocks. We mentioned this before.

src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,36 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
...
+         * header values must be null-terminated
+         * */
+        dst->data = ngx_palloc(r->pool, len + 2 * escape + 1);
+        if (dst->data == NULL) {
+            return NGX_ERROR;
+        }
+
+        ngx_http_lua_escape_uri(dst->data, data, len, type);
+        dst->len = len + 2 * escape;
+        dst->data[dst->len] = '\0';
```
agentzh, line 4394: Do we really need the `\0` byte here? If yes, why? Better leave a code comment here to explain.

t/113-req-header-cookie.t
```diff
@@ -263,9 +263,10 @@ Cookie: boo=123; foo=bar
...
-unsafe byte "0xa" in header value "boo=123\x0Afoo"
-failed to set header
+--- response_body
+Cookie foo: bar%0Dbar
+Cookie baz: 
+Cookie boo: 123%0Afoo
+Cookie: boo=123%0Afoo; foo=bar%0Dbar
 --- no_error_log
-[crit]
+[error]
```
agentzh, line 272: Please always add one and only one newline at the end of the test file (and other source files).

src/ngx_http_lua_util.c
zhuizhuhaomeng: ngx-releng won't complain when return type and function name at the same line

src/ngx_http_lua_util.c
zhuizhuhaomeng: there is a comment on line 4384

src/ngx_http_lua_headers_in.c
```diff
@@ -653,16 +653,19 @@ ngx_http_lua_set_input_header(ngx_http_request_t *r, ngx_str_t key,
...
     ngx_uint_t                        i;
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    if (ngx_http_lua_check_unsafe_string(r, key.data, key.len,
-                                         "header name") != NGX_OK
-        || ngx_http_lua_check_unsafe_string(r, value.data, value.len,
-                                            "header value") != NGX_OK)
-    {
+    rc = ngx_http_lua_copy_escape_header(r, &key, 1);
```
agentzh, line 662: I think "copy escaped header" sounds more correct than "copy escape header". In English we cannot just put two verbs together...

src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,38 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
     return 0;
 }
```
agentzh, line 4366: Style: 2 blank lines are needed here. Forgot to run `ngx-releng` again?

src/ngx_http_lua_util.c
```diff
@@ -4327,5 +4364,38 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
...
+
+    data = dst->data;
+    len = dst->len;
+
+    escape = ngx_http_lua_escape_uri(NULL, data, len, type);
+    if (escape > 0) {
+        /*
+         * we allocate space for the trailling '\0' char here because nginx
+         * header values must be null-terminated
+         * */
```
agentzh, line 4387: Style: `* */` should be `*/`.

src/ngx_http_lua_util.c
agentzh: Gotcha.

src/ngx_http_lua_util.h
```diff
@@ -27,6 +27,9 @@
 #   define NGX_HTTP_SWITCHING_PROTOCOLS 101
 #endif
 
+#define NGX_HTTP_LUA_ESCAPE_HEADER_NAME  7
+
+#define NGX_HTTP_LUA_ESCAPE_HEADER_VALUE 8
```
agentzh, line 32: Please use 2 spaces before `8`, just like the previous `#define` on line 30.

src/ngx_http_lua_headers_out.c
```diff
@@ -488,14 +488,17 @@ ngx_http_lua_set_output_header(ngx_http_request_t *r, ngx_http_lua_ctx_t *ctx,
...
 
     dd("set header value: %.*s", (int) value.len, value.data);
 
-    if (ngx_http_lua_check_unsafe_string(r, key.data, key.len,
-                                         "header name") != NGX_OK
-        || ngx_http_lua_check_unsafe_string(r, value.data, value.len,
-                                            "header value") != NGX_OK)
-    {
+    rc = ngx_http_lua_copy_escape_header(r, &key, 1);
+    if (rc != NGX_OK) {
```
agentzh, line 496: We don't need to introduce the `rc` variable here since it's always used only once. So better be:

```C
if (ngx_http_lua_copy_escape_header(r, &key, 1) != NGX_OK) {
    return NGX_ERROR;
}
```

The same applies to the `if` statement below.

src/ngx_http_lua_util.h
```diff
@@ -169,6 +172,9 @@ void ngx_http_lua_unescape_uri(u_char **dst, u_char **src, size_t size,
 uintptr_t ngx_http_lua_escape_uri(u_char *dst, u_char *src,
     size_t size, ngx_uint_t type);
 
+ngx_int_t ngx_http_lua_copy_escaped_header(ngx_http_request_t *r,
+    ngx_str_t *dst, ngx_uint_t header_name);
```
agentzh, line 176: We usually use `int` or `unsigned` for boolean types (the only exception is in the nginx directive configuration where we use `ngx_flag_t` for boolean config values). So `int` here is better for `is_name` (BTW, it should be `is_name` than `header_name`).

`ngx_uint_t` is like `unsigned long` on x64, BTW, which is 8 bytes long.

src/ngx_http_lua_util.c
```diff
@@ -4328,4 +4365,38 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
 }
 
 
+ngx_int_t
+ngx_http_lua_copy_escaped_header(ngx_http_request_t *r,
+    ngx_str_t *dst, ngx_uint_t is_name)
```
agentzh, line 4370: Like below.

src/ngx_http_lua_util.c
```diff
@@ -2008,8 +2008,45 @@ ngx_http_lua_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
 
                     /* mail_auth is the same as memcached */
```
agentzh, line 2010: We should summary here in a comment about exactly what characters are to be escaped, like `/* " ", "%", %00-%1F */` for the memcached case above.

src/ngx_http_lua_util.c
```diff
@@ -2008,8 +2008,45 @@ ngx_http_lua_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
...
 
+    static uint32_t header_name[] = {
```
agentzh, line 2011: Style: needs 2 spaces before the identifier `header_name`. See above for examples (like `memcached[]`).

src/ngx_http_lua_util.c
```diff
@@ -2008,8 +2008,45 @@ ngx_http_lua_escape_uri(u_char *dst, u_char *src, size_t size, ngx_uint_t type)
...
+                    /*  ~}| {zyx wvut srqp  onml kjih gfed cba` */
+        0xa8000000, /* 1010 1000 0000 0000  0000 0000 0000 0000 */
+
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+        0x00000000, /* 0000 0000 0000 0000  0000 0000 0000 0000 */
+    };
+
+    static uint32_t header_value[] = {
```
agentzh, line 2029: Ditto.

src/ngx_http_lua_util.h
```diff
@@ -500,7 +506,7 @@ ngx_http_lua_check_unsafe_string(ngx_http_request_t *r, u_char *str, size_t len,
                      /* %00-%1F, %7F */
```
agentzh, line 506: You forgot the update this comment to reflect the change below?

I believe a general "unsafe string" makes no sense (as I mentioned in a review comment yesterday). Headers, URIs, URI arguments may have different "unsafe" byte ranges anyway. We'd better be more explicit about our intention.

BTW, do we really need this function now?

t/016-resp-header.t
```diff
@@ -1976,14 +1976,12 @@ Content-Type: application/json
     }
 --- request
 GET /t
---- error_code: 500
 --- response_headers
-header:
+header: value%0Dfoo:bar%0Abar:foo
```
agentzh, line 1980: Do we need to update the documentation to reflect the behavioral changes in this PR?

src/ngx_http_lua_util.h
```diff
@@ -500,7 +506,7 @@ ngx_http_lua_check_unsafe_string(ngx_http_request_t *r, u_char *str, size_t len,
...
 
     static uint32_t  unsafe[] = {
-        0xffffffff, /* 1111 1111 1111 1111  1111 1111 1111 1111 */
+        0xfffffdff, /* 1111 1111 1111 1111  1111 1101 1111 1111 */
```
agentzh, line 509: Why this change?

src/ngx_http_lua_headers_in.c
```diff
@@ -653,16 +653,19 @@ ngx_http_lua_set_input_header(ngx_http_request_t *r, ngx_str_t key,
 {
     ngx_http_lua_header_val_t         hv;
     ngx_http_lua_set_header_t        *handlers = ngx_http_lua_set_handlers;
+    ngx_int_t                         rc;
```
agentzh, line 657: We can remove this blank line, BTW.

src/ngx_http_lua_util.c
```diff
@@ -4328,4 +4365,38 @@ ngx_http_lua_escape_log(u_char *dst, u_char *src, size_t size)
...
+        ? NGX_HTTP_LUA_ESCAPE_HEADER_NAME : NGX_HTTP_LUA_ESCAPE_HEADER_VALUE;
+
+    data = dst->data;
+    len = dst->len;
+
+    escape = ngx_http_lua_escape_uri(NULL, data, len, type);
+    if (escape > 0) {
+        /*
+         * we allocate space for the trailling '\0' char here because nginx
+         * header values must be null-terminated
```
agentzh, line 4387: Will you add a reference to this statement? Like "see src/blah/blah.c" for details.

README.markdown
```diff
@@ -4883,6 +4883,11 @@ ngx.req.set_header
 
 Set the current request's request header named `header_name` to value `header_value`, overriding any existing ones.
 
+The input Lua string header_name and header_value should already be well-formed with the URI encoding.
```
agentzh, line 4886: Missing backticks around `header_name` and `header_value`. See the previous paragraph for examples.

README.markdown
```diff
@@ -4883,6 +4883,11 @@ ngx.req.set_header
...
+For security considerations, this method will automatically escape " ", """, "(", ")", ",", "/", ":", ";", "?",
+"<", "=", ">", "?", "@", "[", "]", "\", "{", "}", %00-%1F, %7F-%FF in header_name and automatically escape
+"%00-%08, %0A-%0F, %7F in header value.
```
agentzh, line 4889: Better use `0xFF` syntax instead of the URI-encoded form.

PR #1718, opened 2020-06-08
test: use iptable -I option to insert rule to the head

zhuizhuhaomeng:
test: use iptable -I option to insert rule to the head of the filter chain in case there are other system default rules  in front of the our own rule.


agentzh:
Merged. Thanks.

PR #1733, opened 2020-06-19
doc: fix the documents regarding the limits on subrequests

eaufavor:
Per the code and the conversation here http://mailman.nginx.org/pipermail/nginx-devel/2015-September/007264.html
The current doc is not accurate.


agentzh:
Merged. Thanks.

PR #1735, opened 2020-06-25
feature(socket.tcp): enhance the logic of parameter verification in

rainingmaster:
In [ngx.socket.tcp:connect](https://github.com/openresty/lua-nginx-module#ngxsocketconnect), if host is a TCP socket, the third parameter could be `nil`, it is to some third part module to write their code like this:
```
function _M.connect(host, port, opt)
    sock = ngx.socket.tcp()
    sock:connect(host, port, opt)
end
```
But if the host is a unix-domain socket, the second parameter could not be a `nil`, we have to write the code like this:
```
function _M.connect(path, opt)
    sock = ngx.socket.tcp()
    if opt ~= nil then
        sock:connect("unix:" .. path, opt)
    else
       sock:connect("unix:" .. path)
   end
end
```
Just like here: https://github.com/openresty/lua-resty-redis/pull/200/files#diff-adb515948c05651051c83fd6fb3ce7ceR143-R149. I think it is a comment use way, so update the second parameter and allow it be nil.

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -970,8 +970,16 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
         n--;
     }
 
-    if (n == 3) {
-        port = luaL_checkinteger(L, 3);
+    /* most popular suit: host:port */
+    if (n == 3 && lua_isnumber(L, 3)) {
+
+        /* Hit the following parameter combination:
+         * sock:connect("127.0.0.1", port)
+         * sock:connect("127.0.0.1", port, opts)
+         * sock:connect("unix:/path", port)
+         * sock:connect("unix:/path", port, opts) */
+
+        port = (int) lua_tointeger(L, 3);
 
         if (port < 0 || port > 65535) {
             lua_pushnil(L);
@@ -987,8 +995,27 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
 
         dd("socket key: %s", lua_tostring(L, -1));
 
-    } else { /* n == 2 */
+    } else if (len >= 5 && ngx_strncasecmp(p, (u_char *) "unix:", 5) == 0) {
+
+        /* Hit the following parameter combination:
+         * sock:connect("unix:/path")
+         * sock:connect("unix:/path", nil)
+         * sock:connect("unix:/path", opts)
+         * sock:connect("unix:/path", nil, opts) */
+
         port = 0;
+
+    } else {
+
+        /* Ban the following parameter combination:
+         * sock:connect("127.0.0.1")
+         * sock:connect("127.0.0.1", nil)
+         * sock:connect("127.0.0.1", opts) 
+         * sock:connect("127.0.0.1", nil, opts) */
+
+        lua_pushnil(L);
+        lua_pushfstring(L, "missing the port number");
+        return 2;
     }
 
     if (!custom_pool) {
--- a/t/058-tcp-socket.t
+++ b/t/058-tcp-socket.t
@@ -4,7 +4,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * 228;
+plan tests => repeat_each() * 231;
 
 our $HtmlDir = html_dir;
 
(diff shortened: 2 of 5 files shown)


agentzh:
@rainingmaster Please use `feature: ` prefix in the commit log titles. Thank you!

src/ngx_http_lua_socket_tcp.c
```diff
@@ -974,7 +974,9 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
         n--;
     }
 
-    if (n == 3) {
+    if (n == 3 &&
+        !(len >= 5 && ngx_strncasecmp(p, (u_char *) "unix:", 5) == 0))
```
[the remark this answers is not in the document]
rainingmaster: Yes, allow the port optional make the code more clear.

rainingmaster:
Hi @agentzh, I have update the code, make the port optional and accept nil, could you help me review again?

README.markdown
```diff
@@ -7246,6 +7246,8 @@ Both IP addresses and domain names can be specified as the `host` argument. In c
 
 If the nameserver returns multiple IP addresses for the host name, this method will pick up one randomly.
 
+`port` is optional and accept nil(for compatibility with host is a unix domain), and it default value is zero.
```
spacewander, line 7249: There should be a space before `(for compatibility`?

t/058-tcp-socket.t
```diff
@@ -4327,3 +4327,36 @@ failed to receive a line: closed []
...
+
+=== TEST 72: port is optional and accept nil, default is 0
+--- config
+    server_tokens off;
+    location = /t {
+        set $port $TEST_NGINX_SERVER_PORT;
+        content_by_lua_block {
+            local sock = ngx.socket.tcp()
+            sock:settimeout(500)
+            local ok, err = sock:connect("127.0.0.1")
```
agentzh, line 4341: I don't like this behavior. This may lead to hard-to-debug mistakes. When the host name part is an INET address, the port cannot be optional.

rainingmaster: I see, I agree with this point. So we should allow like this:
```
sock:connect("unix:/path") -- work
sock:connect("unix:/path", nil) -- work
sock:connect("unix:/path", opt) -- work

sock:connect("127.0.0.1", 80) -- work
sock:connect("127.0.0.1", 80, nil) -- work
sock:connect("127.0.0.1", 80, opt) -- work
```

But like this will be raise an error:
```
sock:connect("127.0.0.1", nil) -- not work
```

In this version, the behaviour should be like above: https://github.com/openresty/lua-nginx-module/pull/1735#discussion_r445900536, but we have to judge the domain is an unix domain or a ip.

agentzh: @rainingmaster My original comment you quoted was for the unix domain socket case, not for the INET case.

rainingmaster: @agentzh hihi, my quoted should be mean *not* unix domain socket?  so it should mean INET case.

rainingmaster:
@agentzh @doujiang24 @spacewander Hi, I have raise a new version for this [concern](https://github.com/openresty/lua-nginx-module/pull/1735#discussion_r457593035), now it will require that the second parameter must be passed in if the `host` is not `unix domain`.

Another point to note is that the `unix domain` has always supported usage like this:

```
connect("unix/path", nil, opts)
```

IMO, I think this usage should be retained, because users can easily use the format of `connect(host, port, opts)` to pass parameters when implementing their own lib, without the need for more `if`.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -970,7 +970,15 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
         n--;
     }
 
-    if (n == 3) {
+    /* INET address */
+    if (len < 5 || ngx_strncasecmp(p, (u_char *) "unix:", 5) != 0) {
+
+        if (n == 2) {
+            lua_pushnil(L);
+            lua_pushfstring(L, "can not get port number");
```
doujiang24, line 978:
```suggestion
            lua_pushfstring(L, "missing the port number");
```

src/ngx_http_lua_socket_tcp.c
```diff
@@ -987,7 +995,26 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
 
         dd("socket key: %s", lua_tostring(L, -1));
 
-    } else { /* n == 2 */
+    } else if (len < 5 || ngx_strncasecmp(p, (u_char *) "unix:", 5) != 0) {
```
doujiang24, line 998: move the hit of `unix socket domain` to this `else if` block may be better?

rainingmaster: done

PR #1739, opened 2020-06-29
test: add test cases for signal processing

zhuizhuhaomeng:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ install:
   - if [ ! -f download-cache/pcre-$PCRE_VER.tar.gz ]; then wget -P download-cache https://ftp.pcre.org/pub/pcre/pcre-$PCRE_VER.tar.gz; fi
   - if [ ! -f download-cache/openssl-$OPENSSL_VER.tar.gz ]; then wget -P download-cache https://www.openssl.org/source/openssl-$OPENSSL_VER.tar.gz || wget -P download-cache https://www.openssl.org/source/old/${OPENSSL_VER//[a-z]/}/openssl-$OPENSSL_VER.tar.gz; fi
   - git clone https://github.com/openresty/test-nginx.git
-  - git clone https://github.com/openresty/openresty.git ../openresty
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -88,6 +88,7 @@ install:
   - git clone https://github.com/openresty/lua-resty-mysql.git ../lua-resty-mysql
   - git clone https://github.com/openresty/stream-lua-nginx-module.git ../stream-lua-nginx-module
   - git clone -b v2.1-agentzh https://github.com/openresty/luajit2.git luajit2
+  - git clone https://github.com/openresty/lua-resty-signal.git ../lua-resty-signal
```
agentzh, line 91: I'm not feeling comfortable adding this new dependency for the test suite. Can we avoid it? Maybe you can just use `os.execute("kill -HUP " .. pid)` or something like that?

.travis.yml
```diff
@@ -118,6 +119,8 @@ script:
   - ./config no-threads shared enable-ssl3 enable-ssl3-method -g --prefix=$OPENSSL_PREFIX -DPURIFY > build.log 2>&1 || (cat build.log && exit 1)
   - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
   - sudo make PATH=$PATH install_sw > build.log 2>&1 || (cat build.log && exit 1)
+  - cd ../lua-resty-signal/
+  - make
```
agentzh, line 123: This will require changes in our [EC2 test cluster](https://openresty.org/en/ec2-test-cluster.html) as well. Better don't do it.

agentzh:
Also, the travis ci is failing BTW.

agentzh:
It seems to be duplicate with #1738. Why?

.travis.yml
zhuizhuhaomeng: OK

.travis.yml
zhuizhuhaomeng: OK

agentzh:
Merged after fixing some minor grammar issues in the test titles and git log titles.

PR #1748, opened 2020-07-14
fix: add lua-resty-string to lua_package_path

rainingmaster:


t/000--init.t
```diff
@@ -10,7 +10,7 @@ $ENV{TEST_NGINX_MEMCACHED_PORT} ||= 11211;
 $ENV{TEST_NGINX_MYSQL_PORT} ||= 3306;
 
 our $http_config = <<'_EOC_';
-    lua_package_path "../lua-resty-mysql/lib/?.lua;;";
+    lua_package_path "../lua-resty-mysql/lib/?.lua;../lua-resty-string/lib/?.lua;;";
```
spacewander, line 13: Better to mention that `lua-resty-string` is required for `lua-resty-mysql`, as we don't use it directly.

PR #1756, opened 2020-07-25
bugfix: type size_t overflow under 32-bit operating system

syzh:
variable size overflow when reading data larger than 4GB under 32-bit operating system

see: https://forum.openresty.us/d/6408-324g


src/ngx_http_lua_socket_tcp.c
```diff
@@ -2396,7 +2396,7 @@ ngx_http_lua_socket_tcp_read(ngx_http_request_t *r,
     ngx_connection_t            *c;
     ngx_buf_t                   *b;
     ngx_event_t                 *rev;
-    size_t                       size;
+    off_t                       size;
```
rainingmaster, line 2399: we need one more space and align here.

syzh: done

PR #1757, opened 2020-07-27
remove semicolon in lua codes

sjnam:
removes semicolon in lua codes in a document file.


doujiang24:
@sjnam Thanks, merged after tweak the commit log.

thibaultcha:
This PR should have updated the wiki document instead of the README.

Issue #1731, opened 2020-06-18
balancer_by_lua context missed in ngx.var.VARIABLE's doc


RocFang:
ngx.var.VARIABLE is supported in balancer_by_lua context, but it seems the balancer_by_lua context is missed in ngx.var.VARIABLE's doc:

https://github.com/openresty/lua-nginx-module#ngxvarvariable

I don't know if there is something special to pay attention? thanks.

doujiang24:
Thanks for your report, fixed in https://github.com/openresty/lua-nginx-module/pull/1761.

PR #1761, opened 2020-08-02
doc: ngx.var.VARIABLE is available in the balancer_by_lua* context

qyihua:
ngx.var.VARIABLE is supported in balancer_by_lua context, add balancer_by_lua* to ngx.var.VARIABLE's doc.
Add balancer_by_lua test case to t/045-ngx-var.t.
#1731

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3462,7 +3462,7 @@ ngx.var.VARIABLE
 
 **syntax:** *ngx.var.VAR_NAME*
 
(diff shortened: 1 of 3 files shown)


rainingmaster:
@qyihua hihi, it looks like that test is broken, could you help to change the count of test case?

qyihua:
> @qyihua hihi, it looks like that test is broken, could you help to change the count of test case?

hi, i change the count of test case now

PR #1774, opened 2020-08-26
add connection param to enable privileged agent api

levy001:

### DIFF
--- a/src/ngx_http_lua_worker.c
+++ b/src/ngx_http_lua_worker.c
@@ -98,7 +98,7 @@ ngx_http_lua_ffi_get_process_type(void)
 
 
 int
-ngx_http_lua_ffi_enable_privileged_agent(char **err)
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_worker.c
```diff
@@ -107,6 +107,7 @@ ngx_http_lua_ffi_enable_privileged_agent(char **err)
                                            ngx_core_module);
 
     ccf->privileged_agent = 1;
+    ccf->connection_n = connections;
```
doujiang24, line 110: we should not overwrite the existing `connection_n`.

zhuizhuhaomeng: why this is not privileged_agent_connections

doujiang24:
already merged in https://github.com/openresty/lua-nginx-module/pull/1828
closing it.

Issue #1779, opened 2020-09-03
build error with nginx 1.7.4 or older version


timebug:
Hello!

When I upgrade lua-nginx-module form v0.10.13 to v0.10.14 (or newer version) with Nginx 1.7.4, it build failed:

```
lua-nginx-module-0.10.14/src/ngx_http_lua_socket_tcp.c: In function 'ngx_http_lua_tcp_queue_conn_op_cleanup':
lua-nginx-module-0.10.14/src/ngx_http_lua_socket_tcp.c:6053:27: error: 'ngx_event_t' has no member named 'posted'
     if (conn_op_ctx->event.posted) {
                           ^
```

I found https://github.com/openresty/lua-nginx-module/commit/761ddca72deb4595ff93df7840dfa39854126123 this pull request added this part of the code, but Nginx include `event.posted` util 1.7.5 version, see https://github.com/nginx/nginx/commit/37d24e7e3be648a6f73bad18dcf53798a7a1aeb0. Considering that the 1.7.x series is still supported, do something compatible here?

PR #1780, opened 2020-09-03
bugfix: build error with nginx 1.7.4 or older version

timebug:
Please see #1779 for more details.

### DIFF
--- a/src/ngx_http_lua_sleep.c
+++ b/src/ngx_http_lua_sleep.c
@@ -159,11 +159,20 @@ ngx_http_lua_sleep_cleanup(void *data)
     }
 
 #ifdef HAVE_POSTED_DELAYED_EVENTS_PATCH
+#if (nginx_version >= 1007005)
     if (coctx->sleep.posted) {
+#else
+    if (coctx->sleep.prev) {
+#endif
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
                        "lua clean up the posted event for pending ngx.sleep");
 
-        ngx_delete_posted_event(&coctx->sleep);
+        /*
+        * We need the extra parentheses around the argument
+        * of ngx_delete_posted_event() just to work around macro issues in
+        * nginx cores older than 1.7.5 (exclusive).
+        */
+        ngx_delete_posted_event((&coctx->sleep));
     }
 #endif
 }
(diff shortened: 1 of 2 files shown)


timebug:
ping @spacewander

spacewander:
It would be great if you also change this: https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_sleep.c#L162

timebug:
@spacewander done.

spacewander (approved):

src/ngx_http_lua_sleep.c
```diff
@@ -159,11 +159,18 @@ ngx_http_lua_sleep_cleanup(void *data)
     }
 
 #ifdef HAVE_POSTED_DELAYED_EVENTS_PATCH
+    ngx_event_t     *ev;
```
agentzh, line 162: We should never declare variables in the middle of a code block. This will lead to compilation errors in some C compilers.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -6024,8 +6026,13 @@ ngx_http_lua_tcp_queue_conn_op_cleanup(void *data)
                    "lua tcp socket abort queueing, conn_op_ctx: %p, u: %p",
                    conn_op_ctx, u);
 
+#if (nginx_version >= 1007005)
     if (conn_op_ctx->event.posted) {
-        ngx_delete_posted_event(&conn_op_ctx->event);
+#else
+    if (conn_op_ctx->event.prev) {
+#endif
+        ev = &conn_op_ctx->event;
```
agentzh, line 6034: Why introduce the extra `ev` variable? It's only used once.

src/ngx_http_lua_sleep.c
```diff
@@ -159,11 +159,18 @@ ngx_http_lua_sleep_cleanup(void *data)
...
     if (coctx->sleep.posted) {
+#else
+    if (coctx->sleep.prev) {
+#endif
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ngx_cycle->log, 0,
                        "lua clean up the posted event for pending ngx.sleep");
 
-        ngx_delete_posted_event(&coctx->sleep);
+        ev = &coctx->sleep;
+        ngx_delete_posted_event(ev);
```
agentzh, line 173: Like below, the `ev` variable is used only once and is useless.

agentzh (changes requested):

src/ngx_http_lua_socket_tcp.c
timebug: @agentzh Thanks for the feedback, it has been fixed.

src/ngx_http_lua_sleep.c
timebug: @agentzh Has been removed.

PR #1783, opened 2020-09-16
Fix LuaJIT library auto-discovery

### LINKED ISSUE #1661: Configuration isn't checking whether the LuaJIT libraries exist in the given paths on Linux
I'm trying to configure and build ngx_lua with `./configure --with-compat --add-dynamic-module=../lua-nginx-module-0.10.16rc4/`, but it is unable to automatically find LuaJIT, and fails the LuaJIT 2.x check with `error: unsupported LuaJIT version; ngx_http_lua_module requires LuaJIT 2.x.`.

It seems like it's not running any of the "auto-discovery" checks, which I believe is because $ngx_found is still "yes" from the previous check. This can be fixed by removing the if statement around the first check in that else block, "LuaJIT library in /usr/local/", or explicitly setting $ngx_found to no.

Even with that change, it passes the first check, even though LuaJIT is installed in `/usr/`, not `/usr/local/`. I don't think it ever checks if the LuaJIT libraries exist in the paths. This is true when `LUAJIT_LIB` and `LUAJIT_INC` were provided explicitly as well, with paths that do not exist:

```
configuring additional dynamic modules
adding module in ../lua-nginx-module-0.10.16rc4/
checking for LuaJIT library in /dkfjd and /asdfdsf (specified by the LUAJIT_LIB and LUAJIT_INC env, with -ldl) ... found
checking for LuaJIT 2.x ... not found
    ./configure: error: unsupported LuaJIT version; ngx_http_lua_module requires LuaJIT 2.x.
```
or compiling as a standard (not dynamic) module with `LUAJIT_INC=/asdfdsf LUAJIT_LIB=/dkfjd ./configure --with-ld-opt="-Wl,-rpath,/usr/lib" --add-module=../lua-nginx-module-0.10.16rc4/`:
```
configuring additional modules
adding module in ../lua-nginx-module-0.10.16rc4/
checking for LuaJIT library in /dkfjd and /asdfdsf (specified by the LUAJIT_LIB and LUAJIT_INC env, with -ldl) ... found
checking for LuaJIT 2.x ... not found
    ./configure: error: unsupported LuaJIT version; ngx_http_lua_module requires LuaJIT 2.x.
```

I fixed this by changing line 7 from `ngx_feature_incs=` to `ngx_feature_incs="#include <lauxlib.h>"`, which I believe was removed in a96e99ab1a770d261fa9eecbd9c7d112f39aa34b and added it back like it is now in 315330ac099a35f5f17a0d453b7c001fe1181148.

Looking at the ngx_feature_libs in the explicit section, it also appears that the ones in the auto discovery section should have `-L/usr/local/include/luajit-2.0` / `-L/usr/include/luajit-2.0` instead of `-L/usr/local/lib` / `-L/usr/lib`

It also doesn't automatically check for LuaJIT 2.1 at all.

**Build Environment:**
ngx_lua v0.10.16rc4
nginx 1.17.9
luajit 2.0.5-2 (I have also tried luajit 2.1 0ad60cc and https://github.com/openresty/luajit2/commit/38cb695de87cfeadcba5eeaf57f39e41d529aa1f, with the same results)
gcc 9.2.1 20200130

dylanplecki:
This change fixes the LuaJIT library auto-discovery build regression from a96e99ab1a770d261fa9eecbd9c7d112f39aa34b and adds additional LuaJIT auto-discovery for LuaJIT-2.1.

Fixes #1661.

### DIFF
--- a/config
+++ b/config
@@ -7,80 +7,69 @@ ngx_feature_run=no
 ngx_feature_incs=
 ngx_feature_test=
 
+check_luajit_lib() {
+    luajit_name=$1
+    luajit_desc=$2
+    luajit_inc=$3
+    luajit_lib=$4
+    luajit_ld_opt=$5
+
+    ngx_feature_path="$luajit_inc"
+    ngx_lua_opt_I="-I$luajit_inc"
+
+    if [ "$luajit_lib" != "" ]; then
+        ngx_feature="LuaJIT library in $luajit_lib and headers in $luajit_inc ($luajit_desc)"
+        ngx_lua_opt_L="-L$luajit_lib"
+    else
+        ngx_feature="LuaJIT default library and headers in $luajit_inc ($luajit_desc)"
+        ngx_lua_opt_L=
+    fi
+
+    printf "checking for $ngx_feature include directory ..."
+    if test -d "$luajit_inc"; then
+        # Don't set ngx_found here, wait for auto/feature below to set it.
+        echo " found"
+    else
+        echo " not found"
+        ngx_found=no
+        return
+    fi
+
+    ngx_feature_libs="$ngx_lua_opt_L -l$luajit_name $luajit_ld_opt"
+    if [ $NGX_RPATH = YES ]; then
+        ngx_feature_libs="-R$luajit_lib $ngx_feature_libs"
+    fi
+
+    # ensure that -I$LUAJIT_INC and -L$LUAJIT_LIB come first
+    SAVED_CC_TEST_FLAGS="$CC_TEST_FLAGS"
+    CC_TEST_FLAGS="$ngx_lua_opt_I $CC_TEST_FLAGS"
+    SAVED_NGX_TEST_LD_OPT="$NGX_TEST_LD_OPT"
+    NGX_TEST_LD_OPT="$ngx_lua_opt_L $NGX_TEST_LD_OPT"
+
+    . auto/feature
+
+    # clean up
+    CC_TEST_FLAGS="$SAVED_CC_TEST_FLAGS"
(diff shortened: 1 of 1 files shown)


thibaultcha:
This PR reminds me a lot of #1577 which wasn't accepted, alas.

dylanplecki:
Thanks for the insight @thibaultcha.

Hopefully, we get this regression fix through (or some version of it, #1577 looks interesting). It will impact LuaJIT auto-discovery in the main Nginx Debian repo in the future, though they're currently pinned to an older version of http-lua for now to support non-LuaJIT builds on odd arches.

Anecdotally, I ran into this bug while attempting to update an internal Nginx Debian repo to Nginx v1.19.2 and http-lua v0.10.17, with the regression taking place after v0.10.16. The package maintainers have a [workaround patch](https://salsa.debian.org/nginx-team/nginx/-/blob/28cea613e4d327458cc715425cfa9ee399c68b11/debian/modules/patches/http-lua/discover-luajit-2.1.patch) for v0.10.13 but will break again after http-lua v0.10.16.

thibaultcha:
@dylanplecki Since so many efforts have been invested into optimizing OpenResty's [LuaJIT fork](https://github.com/openresty/luajit2) for OpenResty's use-cases, it is highly advised to run one of the formal [OpenResty releases](https://openresty.org/en/download.html). It is especially more so the case when considering the Nginx and OpenSSL patches needed by OpenResty, as well as the lack of testing for ngx_http_lua on some of these Nginx versions. Can't you use a formal OpenResty release instead?

thibaultcha:
It seems I should have asked this question in #1661 instead.

Issue #1788, opened 2020-09-21
Is there a little bug in function ngx.decode_args？


YiYaLi:
Code:
    local args = ngx.decode_args("a=hello%world")
    for k, v in pairs(args) do
        ngx.say("k = ", k, ", v = ", v)
    end

Result:
    k = a, v = helloworld

I think the result should be:
    k = a, v = hello%world

YiYaLi:
Yeah, I tested it before. The result is different from the character after '%'.

chenxuuu:
随便改了改，应该可以了

suikabreaker:
> 随便改了改，应该可以了

You need to consider the case when src and dst are the same pointer... that's the point of the interface.

#1862

PR #1789, opened 2020-09-25
bugfix: unescape_uri bug with invalid quoted character

chenxuuu:
fix https://github.com/openresty/lua-nginx-module/issues/1788

```lua
local args = ngx.decode_args("a=hello%1world")
for k, v in pairs(args) do
    ngx.say("k = ", k, ", v = ", v)
end
```
output: `k = a, v = helloorld`, some charactes are missing

### DIFF
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -2156,6 +2156,7 @@ ngx_http_lua_unescape_uri(u_char **dst, u_char **src, size_t size,
 
             state = sw_usual;
 
+            *d++ = '%';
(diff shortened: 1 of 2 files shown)


spacewander (changes requested):
Please add test cases to cover the feature.
And we need to update the doc to explain how we handle invalid quoted character, since this behavior is not standardized across different languages (python will keep the `%` unchanged, while js and Go will raise an error).

spacewander:
Surpassed by https://github.com/openresty/lua-nginx-module/pull/1862

PR #1791, opened 2020-09-28
fix negative content-length number

LubinLew:
proxy mode.
download big file (> 4GB), eg [CentOS-8.2.2004-x86_64-dvd1.iso](https://mirror.xtom.com.hk/centos/8.2.2004/isos/x86_64/CentOS-8.2.2004-x86_64-dvd1.iso) size 8231321600.
ngx.resp.get_headers()['Content-Length'] return  **-35861299**

The problem is Integer Overflow. Code is here
```c
lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
```
Can't write test case for this one..

### DIFF
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -492,8 +492,9 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
             if (p == NULL) {
                 return luaL_error(L, "no memory");
             }
+
             len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
-                         r->headers_out.content_length_n) - p;
+                               r->headers_out.content_length_n) - p;
             lua_pushfstring(L, "%s", (char *) p, len);
         } else {
             lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -485,7 +485,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
     {
         extra++;
         lua_pushliteral(L, "content-length");
-        lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
+        if (r->headers_out.content_length_n > NGX_MAX_INT32_VALUE) {
+            u_char *p = ngx_palloc(r->pool, NGX_OFF_T_LEN);
+            if (p == NULL) {
+                return luaL_error(L, "no memory");
+            }
+            size_t len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
+                         r->headers_out.content_length_n) - p;
+            lua_pushfstring(L, "%s", (char *) p, len);
+        } else {
+            lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
+        }
         lua_rawset(L, -3);
     }
 
--- a/src/ngx_http_lua_headers.c
+++ b/src/ngx_http_lua_headers.c
@@ -424,6 +424,8 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
     int                 count = 0;
     int                 truncated = 0;
     int                 extra = 0;
+    u_char             *p = NULL;
+    size_t              len = 0;
 
     n = lua_gettop(L);
 
@@ -486,11 +488,11 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
         extra++;
         lua_pushliteral(L, "content-length");
         if (r->headers_out.content_length_n > NGX_MAX_INT32_VALUE) {
-            u_char *p = ngx_palloc(r->pool, NGX_OFF_T_LEN);
+            p = ngx_palloc(r->pool, NGX_OFF_T_LEN);
             if (p == NULL) {
                 return luaL_error(L, "no memory");
             }
-            size_t len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
+            len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
                          r->headers_out.content_length_n) - p;
             lua_pushfstring(L, "%s", (char *) p, len);
         } else {
--- a/t/016-resp-header.t
+++ b/t/016-resp-header.t
@@ -8,7 +8,7 @@ use Test::Nginx::Socket::Lua;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 3 + 77);
+plan tests => repeat_each() * (blocks() * 3 + 79);
 
 #no_diff();
 no_long_string();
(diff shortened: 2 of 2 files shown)


src/ngx_http_lua_headers.c
```diff
@@ -485,7 +485,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
     {
         extra++;
         lua_pushliteral(L, "content-length");
-        lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
+        if (r->headers_out.content_length_n > NGX_MAX_INT32_VALUE) {
+            u_char *p = ngx_palloc(r->pool, NGX_OFF_T_LEN);
```
spacewander, line 489: Better to move the declaration of `p` to the top of function, to make the code style consistent.

src/ngx_http_lua_headers.c
```diff
@@ -485,7 +485,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
...
+            if (p == NULL) {
+                return luaL_error(L, "no memory");
+            }
+            size_t len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
```
spacewander, line 493: Ditto

spacewander (changes requested):
> Can't write test case for this one..

Maybe you can set the Content-Length via `ngx.header` and get it later?

src/ngx_http_lua_headers.c
```diff
@@ -485,7 +485,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
...
+                         r->headers_out.content_length_n) - p;
+            lua_pushfstring(L, "%s", (char *) p, len);
```
spacewander, line 495: Maybe we should use `%g` instead so that there is no need to allocate temporary string.

LubinLew: > Maybe we should use `%g` instead so that there is no need to allocate temporary string.

[lua_pushfstring](https://www.lua.org/manual/5.1/manual.html#lua_pushfstring) only support these format
- '%s' (null-terminated string)
 - '%f' (inserts a lua_Number)
 - '%p' (inserts a pointer as a hexadecimal numeral)
 - '%d' (inserts an int)
 - '%c' (inserts an int as a character)
What the meaming is  the `%g` ?

spacewander: I remember it is a LuaJIT extensioin.
Anyway, look like `%f` is more suitable. I am not quite sure and you need to give it a try.

spacewander: I changed my mind. Convert integer to double may cause precision lost when the number is really big. Although it is not a real world case, it would be safer to convert it to string first.

spacewander:
@LubinLew 
Test cases are needed.

src/ngx_http_lua_headers.c
```diff
@@ -485,7 +487,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
...
         extra++;
         lua_pushliteral(L, "content-length");
-        lua_pushfstring(L, "%d", (int) r->headers_out.content_length_n);
+        if (r->headers_out.content_length_n > NGX_MAX_INT32_VALUE) {
+            p = ngx_palloc(r->pool, NGX_OFF_T_LEN);
+            if (p == NULL) {
+                return luaL_error(L, "no memory");
+            }
+            len = ngx_snprintf(p, NGX_OFF_T_LEN, "%O",
+                         r->headers_out.content_length_n) - p;
```
doujiang24, line 496: style: the `r->headers_out` should align up with `p`.

src/ngx_http_lua_headers.c
```diff
@@ -485,7 +487,17 @@ ngx_http_lua_ngx_resp_get_headers(lua_State *L)
...
+            lua_pushfstring(L, "%s", (char *) p, len);
+        } else {
```
doujiang24, line 498: style: need a blank line before this line.

doujiang24 (commented):
@LubinLew Thanks for your contribution, also need a test case @spacewander mentioned.

> Maybe you can set the Content-Length via ngx.header and get it later?

LubinLew (commented):
> @LubinLew Thanks for your contribution, also need a test case @spacewander mentioned.
> 
> > Maybe you can set the Content-Length via ngx.header and get it later?

That  not works. The reason is in last comment.

src/ngx_http_lua_headers.c
doujiang24: @LubinLew Missed this comment?

src/ngx_http_lua_headers.c
doujiang24: @LubinLew Well, the `ngx.header['Content-Length']` should not be the right way for the test case.

I test this case at my side, it works.
```
--- config
    location = /big-upstream {
        content_by_lua_block {
            ngx.header['Content-Length'] = math.pow(2, 33) - 1
            ngx.say('hi')
        }
    }

    location = /t {
        proxy_pass http://127.0.0.1:$TEST_NGINX_SERVER_PORT/big-upstream;
        proxy_buffering off;

        header_filter_by_lua_block {
            local hs, err = ngx.resp.get_headers()
            if err then
                ngx.log(ngx.ERR, "err: ", err)
                return ngx.exit(500)
            end

            print("my Content-Length: ", hs["Content-Length"])

            ngx.header['Content-Length'] = 3
        }
    }
--- request
    GET /t
--- response_body
hi
--- no_error_log
[alert]
--- error_log
my Content-Length: 8589934591
upstream prematurely closed connection while sending to client
```

Will you help to add this test case? Thanks.

LubinLew: that works, thank you, But I really not familiar with Test::Nginx, got 2 problems,
1.  I changed 77 -> 79, so plan number can be equal to tested number. Is this right ?
```txt
plan tests => repeat_each() * (blocks() * 3 + 77);
```
2. test failed at TEST 75 and TEST 77.
```txt
[root@test ~]# prove 016-resp-header.t 
016-resp-header.t .. 221/722 
#   Failed test 'TEST 75: exceeding max header limit (default 100) - pattern "lua exceeding response header limit 101 > 100" should match a line in error.log (req 0)'
#   at /usr/local/share/perl5/Test/Nginx/Socket.pm line 1213.

#   Failed test 'TEST 75: exceeding max header limit (default 100) - pattern "lua exceeding response header limit 101 > 100" should match a line in error.log (req 1)'
#   at /usr/local/share/perl5/Test/Nginx/Socket.pm line 1213.

#   Failed test 'TEST 77: exceeding max header limit (custom limit, 3) - pattern "lua exceeding response header limit 4 > 3" should match a line in error.log (req 0)'
#   at /usr/local/share/perl5/Test/Nginx/Socket.pm line 1213.

#   Failed test 'TEST 77: exceeding max header limit (custom limit, 3) - pattern "lua exceeding response header limit 4 > 3" should match a line in error.log (req 1)'
#   at /usr/local/share/perl5/Test/Nginx/Socket.pm line 1213.
016-resp-header.t .. 338/722 WARNING: TEST 3: set response content-length header - unexpected extra bytes after last chunk in response: "lo\x{0a}"
WARNING: TEST 3: set response content-length header - unexpected extra bytes after last chunk in response: "lo\x{0a}"
016-resp-header.t .. 710/722 # Looks like you failed 4 tests of 722.
016-resp-header.t .. Dubious, test returned 4 (wstat 1024, 0x400)
Failed 4/722 subtests 

Test Summary Report
-------------------
016-resp-header.t (Wstat: 1024 Tests: 722 Failed: 4)
  Failed tests:  221, 225, 269, 273
  Non-zero exit status: 4
Files=1, Tests=722, 14 wallclock secs ( 0.08 usr  0.01 sys +  1.12 cusr  0.52 csys =  1.73 CPU)
Result: FAIL

```

doujiang24: 1. yes.
2. I haven't seen the test failure in Travis CI, please make sure using `--with-debug` to build the test openresty.
https://travis-ci.org/github/openresty/lua-nginx-module/jobs/734467043

PS: for the failure in Travis CI, I have already fixed in master branch.

doujiang24:
@LubinLew Merged after minor style tweaks, Thanks for your contribution.

PR #1825, opened 2020-11-07
[WIP]feature: support socket in some block phase

rainingmaster:

Sometime we need use cosocket in `init_by_lua` to get some data from remote before NGINX init.
This feature allow us use cosocket in some block phase, and we should limit its use way.


config
```diff
@@ -298,6 +298,18 @@ HTTP_LUA_SRCS=" \
             $ngx_addon_dir/src/ngx_http_lua_pipe.c \
             "
 
+if [ $EVENT_MODULES =~ ngx_epoll_module ]; then
+    HTTP_LUA_SRCS="$HTTP_LUA_SRCS $ngx_addon_dir/src/event/ngx_http_lua_poll.c"
```
deba12, line 302: ngx_http_lua_epoll.c ?

rainingmaster: Yes, thx for remind.

src/event/ngx_http_lua_epoll.c
```diff
@@ -0,0 +1,211 @@
...
+
+
+static ngx_int_t ngx_http_lua_epoll_init_event(ngx_cycle_t *cycle);
+static ngx_int_t ngx_http_lua_epoll_set_event(ngx_event_t *ev, ngx_int_t event);
+static ngx_int_t ngx_http_lua_epoll_clear_event(ngx_event_t *ev,
+    ngx_int_t event);
+static ngx_int_t ngx_http_lua_epoll_process_event(ngx_http_request_t *r,
+    ngx_msec_t timer);
+
+static int                  ep = -1;
```
agentzh, line 20: top-level variables should never use short names like this.

t/129-ssl-socket.t
```diff
@@ -2834,3 +2829,66 @@ SSL reused session
 [alert]
 [emerg]
 --- timeout: 10
+
+
+
+=== TEST 35: www.google.com in init_worker_by_lua
+--- http_config
+    init_worker_by_lua_block {
```
agentzh, line 2837: I'm afraid it's not a good idea to block init worker hook for long. Because new requests are already coming in. Blocking for long in this phase can be problematic. We need to make it very clear in the documentation at least. Honestly we should really disable such blocking emulations in this init worker context.

t/058-tcp-socket.t
```diff
@@ -4367,3 +4367,64 @@ connect failed: missing the port number
 finish
 --- no_error_log
 [error]
+
+
+
+=== TEST 73: run in init_worker_by_lua
```
agentzh, line 4373: What makes more sense is to test cases for the `init_by_lua*` and `exit_worker_by_lua*` contexts.

agentzh (changes requested):

src/ngx_http_lua_socket_tcp.c
```diff
@@ -685,7 +707,10 @@ ngx_http_lua_socket_tcp_connect_helper(lua_State *L,
     ngx_memzero(&url, sizeof(ngx_url_t));
     url.url = host;
     url.default_port = port;
-    url.no_resolve = 1;
+
+    if (ctx->context & NGX_HTTP_LUA_CONTEXT_YIELDABLE) {
```
doujiang24, line 711: need some comments for this change.

doujiang24:
As @agentzh said, `init_by_lua` and `exit_by_lua` are more meaningful for this feature, but they doesn't work now.
for the `init_by_lua` we need to create fake request.

src/ngx_http_lua_event.h
```diff
@@ -0,0 +1,178 @@
...
+    ngx_int_t  (*clear_event)(ngx_event_t *ev, ngx_int_t event);
+    ngx_int_t  (*process_event)(ngx_http_request_t *r, ngx_msec_t timer);
+} ngx_http_lua_event_actions_t;
+
+
+extern ngx_http_lua_event_actions_t  ngx_http_lua_event_actions;
+
+extern ngx_http_lua_event_actions_t  ngx_http_lua_epoll;
+extern ngx_http_lua_event_actions_t  ngx_http_lua_poll;
+extern ngx_http_lua_event_actions_t  ngx_http_lua_kqueue;
```
zhuizhuhaomeng, line 26:
```suggestion
#if (NGX_HAVE_EPOLL)
extern ngx_http_lua_event_actions_t  ngx_http_lua_epoll;
#endif
#if (NGX_HAVE_POLL)
extern ngx_http_lua_event_actions_t  ngx_http_lua_poll;
#endif
#if (NGX_HAVE_KQUEUE)
extern ngx_http_lua_event_actions_t  ngx_http_lua_kqueue;
#endif
```

src/ngx_http_lua_module.c
```diff
@@ -890,15 +911,233 @@ ngx_http_lua_init(ngx_conf_t *cf)
         ngx_http_lua_assert(lmcf->lua != NULL);
 
         if (!lmcf->requires_shm && lmcf->init_handler) {
+            cycle = cf->cycle;
```
zhuizhuhaomeng, line 914: we should move all these codes into init_handler

src/ngx_http_lua_util.h
```diff
@@ -40,6 +40,14 @@
                                 | NGX_HTTP_LUA_CONTEXT_SSL_CERT              \
                                 | NGX_HTTP_LUA_CONTEXT_SSL_SESS_FETCH)
 
+#define NGX_HTTP_LUA_CONTEXT_BLOCKED_COSOCKET                                \
+                                        (NGX_HTTP_LUA_CONTEXT_INIT    \
```
zhuizhuhaomeng, line 44: need align

t/058-tcp-socket.t
```diff
@@ -4367,3 +4367,65 @@ connect failed: missing the port number
...
```
zhuizhuhaomeng, line 4373:
```suggestion
=== TEST 73: run in init_by_lua
```

t/129-ssl-socket.t
```diff
@@ -2834,3 +2829,67 @@ SSL reused session
...
+--- error_log_like chop
+\Aconnected: 1
+ssl handshake: userdata
+sent http request: 59 bytes.
+received: HTTP/1.1 (?:200 OK|302 Found)
+close: 1 nil
+\z
+[error]
+--- timeout: 5
+--- ONLY
```
zhuizhuhaomeng, line 2895:
```suggestion
--- ONLY
```
```suggestion

```

PR #1828, opened 2020-11-23
feature: add privileged agent connections.

yaolaiyaoqu:
add privileged agent connections argument in ngx_http_lua_ffi_enable_privileged_agent.

### DIFF
--- a/src/ngx_http_lua_worker.c
+++ b/src/ngx_http_lua_worker.c
@@ -98,7 +98,7 @@ ngx_http_lua_ffi_get_process_type(void)
 
 
 int
-ngx_http_lua_ffi_enable_privileged_agent(char **err)
+ngx_http_lua_ffi_enable_privileged_agent(char **err, unsigned int connections)
 {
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_worker.c
```diff
@@ -107,6 +107,7 @@ ngx_http_lua_ffi_enable_privileged_agent(char **err)
                                            ngx_core_module);
 
     ccf->privileged_agent = 1;
+    ccf->privileged_agent_connections = connections;
```
spacewander, line 110: This line can't even be built.

spacewander (changes requested):

src/ngx_http_lua_worker.c
doujiang24: It depends on the sister PR: https://github.com/openresty/openresty/pull/686

doujiang24 (approved):

zhuizhuhaomeng (approved):

PR #1835, opened 2020-12-03
bugfix: memory leak of debug log

hawkxiang:
If turn on nginx debug log `./configure --with-debug ...`, but not select `debug_http` log level. 
ngx_log_debug1(NGX_LOG_DEBUG_HTTP, ...) will be ignored, malloc_trim() can't be executed as function parameter.
And memory isn't returned to OS immediately. This bug may cause serious memory leak, OOM.
截屏2020-12-03 下午5 52 27
Simply,  `ngx_log_debug1` is optional, but `malloc_trim` obligatory. 


hawkxiang:
I don't think that CI failed is caused by the PR code. 
@spacewander

spacewander (approved):

PR #1880, opened 2021-05-06
change: changed from graceful shutdown to abort when lua_resume return LUA_ERRMEM.

zhuizhuhaomeng:
If the worker voluntary enter into graceful shutdown state, the master will not respawn worker with the same worker id until the worker process exit.
when reuseport is on, the new connections assigned to the voluntary graceful shutdown worker will not be accepted.
when reuseport is off, the new connections will be accepted by other processes which are still working. But if all workers are in voluntary shutdown state, the new connections will not be accepted.

So the worker should not enter into the voluntary graceful shutdown state when LUA_ERRMEM occurs.

### DIFF
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -1471,7 +1471,7 @@ user_co_done:
                 break;
 
             case LUA_ERRMEM:
-                err = "memory allocation error when running lua coroutine";
+                err = "[lua] memory allocation error";
                 ngx_log_error(NGX_LOG_ALERT, r->connection->log, 0, err);
                 abort();
                 break;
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_util.c
```diff
@@ -1471,8 +1471,9 @@ ngx_http_lua_run_thread(lua_State *L, ngx_http_request_t *r,
                 break;
 
             case LUA_ERRMEM:
-                err = "memory allocation error";
-                ngx_quit = 1;
+                err = "memory allocation error when running lua coroutine";
```
doujiang24, line 1474:
```suggestion
                err = "[lua] memory allocation error";
```

spacewander (approved):

PR #1899, opened 2021-06-29
change: removed the openssl 1.0.2 in CI since it has been EOL ove a year, fix #1897.

### LINKED ISSUE #1897: The CI is broken because openresty.org is signed by Let's Encrypt
Currently, CI with openssl 1.0.2h is broken:
https://travis-ci.com/github/openresty/lua-nginx-module/builds/231035966

I investigated it and found out it is because the certificate of openresty.org is signed by Let's Encrypt now. See https://community.letsencrypt.org/t/openssl-client-compatibility-changes-for-let-s-encrypt-certificates/143816 for the details.

CC @doujiang24

doujiang24:


spacewander:
CentOS 7 is still using openssl 1.0.2?

doujiang24:
yes, centos 7 is still using 1.0.2.
committed another fix after investigating the behavior in centos 7.

doujiang24:
Oh, my bad. The new fix doesn't work since the issue is not due to the root certificates.
And the curl in centos7 is not using OpenSSL.

by the way, openresty.org has used the LE certificate for a long time.

spacewander (approved):

PR #1907, opened 2021-08-17
chore: check context in ngx.timer.* for exit process

rainingmaster:
Since nginx has been confirmed that all timers have been cleaned up when exit worker is executed, all timers will no longer be executed in exit worker phase. 
Reference https://github.com/nginx/nginx/blob/f02e2a734ef472f0dcf83ab2e8ce96d1acead8a5/src/os/unix/ngx_process_cycle.c#L715

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1672,6 +1672,8 @@ For example,
  }
 ```
 
+It's not allowed to create a timer (even a 0-delay timer) here since it runs after all timers have been processed.
+
 This directive was first introduced in the `v0.10.18` release.
 
 [Back to TOC](#directives)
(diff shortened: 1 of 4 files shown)


README.markdown
```diff
@@ -8539,6 +8542,8 @@ table. It won't share the same `ngx.ctx` with the Lua handler creating the timer
 If you need to pass data from the timer creator to the timer handler, please
 use the extra parameters of `ngx.timer.at()`.
 
+Since nginx has been confirmed that all timers have been cleaned up when exit worker is executed, all timers will no longer be executed in exit worker phase. Reference <https://github.com/nginx/nginx/blob/f02e2a734ef472f0dcf83ab2e8ce96d1acead8a5/src/os/unix/ngx_process_cycle.c#L715>.
```
doujiang24, line 8545:
```
It's not allowed to create a timer (even a 0-delay timer) here since it runs after all timers have been processed.
```
better move this note to here?
https://github.com/openresty/lua-nginx-module/blob/e718d797cf28988251c329e247c7cd475d7e2bd9/README.markdown#exit_worker_by_lua_block

doujiang24: And we can keep the reference only in the code comments.

doujiang24 (approved):
One ci is failed, but it is not related to this change.
https://app.travis-ci.com/github/openresty/lua-nginx-module/jobs/531916997#L922

spacewander (approved):

Issue #38, opened 2011-05-20
ngx.location.capture header inheritance


rik1083:
Hi
I spoke with you over nginx mailing list yesterday
I was having problem where the request headers weren't getting passed through on a sub-request
Anyway, I got to the bottom of it:
Headers ARE passed if 'method' is left unset (it defaults to GET I assume)
But if method IS set, to ngx.HTTP_GET or ngx.HTTP_HEAD (I didnt test POST) then they are not passed through

Hope this helps & keep up the good work

PR #1908, opened 2021-08-18
feature: implemented keepalive pooling in 'balancer_by_lua*'. (#38)

zhuizhuhaomeng:
Co-authored-by: Thibault Charbonnier <thibaultcha@me.com>

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -1144,6 +1144,7 @@ Directives
 * [log_by_lua_file](#log_by_lua_file)
 * [balancer_by_lua_block](#balancer_by_lua_block)
 * [balancer_by_lua_file](#balancer_by_lua_file)
+* [balancer_keepalive](#balancer_keepalive)
 * [lua_need_request_body](#lua_need_request_body)
 * [ssl_client_hello_by_lua_block](#ssl_client_hello_by_lua_block)
 * [ssl_client_hello_by_lua_file](#ssl_client_hello_by_lua_file)
@@ -2708,6 +2709,29 @@ Equivalent to [balancer_by_lua_block](#balancer_by_lua_block), except that the f
 When a relative path like `foo/bar.lua` is given, they will be turned into the absolute path relative to the `server prefix` path determined by the `-p PATH` command-line option while starting the Nginx server.
 
 This directive was first introduced in the `v0.10.0` release.
+
+[Back to TOC](#directives)
+
+balancer_keepalive
+------------------
+
+**syntax:** *balancer_keepalive &lt;total-connections&gt;*
+
+**context:** *upstream*
+
+**phase:** *loading-config*
+
+The `total-connections` parameter sets the maximum number of idle
+keepalive connections to upstream servers that are preserved in the cache of
+each worker process. When this number is exceeded, the least recently used
+connections are closed.
+
+It should be particularly noted that the keepalive directive does not limit the
+total number of connections to upstream servers that an nginx worker process
+can open. The connections parameter should be set to a number small enough to
+let upstream servers process new incoming connections as well.
+
+This directive was first introduced in the `v0.10.21` release.
 
 [Back to TOC](#directives)
 
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -2288,6 +2288,26 @@ When a relative path like <code>foo/bar.lua</code> is given, they will be turned
 
 This directive was first introduced in the <code>v0.10.0</code> release.
 
+== balancer_keepalive ==
+
+'''syntax:''' ''balancer_keepalive <total-connections>''
+
+'''context:''' ''upstream''
+
+'''phase:''' ''loading-config''
+
+The <code>total-connections</code> parameter sets the maximum number of idle
+keepalive connections to upstream servers that are preserved in the cache of
+each worker process. When this number is exceeded, the least recently used
+connections are closed.
+
+It should be particularly noted that the keepalive directive does not limit the
+total number of connections to upstream servers that an nginx worker process
+can open. The connections parameter should be set to a number small enough to
+let upstream servers process new incoming connections as well.
+
+This directive was first introduced in the <code>v0.10.21</code> release.
+
 == lua_need_request_body ==
 
 '''syntax:''' ''lua_need_request_body <on|off>''
--- a/src/ngx_http_lua_balancer.c
+++ b/src/ngx_http_lua_balancer.c
@@ -15,10 +15,34 @@
 #include "ngx_http_lua_util.h"
 #include "ngx_http_lua_directive.h"
 
+#define NGX_BALANCER_DEF_HOST_LEN  32
+typedef struct {
+    ngx_queue_t                    queue;
+    ngx_queue_t                    hnode;
+    ngx_uint_t                     hash;
+    ngx_http_lua_srv_conf_t       *lscf;
+    ngx_connection_t              *connection;
+    socklen_t                      socklen;
+    ngx_sockaddr_t                 sockaddr;
+    ngx_str_t                      host;
+    /* try to avoid allocating memory from the connection pool */
+    u_char                         host_data[NGX_BALANCER_DEF_HOST_LEN];
+} ngx_http_lua_balancer_ka_item_t; /*balancer keepalive item*/
+
 
 struct ngx_http_lua_balancer_peer_data_s {
-    /* the round robin data must be first */
-    ngx_http_upstream_rr_peer_data_t    rrp;
+    ngx_uint_t                          keepalive_requests;
+    ngx_msec_t                          keepalive_timeout;
+
+    void                               *data;
+
+    ngx_event_get_peer_pt               original_get_peer;
+    ngx_event_free_peer_pt              original_free_peer;
+
+#if (NGX_HTTP_SSL)
+    ngx_event_set_peer_session_pt       original_set_session;
+    ngx_event_save_peer_session_pt      original_save_session;
+#endif
 
     ngx_http_lua_srv_conf_t            *conf;
     ngx_http_request_t                 *request;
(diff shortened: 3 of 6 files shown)


src/ngx_http_lua_balancer.c
```diff
@@ -450,64 +793,59 @@ ngx_http_lua_balancer_save_session(ngx_peer_connection_t *pc, void *data)
...
     ngx_url_t              url;
     ngx_http_lua_ctx_t    *ctx;
     ngx_http_upstream_t   *u;
 
-    ngx_http_lua_main_conf_t           *lmcf;
     ngx_http_lua_balancer_peer_data_t  *bp;
 
     if (r == NULL) {
-        *err = "no request found";
+        p = ngx_snprintf(errbuf, *errbuf_size, "no request found");
```
doujiang24, line 816: why do we need this change?

src/ngx_http_lua_balancer.c
```diff
@@ -15,10 +15,32 @@
...
+typedef struct {
+    ngx_queue_t                    queue;
+    ngx_queue_t                    hnode;
+    ngx_uint_t                     hash;
+    ngx_http_lua_srv_conf_t       *lscf;
+    ngx_connection_t              *connection;
+    socklen_t                      socklen;
+    ngx_sockaddr_t                 sockaddr;
+    ngx_str_t                      host;
+    u_char                         hostbuf[32];
```
doujiang24, line 27: better use a meaning full macro for the hardcode `32`.

And maybe host_data may be a better name? it's not actually a buffer.

doujiang24: And better add a comment for this field.
```
try to avoid allocating memory from the connection pool
```

doujiang24 (approved):

README.markdown
```diff
@@ -2597,6 +2598,29 @@ This directive was first introduced in the `v0.10.0` release.
...
+keepalive connections to upstream servers that are preserved in the cache of
+each worker process. When this number is exceeded, the least recently used
+connections are closed.
+
+It should be particularly noted that the keepalive directive does not limit the
+total number of connections to upstream servers that an nginx worker process
+can open. The connections parameter should be set to a number small enough to
+let upstream servers process new incoming connections as well.
+
+This directive was first introduced in the `v0.10.19.5` release.
```
xiaocang, line 2620: typo here?
`v0.10.19.5` -> `v0.10.19`

doc/HttpLuaModule.wiki
```diff
@@ -2186,6 +2186,26 @@ When a relative path like <code>foo/bar.lua</code> is given, they will be turned
...
+keepalive connections to upstream servers that are preserved in the cache of
+each worker process. When this number is exceeded, the least recently used
+connections are closed.
+
+It should be particularly noted that the keepalive directive does not limit the
+total number of connections to upstream servers that an nginx worker process
+can open. The connections parameter should be set to a number small enough to
+let upstream servers process new incoming connections as well.
+
+This directive was first introduced in the <code>v0.10.19.5</code> release.
```
xiaocang, line 2207: Ditto

src/ngx_http_lua_balancer.c
```diff
@@ -520,22 +860,91 @@ ngx_http_lua_ffi_balancer_set_current_peer(ngx_http_request_t *r,
 
     if (ngx_parse_url(r->pool, &url) != NGX_OK) {
         if (url.err) {
-            *err = url.err;
+            p = ngx_snprintf(errbuf, *errbuf_size, "%s", url.err);
```
xiaocang, line 863: `errbuf` is undeclared before

xiaocang (changes requested):

src/ngx_http_lua_balancer.c
```diff
@@ -415,12 +574,205 @@ ngx_http_lua_balancer_free_peer(ngx_peer_connection_t *pc, void *data,
...
+    if (c->close || c->read->timedout) {
+        goto close;
+    }
+
+    n = recv(c->fd, buf, 1, MSG_PEEK);
+
+    if (n == -1 && ngx_socket_errno == NGX_EAGAIN) {
+        ev->ready = 0;
+
+        if (ngx_handle_read_event(c->read, 0) == NGX_OK) {
```
doujiang24, line 759:
```suggestion
        if (ngx_handle_read_event(c->read, 0) != NGX_OK) {
```

src/ngx_http_lua_balancer.c
```diff
@@ -799,4 +1184,162 @@ ngx_http_lua_ffi_balancer_recreate_request(ngx_http_request_t *r,
...
+    {
+        item = ngx_queue_data(q, ngx_http_lua_balancer_ka_item_t, hnode);
+        if (item->hash != hash) {
+            continue;
+        }
+
+        if (ngx_memn2cmp((u_char *) &item->sockaddr,
+                         (u_char *) sockaddr,
+                         item->socklen, socklen) == 0
+            && name->len == item->host.len
```
doujiang24, line 1317: maybe we should move these length compare expressions at the front?
I'm not sure if the C compiler is clever enough to do such optimizations.

src/ngx_http_lua_balancer.c
```diff
@@ -799,4 +1184,162 @@ ngx_http_lua_ffi_balancer_recreate_request(ngx_http_request_t *r,
...
+        name = u->ssl_name;
+    }
+
+    if (name.len == 0) {
+        goto done;
+    }
+
+    /*
+     * ssl name here may contain port, notably if derived from $proxy_host
+     * or $http_host; we have to strip it
```
doujiang24, line 1244: better add an example of the `ssl name` value here.

Triple-Z:
Any update on this issue?

tzssangglass:
hi @zhuizhuhaomeng, we found an edge bug in the keepalive pool.

in short, in the balancer phase, when obtaining a connection from the upstream connection pool, the cached attribute of the peer connection is set to 1(`pc->cached = 1`;), indicating that the connection is obtained from the cache.

if an error occurs during the use of this connection, such as "upstream prematurely closed connection" the system will increase the tries attribute of the peer connection by executing `u->peer.tries++`.

tries represents the maximum number of attempts to connect to an upstream server. It equals the normal 1 try + set_more_tries.

the occurrence of `u->peer. tries++` is unexpected and it results in the actual retry count exceeding `set_more_tries`.

here is a reproduce test case:

```perl
# vim:set ft= ts=4 sw=4 et:

use Test::Nginx::Socket::Lua;
use Cwd qw(cwd);

log_level('info');
repeat_each(1);

plan tests => repeat_each() * (blocks() * 6);

my $pwd = cwd();

no_long_string();

check_accum_error_log();
run_tests();

__DATA__

=== TEST 1: sanity
--- http_config
    lua_shared_dict request_counter 1m;
    upstream my_upstream {
        server 127.0.0.1;
        balancer_by_lua_block {

            if not ngx.ctx.tries then
                ngx.ctx.tries = 0
            end

            ngx.ctx.tries = ngx.ctx.tries + 1
            ngx.log(ngx.INFO, "tries ", ngx.ctx.tries)

            if ngx.ctx.tries == 1 then
                balancer.set_more_tries(5)
            end
            
            local host = "127.0.0.1"
            local port = 8090;

            local ok, err = balancer.set_current_peer(host, port)
            if not ok then
                ngx.log(ngx.ERR, "failed to set the current peer: ", err)
                return ngx.exit(500)
            end

            balancer.set_timeouts(60000, 60000, 60000)

            local ok, err = balancer.enable_keepalive(60, 100)
            if not ok then
                ngx.log(ngx.ERR, "failed to enable keepalive: ", err)
                return ngx.exit(500)
            end
        }
    }

    server {
        listen 0.0.0.0:8090;
        location /hello {
            content_by_lua_block{                
                local request_counter = ngx.shared.request_counter
                local first_request = request_counter:get("first_request")
                if first_request == nil then
                    request_counter:set("first_request", "yes")
                    ngx.print("hello")
                else
                    ngx.exit(ngx.HTTP_CLOSE)
                end
            }
        }
    }
--- config
    location = /t {
        proxy_pass http://my_upstream;
        proxy_set_header Connection "keep-alive";
        
        rewrite_by_lua_block {
           ngx.req.set_uri("/hello")
        }
    }
--- pipelined_requests eval
["GET /t HTTP/1.1" , "GET /t HTTP/1.1"]
--- response_body eval
["hello", qr/502/]
--- error_code eval
[200, 502]
--- error_log eval
qr/tries 7/
```

in this case:
1. the first request is normal. balancer creates a connection with the upstream and puts the connection into the balancer connection pool after the first request end
2. the second request, pulling a healthy connection from the balancer connection pool and using it, here would mark connection as cached:

	https://github.com/openresty/lua-nginx-module/blob/6738c3a3b0d62b98ff35acf7f7c4a191207da918/src/ngx_http_lua_balancer.c#L1480

 and the upstream shut down the connection by `ngx.exit(ngx.HTTP_CLOSE)`, this is 1 normal try.
3. enter the retry phase, the shutdown just upstream is marked by nginx as `NGX_HTTP_UPSTREAM_FT_ERROR`, and would enter the follow code:
```C
    if (u->peer.cached && ft_type == NGX_HTTP_UPSTREAM_FT_ERROR) {
        /* TODO: inform balancer instead */
        u->peer.tries++;
    }
```
https://github.com/nginx/nginx/blob/e734df6664e70f118ca3140bcef6d4f1750fa8fa/src/http/ngx_http_upstream.c#L4366-L4369

4. `u->peer.tries++` increases the retry times by one.

it seems that Nginx believes that a cached connection improperly closed upstream is not a real retry, so it compensates for a retry.

in the upstream shutdown scenario, this mechanism drains the connections in the balancer connection pool and tries as many times as set by `set_more_tries`.

this results in executing `balancer_by_lua_block` over `set_more_tries`.

tzssangglass:
one way to fix this is to change this mechanism and implement `notify` for `ngx_http_lua_balancer`, like this: https://github.com/Kong/openresty-patches-review/pull/1/files

the another way is not to handle it, we just need to know that it exists, and we can retrieve the `cached` of connection to determine whether the connection was obtained from the balancer connection pool.

tzssangglass:
cc @dndx

zhuizhuhaomeng:
@tzssangglass would you please prepare a patch for this issue?

tzssangglass:
I'd like to know which of the above fixes you prefer

zhuizhuhaomeng:
I prefer the first one(kong's implementation).

PR #1924, opened 2021-09-24
bugfix: we should ignore match limit in DFA mode.

balusch:
lua_regex_match_limit takes effect globally for all regexes used in ngx.re
api. However, the match_limit field of pcre_extra block is not supported in
pcre_dfa_exec()(so does match_limit_recursion), because it is meaningless
in DFA matching. And pcre_dfa_exec() will return -18(PCRE_ERROR_DFA_UMLIMIT)
when the directive is used. Here we just ignore the directive in DFA mode.

### DIFF
--- a/t/120-re-find.t
+++ b/t/120-re-find.t
@@ -917,3 +917,35 @@ to: 4
 pos: 5
 --- no_error_log
 [error]
+
+
+
+=== TEST 32: ignore match limit in DFA mode
+--- http_config
+    lua_regex_match_limit 1;
+--- config
+    location /re {
+        content_by_lua_block {
+            local s = "This is <something> <something else> <something further> no more"
+            local from, to, err = ngx.re.find(s, "<.*>", "d")
+            if from then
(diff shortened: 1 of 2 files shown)


t/120-re-find.t
```diff
@@ -917,3 +917,35 @@ to: 4
...
 [error]
+
+
+
+=== TEST 32: ignore match limit in DFA mode
+--- http_config
+    lua_regex_match_limit 1;
+--- config
+    location /re {
+        content_by_lua '
```
doujiang24, line 928: we should use `content_by_lua_block` in new tests.

doujiang24 (commented):
just one style issue, otherwise, it looks good to me.

t/120-re-find.t
balusch: done

zhuizhuhaomeng (approved):

doujiang24:
@balusch Thanks for your nice contribution!

PR #1926, opened 2021-09-28
change: do not need to create the Lua request ctx data table from C

doujiang24:
since we already changed to create the table from Lua in resty.core.ctx.

related change:
https://github.com/openresty/lua-resty-core/commit/2d667ec81497a7a2ec10bd6b7d1f27509637c928


doujiang24:
Notice: people who use the raw ffi C api to build their own Lua library(just like resty.core.ctx) need to require resty.core.ctx first, or added the ctx data table to registry table by themself(just like resty.core.ctx does).

zhuizhuhaomeng (approved):

PR #1931, opened 2021-10-12
bugfix: nginx crash when resolve an not exist domain in thread create…

zhuizhuhaomeng:
…d by ngx.thread.spawn.

https://github.com/openresty/lua-nginx-module/issues/1915
When got the dns reply, the ngx.thread.wait return and call ngx.exit.
And then all the three 'ngx_resolver_ctx_t *ctx' will be freed. First free by ngx_http_lua_socket_resolve_handler
, the other two free by the ngx_http_lua_udp_resolve_cleanup.

But the ctx was added to a link list and the freed ctx was process again by the function ngx_resolver_process_a.

the backtrace free the first ctx
(gdb) bt
(gdb) c

the backtrace free the other ctx
Continuing.
(gdb) bt

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,8 +59,6 @@ services:
  - mysql
 
 before_install:
-  - sudo apt update
-  - sudo apt install --only-upgrade ca-certificates
   - '! grep -n -P ''(?<=.{80}).+'' --color `find src -name ''*.c''` `find . -name ''*.h''` || (echo "ERROR: Found C source lines exceeding 80 columns." > /dev/stderr; exit 1)'
   - '! grep -n -P ''\t+'' --color `find src -name ''*.c''` `find . -name ''*.h''` || (echo "ERROR: Cannot use tabs." > /dev/stderr; exit 1)'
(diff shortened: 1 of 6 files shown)


src/ngx_http_lua_socket_tcp.c
```diff
@@ -6084,10 +6075,7 @@ ngx_http_lua_tcp_resolve_cleanup(void *data)
         return;
     }
 
-    /* just to be safer */
-    rctx->handler = ngx_http_lua_socket_empty_resolve_handler;
-
-    ngx_resolve_name_done(rctx);
+    rctx->handler = ngx_resolve_name_done;
```
doujiang24, line 6078: better add a comment `postpone free the rctx in the handler`.

.travis.yml
```diff
@@ -59,6 +59,8 @@ services:
  - mysql
 
 before_install:
+  - sudo apt update
+  - sudo apt install --only-upgrade ca-certificates
```
doujiang24, line 63: better move these changes to another PR.

doujiang24 (approved):
LGTM.
Also, seems it worthing some more tests that resolve across requests?

Issue #1954, opened 2021-10-28
New atomic functions(cas, cog) for shm


Miniwoffer:
## Proposal
Add two new functions to shm

### ngx.shared.DICT.cog
-------------------
**syntax:** *value, flags = ngx.shared.DICT:cog(key, old_value?, old_flags?)*

Similar to the get method, but only returns if
`old_value` or `old_flags` do not match.

If `old_value` or `old_flags` is `nil`
it will be ignored when comparing.

In case of match, `nil, false` will be returned.

### ngx.shared.DICT.cas
-------------------
**syntax:** *success, err, forcible = ngx.shared.DICT:cas(key, old_value?, old_flags?, value?, flags?, exptime?)*

Conditionally sets key-value pair in shm.

If `old_value` or `old_flags` is `nil` it will
be ignored.

If either `value` or `flags` is `nil` it will
remain unchanged. If both are `nil`, key-value pair will be deleted.

In case of mismatch, `false, false` will be returned.

### cog
``` lua
local lru = require "resty.lrucache".new(100)
local cjson = require "cjson.safe"
local shm = ngx.shared.config

local function get(key)
        local lru_data, _, lru_flag = lru:get(key)

        local shm_data, shm_flag = shm:cog(key, nil, lru_flag)
        if not shm_data then
                if shm_flag == false then -- boolean false returned on no-error
                        return lru_data
                end

                return nil, shm_flag
        end

        local val, err = cjson.decode(shm_data)
        if not val then
                return nil, err
        end

        lru:set(key, val, nil, shm_flag)

        return val, shm_flag
end
```
### cas
``` lua
local shm = ngx.shared.config
local function multiply(key, factor)
        for _=1,100 do
                local val, flags = shm:get(key)

                if not val then
                        return shm:set(key, 1, nil, 1)
                end

                local ok, err = shm:cas(key, nil, flags, val * factor, flags+1)

                if ok then
                        return ok
                end

                if err ~= false then
                        return nil, err
                end

                ngx.sleep(0.0001)
        end

        return nil, "out of tries"
end
```
## Pull requests
 - lua-nginx-module https://github.com/openresty/lua-nginx-module/pull/1955
 - stream-lua-nginx https://github.com/openresty/stream-lua-nginx-module/pull/274
 - lua-resty-core https://github.com/openresty/lua-resty-core/pull/368

PR #1955, opened 2021-10-28
New atomic functions(cas, cog) for shm

Miniwoffer:

See: https://github.com/openresty/lua-nginx-module/issues/1954

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,11 +88,11 @@ install:
   - git clone https://github.com/openresty/rds-json-nginx-module.git ../rds-json-nginx-module
   - git clone https://github.com/openresty/srcache-nginx-module.git ../srcache-nginx-module
   - git clone https://github.com/openresty/redis2-nginx-module.git ../redis2-nginx-module
-  - git clone https://github.com/openresty/lua-resty-core.git ../lua-resty-core
+  - git clone -b atomic_shm https://github.com/webcore-no/lua-resty-core.git ../lua-resty-core
   - git clone https://github.com/openresty/lua-resty-lrucache.git ../lua-resty-lrucache
   - git clone https://github.com/openresty/lua-resty-mysql.git ../lua-resty-mysql
   - git clone https://github.com/openresty/lua-resty-string.git ../lua-resty-string
-  - git clone https://github.com/openresty/stream-lua-nginx-module.git ../stream-lua-nginx-module
+  - git clone -b atomic_shm https://github.com/webcore-no/stream-lua-nginx-module.git ../stream-lua-nginx-module
   - git clone -b v2.1-agentzh https://github.com/openresty/luajit2.git luajit2
 
 before_script:


xiaocang:
Hi, @Miniwoffer I was wondering what the `cas` and `cog` abbreviations refer to? seems not very common

Miniwoffer:
cas(compare and swap)
cog(compare or get)

cas refers to the atomic operation(https://en.wikipedia.org/wiki/Compare-and-swap), and is also used by memcached amogst others.
cog is not a common abbreviations, just something derived from cas.

t/167-atomic-shdict.t
```diff
@@ -0,0 +1,658 @@
...
+            ngx.say(val, " ", type(val), " ", flags )
+        }
+    }
+--- request
+GET /test
+--- response_body
+nil nil nil
+--- no_error_log
+[error]
+
```
xiaocang, line 656: Maybe we should remove the extra blank lines here

t/167-atomic-shdict.t
```diff
@@ -0,0 +1,658 @@
...
+        }
+    }
+--- request
+GET /test
+--- response_body
+32 number false false
+--- no_error_log
+[error]
+
+
```
xiaocang, line 63: There are always only three blank lines between two use cases
maybe you should lint the code with `reindex` tool, https://github.com/openresty/openresty-devel-utils/blob/master/reindex

README.markdown
```diff
@@ -6884,6 +6888,26 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+
+Similar to the [get](#ngxshareddictget) method, but only returns if
+`old_value` or `old_flags` do not match.
+
+If `old_value` or `old_flags` is `nil`
+it will be ignored when comparing.
+
+In case of match, `nil, false` will be returned.
+
+This method was first introduced in the `0.XX.X` release.
```
xiaocang, line 6905: `0.XX.X` -> `0.10.21`

README.markdown
```diff
@@ -6984,6 +7008,28 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+
+If `old_value` or `old_flags` is `nil` it will
+be ignored.
+
+If either `value` or `flags` is `nil` it will
+remain unchanged. If both are `nil`, key-value pair will be deleted.
+
+In case of mismatch, `false, false` will be returned.
+
+This method was first introduced in the `0.XX.X` release.
```
xiaocang, line 7027: Ditto

t/167-atomic-shdict.t
```diff
@@ -0,0 +1,658 @@
+# vim:set ft= ts=4 sw=4 et fdm=marker:
+use Test::Nginx::Socket::Lua;
+
+#worker_connections(1014);
+#master_process_enabled(1);
+#log_level('warn');
+
+#repeat_each(2);
```
xiaocang, line 8: It looks like this test case can be executed repeatedly, maybe we could uncomment this line

xiaocang (changes requested):

xiaocang:
LGTM.

@zhuizhuhaomeng @doujiang24 could you help with this PR ?

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+
+    dd("lookup returns %d", (int) rc);
+
+    if ((match_flags || match_value) &&
+        ((rc == NGX_DECLINED || rc == NGX_DONE) ||
+         (match_flags && old_user_flags != (int) sd->user_flags) ||
+         (match_value && (old_value_type != sd->value_type ||
+                old_str_value_len != sd->value_len ||
+                ngx_memcmp(old_str_value_buf, sd->data + key_len,
+                           (size_t) old_str_value_len)))))
```
doujiang24, line 1410: this is too huge, maybe we can split them into two `if` conditions?

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+#if 1
+    ngx_http_lua_shdict_expire(ctx, 1);
+#endif
+
+    rc = ngx_http_lua_shdict_lookup(zone, hash, key, key_len, &sd);
+
+    dd("lookup returns %d", (int) rc);
+
+    if ((match_flags || match_value) &&
+        ((rc == NGX_DECLINED || rc == NGX_DONE) ||
```
doujiang24, line 1405: maybe `rc != NGX_OK` can be better?

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+
+#if 1
+    ngx_http_lua_shdict_expire(ctx, 1);
+#endif
+
+    rc = ngx_http_lua_shdict_lookup(zone, hash, key, key_len, &sd);
+
+    dd("lookup returns %d", (int) rc);
+
+    if ((match_flags || match_value) &&
```
doujiang24, line 1404: style: `&&` should move the front of the next line.

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+        ((rc == NGX_DECLINED || rc == NGX_DONE) ||
+         (match_flags && old_user_flags != (int) sd->user_flags) ||
+         (match_value && (old_value_type != sd->value_type ||
+                old_str_value_len != sd->value_len ||
+                ngx_memcmp(old_str_value_buf, sd->data + key_len,
+                           (size_t) old_str_value_len)))))
+    {
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+        *match = 1;
```
doujiang24, line 1413: does this mean not matching? better change the name to unmatch, if yes.

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+        return NGX_DECLINED;
+    }
+
+    if (rc == NGX_OK || rc == NGX_DONE) {
+        if (value_type == LUA_TNIL && !set_user_flags) {
+            goto remove;
+        }
+
+        if (set_user_flags && value_type == LUA_TNIL) {
```
doujiang24, line 1422: better add a comment for this branch, it means do not change the value, right?

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+        {
+
+            ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ctx->log, 0,
+                           "lua shared dict set: found old entry and value "
+                           "size matched, reusing it");
+
+            ngx_queue_remove(&sd->queue);
+            ngx_queue_insert_head(&ctx->sh->lru_queue, &sd->queue);
+
+
```
doujiang24, line 1454: style: one blank line is enough.

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+        ngx_queue_remove(&sd->queue);
+
+        node = (ngx_rbtree_node_t *)
+                   ((u_char *) sd - offsetof(ngx_rbtree_node_t, color));
+
+        ngx_rbtree_delete(&ctx->sh->rbtree, node);
+
+        ngx_slab_free_locked(ctx->shpool, node);
+
```
doujiang24, line 1511: style: this blank line can be removed.

src/ngx_http_lua_shdict.c
```diff
@@ -1305,6 +1305,461 @@ ngx_http_lua_ffi_shdict_udata_to_zone(void *zone_udata)
...
+            if (node != NULL) {
+                goto allocated;
+            }
+        }
+
+        ngx_shmtx_unlock(&ctx->shpool->mutex);
+
+        *errmsg = "no memory";
+        return NGX_ERROR;
+    }
```
doujiang24, line 1554: maybe we can put these lines into a common function?

src/ngx_http_lua_shdict.c
```diff
@@ -1703,6 +2158,145 @@ ngx_http_lua_ffi_shdict_get(ngx_shm_zone_t *zone, u_char *key,
 }
 
 
+int
+ngx_http_lua_ffi_shdict_get_conditonal(ngx_shm_zone_t *zone, u_char *key,
```
doujiang24, line 2162: seems it's a useless function now?

doujiang24:
I'm fine with the `cas` API. We'd better wait for @agentzh 's comment for the `cos` API.

src/ngx_http_lua_shdict.c
Miniwoffer: I could rewirte it to a lot more readable, but some code duplication.
```suggestion
    if (match_flags || match_value) {
        if (rc != NGX_OK) {
            ngx_shmtx_unlock(&ctx->shpool->mutex);
            *match = 1;
            return NGX_DECLINED;
        }

        if (match_flags && old_user_flags != (int) sd->user_flags) {
            ngx_shmtx_unlock(&ctx->shpool->mutex);
            *match = 1;
            return NGX_DECLINED;
        }

        if (match_value) {
            if (old_value_type != sd->value_type
                || old_str_value_len != sd->value_len
                || ngx_memcmp(old_str_value_buf, sd->data + key_len,
                              (size_t) old_str_value_len))
            {
                ngx_shmtx_unlock(&ctx->shpool->mutex);
                *match = 1;
                return NGX_DECLINED;
            }
        }
    }
```

doujiang24: I think this may be better
```c
    if (match_flags
        && (rc != NGX_OK || old_user_flags != (int) sd->user_flags))
    {
        ngx_shmtx_unlock(&ctx->shpool->mutex);
        *match = 1;
        return NGX_DECLINED;
    }

    if (match_value
        && (rc != NGX_OK
            || old_value_type != sd->value_type
            || old_str_value_len != sd->value_len
            || ngx_memcmp(old_str_value_buf, sd->data + key_len,
                           (size_t) old_str_value_len)
        ))
    {
        ngx_shmtx_unlock(&ctx->shpool->mutex);
        *match = 1;
        return NGX_DECLINED;
    }
```

src/ngx_http_lua_shdict.c
doujiang24: yeah, LGTM.

README.markdown
```diff
@@ -6884,6 +6888,26 @@ See also [ngx.shared.DICT](#ngxshareddict).
 
 [Back to TOC](#nginx-api-for-lua)
 
+ngx.shared.DICT.cog
+-------------------
+**syntax:** *value, flags = ngx.shared.DICT:cog(key, old_value?, old_flags?)*
+
+**context:** *set_by_lua&#42;, rewrite_by_lua&#42;, access_by_lua&#42;, content_by_lua&#42;, header_filter_by_lua&#42;, body_filter_by_lua&#42;, log_by_lua&#42;, ngx.timer.&#42;, balancer_by_lua&#42;, ssl_certificate_by_lua&#42;, ssl_session_fetch_by_lua&#42;, ssl_session_store_by_lua&#42;, ssl_client_hello_by_lua&#42;*
+
+Similar to the [get](#ngxshareddictget) method, but only returns if
```
agentzh, line 6897: What does the "only returns if" part mean? If the if condition is not met, this method will never return? It cannot be. Maybe you mean "will only return the value and the flags" here? And you should also be explicit about what values it returns otherwise (like two nil values?)

The `cog` name is a bit confusing. How about `get_if_not_eq`?

README.markdown
```diff
@@ -6884,6 +6888,26 @@ See also [ngx.shared.DICT](#ngxshareddict).
...
+`old_value` or `old_flags` do not match.
+
+If `old_value` or `old_flags` is `nil`
+it will be ignored when comparing.
```
agentzh, line 6901: This semantics does not feel right to me. When old value is nil, then it means the expected value in the shdict should also be nil. It should not get ignored? So it always return `nil, nil`in this case?

agentzh (changes requested):

README.markdown
Miniwoffer: Yeah the "only returns if" part is not true. 
Maybe something like:
```
Similar to the [get](#ngxshareddictget) method, but if ``old_value`` and
``old_flags`` match value and flag in dictionary ngx.shared.DICT, 
returns ``nil, true``.
```
I'm fine with ``get_if_not_eq``

README.markdown
bjne: When old value is nil, it should be ignored when comparing (so value could be anything), when old flag is nil, flag is ignored.
This way the api user could have either old_value or old_flag (or both) and whichever is not nil will be compared to the value in shm

xiaocang:
Hi @Miniwoffer, may I ask if there is a recent update commit in this PR?

Miniwoffer:
> Hi @Miniwoffer, may I ask if there is a recent update commit in this PR?

I was waiting for a comment on my answer to @agentzh s question. For some reason my comments are hidden
Screenshot from 2022-01-21 15-58-55

Il just commit my suggestions

Miniwoffer:
@agentzh 

In ```get_if_not_eq```, ```nil``` for ```old_value``` or ```old_flag``` is equal to any value.
Why?
It adds more functionality.

In the case of ```get_if_not_eq(key, nil, flag)``` if we where to interpret ```nil```
as literal ```nil``` it would simply just be a ```get(key)```.
But if we interpret ```nil``` as any it will act as a ```get_if_flag_not_eq(key, flag)```.

Same goes for ```get_if_not_eq(key, value, nil)``` this would be the same as
```get_if_not_eq(key, value, 0)``` since 0==nil in shm.
But if we interpret ```nil``` as any it would act as a ```get_if_value_not_eq(key, value)```

We could off course add 
```
get_if_value_no_eq(key, value)
get_if_flag_not_eq(key, flag)
```
and have the same logic, if the API is too unintelligible.

I could also try and rewrite the documentation, if wording similar to ```nil==any```
would help.

LoganDark:
I would strongly prefer `compare_swap` over `cas`, and for the other, `compare_get`, `fetch` or something. `get_if_not_eq` is such a horrible sore to type, I really hope a name like that never makes it into the main release.

Miniwoffer:
@LoganDark 
I agree that ```compare_swap``` and ```compare_get``` are cleaner.
The only disadvantage is that it's less self-documenting.

But then again maybe that can be an advantage, forcing the user not to make assumptions about its functionality.

@xiaocang
Also, this pull request has been idle for a while, and I'm unsure of what to do; is there anything I can do to get the process going?

LoganDark:
> it's less self-documenting

The best thing that can be done for documentation is to ensure the README is complete and up-to-date; this allows the creation of typings:

image

Since the README is already incredibly detailed and useful, I feel like `compare_swap` is more than good enough. But keep in mind that's just a suggestion. Even I think `compare_get` is a little weird...

PR #1958, opened 2021-10-30
feature: added new ffi function ngx_http_lua_ffi_ssl_ciphers.

zhuizhuhaomeng:

### DIFF
--- a/src/ngx_http_lua_ssl_certby.c
+++ b/src/ngx_http_lua_ssl_certby.c
@@ -962,6 +962,109 @@ ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
 
 
 int
+ngx_http_lua_ffi_ssl_ciphers(ngx_http_request_t *r, char **pciphers,
+    size_t *cipherslen, char **err)
+{
+    ngx_pool_t          *pool;
+    ngx_ssl_conn_t      *ssl_conn;
+    ngx_connection_t    *c;
+
+    if (r->connection == NULL || r->connection->ssl == NULL) {
+        *err = "bad request";
+        return NGX_ERROR;
+    }
+
+    ssl_conn = r->connection->ssl->connection;
+    if (ssl_conn == NULL) {
+        *err = "bad ssl conn";
+        return NGX_ERROR;
+    }
+
+    pool = r->pool;
+    c = ngx_ssl_get_connection(ssl_conn);
+
+#ifdef SSL_CTRL_GET_RAW_CIPHERLIST
+
+    int                n, i, bytes;
+    size_t             len;
+    u_char            *ciphers, *p;
+    const SSL_CIPHER  *cipher;
+
+    bytes = SSL_get0_raw_cipherlist(c->ssl->connection, NULL);
+    n = SSL_get0_raw_cipherlist(c->ssl->connection, &ciphers);
+
+    if (n <= 0) {
+        *cipherslen = 0;
+        return NGX_OK;
+    }
+
+    len = 0;
+    n /= bytes;
+
+    for (i = 0; i < n; i++) {
+        cipher = SSL_CIPHER_find(c->ssl->connection, ciphers + i * bytes);
+
+        if (cipher) {
+            len += ngx_strlen(SSL_CIPHER_get_name(cipher));
+
+        } else {
+            len += sizeof("0x") - 1 + bytes * (sizeof("00") - 1);
+        }
+
+        len += sizeof(":") - 1;
+    }
+
+    *pciphers = ngx_pnalloc(pool, len);
+    if (*pciphers == NULL) {
+        return NGX_ERROR;
+    }
+
+    p = (u_char *) *pciphers;
+
+    for (i = 0; i < n; i++) {
+        cipher = SSL_CIPHER_find(c->ssl->connection, ciphers + i * bytes);
+
+        if (cipher) {
+            p = ngx_sprintf(p, "%s", SSL_CIPHER_get_name(cipher));
+
+        } else {
+            p = ngx_sprintf(p, "0x");
+            p = ngx_hex_dump(p, ciphers + i * bytes, bytes);
+        }
+
+        *p++ = ':';
+    }
+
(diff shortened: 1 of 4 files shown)


src/ngx_http_lua_ssl_certby.c
```diff
@@ -961,6 +961,109 @@ ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
...
+        } else {
+            len += sizeof("0x") - 1 + bytes * (sizeof("00") - 1);
+        }
+
+        len += sizeof(":") - 1;
+    }
+
+    *pciphers = ngx_pnalloc(pool, len);
+    if (*pciphers == NULL) {
+        return NGX_ERROR;
```
doujiang24, line 1019: should set the `err` here.

src/ngx_http_lua_ssl_certby.c
```diff
@@ -961,6 +961,109 @@ ngx_http_lua_ffi_ssl_raw_client_addr(ngx_http_request_t *r, char **addr,
...
+        == NULL)
+    {
+        *cipherslen = 0;
+        return NGX_OK;
+    }
+
+    *cipherslen = ngx_strlen(buf);
+    *pciphers = ngx_pnalloc(pool, *cipherslen);
+    if (*pciphers == NULL) {
+        return NGX_ERROR;
```
doujiang24, line 1056: ditto should set the `err` here.

doujiang24 (changes requested):

doujiang24 (commented):
two more ideas:
1. we'd better pass a memory buffer pointer from Lua to reduce memory allocation in the connection pool. But the current implementation is also acceptable to me.
2. we can call `ngx_ssl_get_ciphers` in the `ngx_http_lua_ffi_ssl_ciphers` function if we decide to allocation memory from the connection pool to avoid coping code.

zhuizhuhaomeng:
> two more ideas:
> 
> 1. we'd better pass a memory buffer pointer from Lua to reduce memory allocation in the connection pool. But the current implementation is also acceptable to me.
> 2. we can call `ngx_ssl_get_ciphers` in the `ngx_http_lua_ffi_ssl_ciphers` function if we decide to allocation memory from the connection pool to avoid coping code.

I prefer the latter one because it is hard to decide the length of the buffer.

doujiang24 (approved):

bjne:
The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is
perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client
and server.

xiaocang (approved):

zhuizhuhaomeng:
> The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client and server.

Can you be more specific about your application scenario?

bjne:
> > The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client and server.
> 
> Can you be more specific about your application scenario?

A common usecase where is probably to check if client supports EDCSA ciphers, and serve certs accordingly. Splitting/iterating/garbage collecting colon-separated strings here is omho less effective than returning uint32_t[]
of uint32_t SSL_CIPHER_get_protocol_id(const SSL_CIPHER *c);

Also, only the enabled ciphers (SSL_get_ciphers/SSL_get_client_ciphers) is more useful than also needing to know
what ciphers are actually enabled to make decision.

zhuizhuhaomeng:
> > > The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client and server.
> > 
> > 
> > Can you be more specific about your application scenario?
> 
> A common usecase here is probably to check if client supports EDCSA ciphers, and serve certs accordingly. Splitting/iterating/garbage collecting colon-separated strings here is less effective than returning uint32_t[] of uint32_t SSL_CIPHER_get_protocol_id(const SSL_CIPHER *c);
> 
> Also, only the enabled ciphers (SSL_get1_supported_ciphers/SSL_get_client_ciphers) is more useful than also needing to know what ciphers are actually enabled to make decision.

Would you please submit a PR?

bjne:
> > > > The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client and server.
> > > 
> > > 
> > > Can you be more specific about your application scenario?
> > 
> > 
> > A common usecase here is probably to check if client supports EDCSA ciphers, and serve certs accordingly. Splitting/iterating/garbage collecting colon-separated strings here is less effective than returning uint32_t[] of uint32_t SSL_CIPHER_get_protocol_id(const SSL_CIPHER *c);
> > Also, only the enabled ciphers (SSL_get1_supported_ciphers/SSL_get_client_ciphers) is more useful than also needing to know what ciphers are actually enabled to make decision.
> 
> Would you please submit a PR?

Yes, I can do that.

bjne:
> > > > > The output format of ssl_get_chipers is imho ineffective for what it is useful for. What I would like to get out is perhaps a table (table could be passed as input for reuse) or just a int array of ciphers shared between client and server.
> > > > 
> > > > 
> > > > Can you be more specific about your application scenario?
> > > 
> > > 
> > > A common usecase here is probably to check if client supports EDCSA ciphers, and serve certs accordingly. Splitting/iterating/garbage collecting colon-separated strings here is less effective than returning uint32_t[] of uint32_t SSL_CIPHER_get_protocol_id(const SSL_CIPHER *c);
> > > Also, only the enabled ciphers (SSL_get1_supported_ciphers/SSL_get_client_ciphers) is more useful than also needing to know what ciphers are actually enabled to make decision.
> > 
> > 
> > Would you please submit a PR?
> 
> Yes, I can do that.

Here is the RFC pull request: https://github.com/openresty/lua-nginx-module/pull/1962

Issue #1992, opened 2022-01-15
content_by_lua_file should not return 404 if file permission denied


jizhuozhi:
There are three situations in opening a file in the Linux file system
1. Open successfully and execute
2. The file exists but has no permissions
3. The file does not exist

In some cases, developers may set wrong permissions when deploying lua scripts. At this time, content_by_lua_file will directly return 404, which is inappropriate. It stands to reason that it should return 403 or other 5xx series exception codes. , which makes it easier for us to use other anomaly detection systems to identify anomalies.

However, in order to be compatible with the old version, the configuration of the status code in both cases of not found and no permission can be provided, and both are set to 404 by default.

jizhuozhi:
Here is an example code to fix this issue, but could be more graceful?
```c
    if (rc != 0) {
        /*  Oops! error occurred when loading Lua script */
        switch (rc) {
        case LUA_ERRMEM:
            err = "memory allocation error";
            break;

        case LUA_ERRFILE:
            if (errno == ENOENT)
                errcode = NGX_HTTP_NOT_FOUND;
            else if (errno == EACCES)
                errcode = NGX_HTTP_FORBIDDEN;
            else
                errcode = NGX_HTTP_INTERNAL_SERVER_ERROR;
            /* fall through */

        default:
            if (lua_isstring(L, -1)) {
                err = lua_tostring(L, -1);

            } else {
                err = "unknown error";
            }
        }

        goto error;
    }
```

PR #1995, opened 2022-01-18
 feature: content_by_lua_file return 503 if read err and file is existed #1992

jizhuozhi:


zhuizhuhaomeng:
I think 503 is better since content_by_lua_file will execute the Lua script in the file but not return the file itself.

jizhuozhi:
> I think 503 is better since content_by_lua_file will execute the Lua script in the file but not return the file itself.

LGTM, this may be an appropriate semantic.

t/002-content.t
```diff
@@ -850,3 +852,13 @@ GET /lua
 --- error_code: 500
 --- error_log eval
 qr/failed to load inlined Lua code: /
+
```
membphis, line 855: style: three blank lines between different test cases

jizhuozhi:
Maybe we need a CONTRIBUTING GUIDE, there are too many code style problems when submitting code and writing test cases

xiaocang:
> Maybe we need a CONTRIBUTING GUIDE, there are too many code style problems when submitting code and writing test cases

Here's the coding style of c: http://openresty.org/en/c-coding-style-guide.html
This is the tool to check `.t` file: https://github.com/openresty/openresty-devel-utils/blob/master/reindex

t/002-content.t
```diff
@@ -1,5 +1,7 @@
 # vim:set ft= ts=4 sw=4 et fdm=marker:
 
+use strict;
+use warnings FATAL => 'all';
```
xiaocang, line 4: Here is the duplicate code, because the following `use Test::Nginx::Socket::Lua` will introduces the same restriction

xiaocang (changes requested):
LGTM

t/002-content.t
jizhuozhi: Sorry, it's CLion auto completed, I will remove it

zhuizhuhaomeng:
@jizhuozhi  would you please change 503 to 500(Internal Server Error).

the 503 error:
```
The HyperText Transfer Protocol (HTTP) 503 Service Unavailable server error response code indicates that the server is not ready to handle the request.

Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions and the Retry-After HTTP header should, if possible, contain the estimated time for the recovery of the service.
```

jizhuozhi:
> @jizhuozhi would you please change 503 to 500(Internal Server Error).
> 
> the 503 error:
> 
> ```
> The HyperText Transfer Protocol (HTTP) 503 Service Unavailable server error response code indicates that the server is not ready to handle the request.
> 
> Common causes are a server that is down for maintenance or that is overloaded. This response should be used for temporary conditions and the Retry-After HTTP header should, if possible, contain the estimated time for the recovery of the service.
> ```

I would not change the response code from `503` to `500`, cause as a proxy server, we need and must distinguish between proxy server errors and upstream business system errors.

jizhuozhi:
@zhuizhuhaomeng And I think it is a special case of maintenance, cause file permission almost never changed on the fly without system admins. But for some careless admins, they might set wrong permissions

jizhuozhi:
@xiaocang Hello, I'm sorry too long to remember resolving the conflict, PTAL.

zhuizhuhaomeng (approved):

PR #2005, opened 2022-02-08
bugfix: the error message should use the first line rather than the last line of the code block when load lua code block failed.

zhuizhuhaomeng:

### DIFF
--- a/src/ngx_http_lua_directive.c
+++ b/src/ngx_http_lua_directive.c
@@ -1302,6 +1302,9 @@ ngx_http_lua_gen_chunk_name(ngx_conf_t *cf, const char *tag, size_t tag_len,
 {
     u_char      *p, *out;
     size_t       len;
+    ngx_uint_t   start_line;
+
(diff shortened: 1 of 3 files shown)


src/ngx_http_lua_directive.c
```diff
@@ -1328,10 +1331,13 @@ ngx_http_lua_gen_chunk_name(ngx_conf_t *cf, const char *tag, size_t tag_len,
 
 found:
 
+    lmcf = ngx_http_conf_get_module_main_conf(cf, ngx_http_lua_module);
+    start_line = lmcf->directive_line ? : cf->conf_file->line;
```
agentzh, line 1335: We should never use the GNU C language extension. Never. We should also add this check to the ngx-releng tool?

agentzh (changes requested):

agentzh (approved):

xiaocang (approved):

PR #2012, opened 2022-02-15
doc: change desc for ssl_client_hello_by_lua

chronolaw:
There is no 1.19.9.2, should be 1.21.4.1.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -2738,7 +2738,7 @@ This directive requires OpenSSL 1.1.1 or greater.
 
 If you are using the [official pre-built
 packages](https://openresty.org/en/linux-packages.html) for
-[OpenResty](https://openresty.org/) 1.19.9.2 or later, then everything should
+[OpenResty](https://openresty.org/) 1.21.4.1 or later, then everything should
 work out of the box.
 
 If you are not using one of the [OpenSSL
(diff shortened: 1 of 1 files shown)


zhuizhuhaomeng (approved):

zhuizhuhaomeng:
merged with the following patch
```
index 5c79dda3..fa83c6dc 100644
--- a/doc/HttpLuaModule.wiki
+++ b/doc/HttpLuaModule.wiki
@@ -2319,7 +2319,7 @@ This directive requires OpenSSL 1.1.1 or greater.
 
 If you are using the [official pre-built
 packages](https://openresty.org/en/linux-packages.html) for
-[OpenResty](https://openresty.org/) 1.19.9.2 or later, then everything should
+[OpenResty](https://openresty.org/) 1.21.4.1 or later, then everything should
 work out of the box.
 
 If you are not using one of the [OpenSSL
@@ -2330,7 +2330,7 @@ in order to use this directive:
 https://openresty.org/en/openssl-patches.html
 
 Similarly, if you are not using the Nginx core shipped with
-[OpenResty](https://openresty.org) 1.19.9.2 or later, you will need to apply
+[OpenResty](https://openresty.org) 1.21.4.1 or later, you will need to apply
 patches to the standard Nginx core:
 
 https://openresty.org/en/nginx-ssl-patches.html
```

PR #2052, opened 2022-05-25
optimize: destroy pipe proc when freeing the request.

zhuizhuhaomeng:

### DIFF
--- a/src/ngx_http_lua_pipe.c
+++ b/src/ngx_http_lua_pipe.c
@@ -1197,6 +1197,11 @@ ngx_http_lua_pipe_put_error(ngx_http_lua_pipe_ctx_t *pipe_ctx, u_char *errbuf,
 
     case PIPE_ERR_CLOSED:
         *errbuf_size = ngx_snprintf(errbuf, *errbuf_size, "closed") - errbuf;
+        if (pipe_ctx->c != NULL) {
+            ngx_close_connection(pipe_ctx->c);
+            pipe_ctx->c = NULL;
+        }
+
         break;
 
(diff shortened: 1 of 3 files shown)


src/ngx_http_lua_pipe.c
```diff
@@ -1139,6 +1158,10 @@ ngx_http_lua_ffi_pipe_proc_destroy(ngx_http_lua_ffi_pipe_proc_t *proc)
         }
     }
 
+    *pipe->cleanup = NULL;
```
agentzh, line 1161: Should we test if `pipe->cleanup` is already NULL here?

src/ngx_http_lua_pipe.c
```diff
@@ -1174,6 +1197,11 @@ ngx_http_lua_pipe_put_error(ngx_http_lua_pipe_ctx_t *pipe_ctx, u_char *errbuf,
 
     case PIPE_ERR_CLOSED:
         *errbuf_size = ngx_snprintf(errbuf, *errbuf_size, "closed") - errbuf;
+        if (pipe_ctx->c != NULL) {
+            ngx_close_connection(pipe_ctx->c);
+            pipe_ctx->c = NULL;
+        }
```
agentzh, line 1203: I'm worried that this may break the full-duplex feature.

agentzh: Maybe we should make sure both directions are closed before we proactively close and free up the pipe? We can mark this as a TODO for now (and create a github issue to record this?)

agentzh (approved):

PR #2104, opened 2022-09-30
doc: add release description for ngx.worker.pids().

zhuizhuhaomeng:


spacewander (approved):

agentzh:
This PR has several issues that need to be addressed:
1. It fails to build on Windows.
2. It may overrun the input `pids` array. it never checks the buffer size when writing to it.
3. It should never allocate a 4KB buffer even when this API is never used. And we should really reuse our temp buf in resty.core.base to save the buffer altogether.

PR #2136, opened 2022-12-15
bugfix: fix receiveuntil rest bytes count.

rodman10:


rodman10:
When pattern recoverying matched, the rest bytes should decrease by the distance between new and old state instead of just new state.

zhuizhuhaomeng (approved):

zhuizhuhaomeng:
@rodman10  good catch

Issue #2131, opened 2022-12-07
Tcp cosocket receiveuntil call mixed by other reading calls get unexpected result


rodman10:
If call `receiveuntil` followed by `receive` call when the boundary pattern has ambiguity, 
```nginx
location /echo {
    return 200 -----abc--abc;
}

location /socket {
   default_type 'text/plain';
   content_by_lua_block {
       local sock = ngx.socket.tcp()
        
       ...
        
       local reader = sock:receiveuntil("--abc")
        
       local line, err, part = reader(2)
       ngx.say("read 1: ", "[", line, "]")
        
       line, err, part = sock:receive(4)
       ngx.say("read 2: ", "[", line, "]")
        
       line, err, part = reader()
       ngx.say("read 3: ", "[", line, "]")
    }
}
```
,

PR #2137, opened 2022-12-26
bugfix: mixed recv api call causes unexpected result.

rodman10:
Please see https://github.com/openresty/lua-nginx-module/issues/2131 for more details.

Through recovering bytes from compiled pattern, other read calls are able to read these pending bytes instead of the following data, which avoids generating unexpected result. More details could be seen in code and comment.


zhuizhuhaomeng:
@rodman10  please add test cases to covert this issue

rodman10:
> @rodman10 please add test cases to covert this issue
Okay, I have found some futher improvement and change to draft now.

PR #2145, opened 2023-01-13
bugfix: mixed recv api call causes unexpected result.

rodman10:
Please see https://github.com/openresty/lua-nginx-module/issues/2131 for more details.

Through recovering bytes from compiled pattern, other read calls are able to read these pending bytes instead of the following data, which avoids generating unexpected result. More details could be seen in code and comment.

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -75,6 +75,8 @@ static void ngx_http_lua_socket_dummy_handler(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u);
 static int ngx_http_lua_socket_tcp_receive_helper(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u, lua_State *L);
+static void ngx_http_lua_socket_tcp_read_prepare(ngx_http_request_t *r,
+    ngx_http_lua_socket_tcp_upstream_t *u, void *data, lua_State *L);
 static ngx_int_t ngx_http_lua_socket_tcp_read(ngx_http_request_t *r,
     ngx_http_lua_socket_tcp_upstream_t *u);
 static int ngx_http_lua_socket_tcp_receive_retval_handler(ngx_http_request_t *r,
@@ -2163,8 +2165,6 @@ ngx_http_lua_socket_tcp_receive_helper(ngx_http_request_t *r,
     ngx_http_lua_ctx_t                  *ctx;
     ngx_http_lua_co_ctx_t               *coctx;
 
-    u->input_filter_ctx = u;
-
     ctx = ngx_http_get_module_ctx(r, ngx_http_lua_module);
 
     if (u->bufs_in == NULL) {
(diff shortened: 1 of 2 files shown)


zhuizhuhaomeng (dismissed):

src/ngx_http_lua_socket_tcp.c
```diff
@@ -2514,6 +2516,87 @@ ngx_http_lua_socket_read_any(void *data, ssize_t bytes)
...
+
+    ngx_memcpy(b, new_cl->buf, sizeof(ngx_buf_t));
+
+    b->last = ngx_copy(b->last, cp->pattern.data, cp->state);
+    b->last = ngx_copy(b->last, u->buf_in->buf->pos, size);
+
+    u->buf_in->next = ctx->free_recv_bufs;
+    ctx->free_recv_bufs = u->buf_in;
+
+    u->bufs_in = new_cl;
```
zhuizhuhaomeng, line 2592: @rodman10  The old u->bufs_in maybe has multiple ngx_chain_t in it.
Should we deal with this situation?

rodman10: In `ngx_http_lua_socket_push_input_data` function, when `nbufs` more than one, it will always point to the last one?

rodman10: In `ngx_http_lua_socket_push_input_data` function, when `nbufs` more than one, it will always point to the last one?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4915,13 +5003,28 @@ ngx_http_lua_socket_cleanup_compiled_pattern(lua_State *L)
...
 
     cp = lua_touserdata(L, 1);
-    if (cp == NULL || cp->recovering == NULL) {
+    if (cp == NULL) {
+        return 0;
+    }
+
+    u = cp->upstream;
+    if (u != NULL) {
+        /* ensure request not released */
```
doujiang24, line 5019: what does this comment mean? could you please make it more clear?

`u->cleanup != NULL` means request is not released?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -4915,13 +5003,28 @@ ngx_http_lua_socket_cleanup_compiled_pattern(lua_State *L)
...
+        if (u->cleanup != NULL) {
+            ngx_http_lua_socket_tcp_read_prepare(u->request, u, NULL, L);
+        }
+
+        u->input_filter_ctx = NULL;
```
doujiang24, line 5024: better put this line in `else` branch?

src/ngx_http_lua_socket_tcp.c
rodman10: If `socket` obj is freed before `pattern` obj, `ngx_http_lua_socket_tcp_finalize` will be called and set `cleanup` field to `NULL`.
I will try to make the comment more clear.

rodman10: It seems the inner `if` statement is meaningless, since `request` or `socket` has been freed before pattern and function `ngx_http_lua_socket_tcp_finalize` has set field `upstream` to [NULL](https://github.com/openresty/lua-nginx-module/pull/2145/files#diff-ebaa829579f91b18746b0bb13948131e7dcaa19dbae1d843c6fbbebc682951a2R4331)

doujiang24 (approved):

PR #2149, opened 2023-01-28
bugfix: Fix remove bad white space from headers

ram2012k:
Please Check Issue #2140
removing leading and trailing bad white-space from key of ngx.req.set_header(key, value) and ngx.header["key "] = 'value' as they are causing failure in the frameworks that are using openresty .
#2140


zhuizhuhaomeng:
Doing this on every request incurs an unnecessary performance penalty. This should be done at the control surface.

PR #2182, opened 2023-04-22
Set headers in the tail of a list rather than in the head

oxpa:
It seems that ngx_http_variable_headers_internal function in nginx/src/http/ngx_http_variables.c relies on the fact that the last header values is stored in the last element of a list. Therefore we should avoid changing a random element in the head of a list of headers and either change the last element or add a new element to the list.

Not sure if this is the proper fix.
Thanks to @ssdr for pointing out the problem in #2177 

### DIFF
--- a/src/ngx_http_lua_headers_out.c
+++ b/src/ngx_http_lua_headers_out.c
@@ -319,61 +319,39 @@ ngx_http_set_builtin_multi_header(ngx_http_request_t *r,
 
     headers = (ngx_table_elt_t **) ((char *) &r->headers_out + hv->offset);
 
-    if (hv->no_override) {
-        for (h = *headers; h; h = h->next) {
-            if (!h->hash) {
-                h->value = *value;
-                h->hash = hv->hash;
-                return NGX_OK;
-            }
-        }
-
-        goto create;
-    }
-
-    /* override old values (if any) */
-
-    if (*headers) {
-        for (h = (*headers)->next; h; h = h->next) {
+    for (h = *headers; h; h = h->next) {
+        if (!hv->no_override) {
             h->hash = 0;
             h->value.len = 0;
         }
+    }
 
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_headers_out.c
```diff
@@ -319,62 +319,40 @@ ngx_http_set_builtin_multi_header(ngx_http_request_t *r,
...
+    for (h = *headers; h; h = h->next) {
+        if (!hv->no_override) {
             h->hash = 0;
             h->value.len = 0;
         }
+    }
 
-        h = *headers;
-
+    if (h->hash == 0) {
```
ssdr, line 329: Oh no, I'm afraid here `h` is NULL ptr after last for loop iterate...

src/ngx_http_lua_headers_out.c
```diff
@@ -319,62 +319,40 @@ ngx_http_set_builtin_multi_header(ngx_http_request_t *r,
...
         }
 
-        return NGX_OK;
-    }
-
-create:
-
-    for (ph = headers; *ph; ph = &(*ph)->next) { /* void */ }
+    } else {
+        ho = ngx_list_push(&r->headers_out.headers);
```
ssdr, line 337: when last element is valid and other elements are not, here will make a new element, right? Does it mean that the invalid elements never be reused in pool? So why not just nullifying the first element `next` for simplicity?

ssdr:
Actually when set builtin multi `input` header, it just nullifying the next pointer.

can we do same thing for output headers?

src/ngx_http_lua_headers_out.c
ssdr: Actually when set builtin multi input header, it just nullifying the next pointer.

can we do same thing for output headers?

PR #2190, opened 2023-05-12
bugfix: when value type is SHDICT_TNUMBER, then it should get union n…

willmafh:
…umber field instead of bool field

### DIFF
--- a/src/ngx_http_lua_shdict.c
+++ b/src/ngx_http_lua_shdict.c
@@ -654,7 +654,7 @@ ngx_http_lua_shared_dict_get(ngx_shm_zone_t *zone, u_char *key_data,
             return NGX_ERROR;
         }
 
-        ngx_memcpy(&value->value.b, data, len);
+        ngx_memcpy(&value->value.n, data, len);
         break;
 
     case SHDICT_TBOOLEAN:


spacewander (approved):

src/ngx_http_lua_shdict.c
```diff
@@ -654,7 +654,7 @@ ngx_http_lua_shared_dict_get(ngx_shm_zone_t *zone, u_char *key_data,
             return NGX_ERROR;
         }
 
-        ngx_memcpy(&value->value.b, data, len);
+        ngx_memcpy(&value->value.n, data, len);
```
zhuizhuhaomeng, line 657:
```suggestion
        value->value.n = *(double *) data;
```

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_shdict.c
willmafh: I believe it's better to use `ngx_memcpy(&value->value.n, data, len);`, since in pop helper and many other places, all SHDICT_TNUMBER type value are stored by ngx_memcpy. What do you think?

zhuizhuhaomeng: This is not a big deal. This function is only used by the test cases.

zhuizhuhaomeng (approved):

PR #2192, opened 2023-05-30
doc: according to context of lua_socket_pool_size directive, `location if` should be added

willmafh:
* context of lua_socket_pool_size directive
location_if

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -3240,7 +3240,7 @@ lua_socket_pool_size
 
 **default:** *lua_socket_pool_size 30*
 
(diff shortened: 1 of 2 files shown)


cppcoffee (approved):

zhuizhuhaomeng:
@willmafh Most of the directive of lua-nginx-module does not support in the if location.
I don't think it is a good idea to add support for this directive.

zhuizhuhaomeng:
I will close this PR.  @willmafh 
If there is a strong demand, we reopen this PR again

PR #2236, opened 2023-10-13
bugfix: don't include pcre.h with PCRE2 used

Ansuel:
pcre.h is a PCRE header and is not exposed by PCRE2 library causing compilation error as the header is not found.

Don't include pcre.h if nginx is compiled with PCRE2 support enabled.

Fixes: cb83e33e2657 ("feature: support pcre2")

### DIFF
--- a/src/ngx_http_lua_common.h
+++ b/src/ngx_http_lua_common.h
@@ -54,7 +54,7 @@ typedef struct {
 #endif
 
 
-#if (NGX_PCRE)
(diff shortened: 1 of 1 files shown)


swananan (approved):
LGTM, thanks for this bugfix, could you raise another pr for stream-lua-nginx-module

Ansuel:
@swananan doing it. Also I force pushed as I added !defined to make this more clear.

zhuizhuhaomeng (approved):

Issue #2211, opened 2023-07-29
HTTP/2 request processing optionally


splitice:
PR https://github.com/openresty/lua-nginx-module/pull/2174 removes support for http request processing.

I think this should be optionally still supported, it is after all perfectly servisable in a HTTP/2 request response flow. Perhaps a timeout can be added to error on grpc (rather than spiking the request)?

oowl:
The problem is not only GRPC requests, All HTTP2 traffic may  have long connection stream, you can see my analyze in https://github.com/openresty/lua-nginx-module/issues/2172#issuecomment-1486378063

oowl:
HTTP2 is stream base protocol, read the whole body is theoretically impossible, and it does not make sense.

splitice:
The way nginx does it, potentially ignoring Content-Length seems a bit strange. However most request / responses do follow the non infinite upload property...

I think a timeout would be a good idea (maximum time spent reading request data) for sure.

oowl:
> I think a timeout would be a good idea (maximum time spent reading request data) for sure.
I agree 😆

zhuizhuhaomeng:
@oowl  I think we need to revert the previous PR and disable in GRPC only.  cc @splitice 


> HTTP2 is stream base protocol, read the whole body is theoretically impossible, and it does not make sense.

splitice:
It will block the same amount as http/1.1.That should be acceptable and consistent in behaviour.

oowl:
> Will a large HTTP request body in http2 connection block other HTTP requestes in the same connection? @oowl

No, Nginx process HTTP2 one `ngx_http_request_t` per stream, so every stream is isolated, one stream traffic spike can not influence other HTTP2 stream traffic. 

And  I have chat with @zhuizhuhaomeng on IM about https://github.com/openresty/lua-nginx-module/pull/2174 , It can not be revert, Openresty  can not support HTTP2 read  body now, So maybe we need to consider this is a feature(`read _body` in HTTP2)， And I think `read_body` timeout is a good idea in HTTP2 processing traffic. @zhuizhuhaomeng @splitice

swananan:
[...]

PR #2174, opened 2023-03-29
bugfix: disable http2 in body read due to http2 stream processing bug

oowl:
I hereby granted the copyright of the changes in this pull request
to the authors of this lua-nginx-module project.

PR #2237, opened 2023-10-16
changes: modify ngx.read_body api limitation in HTTP/2 and HTTP/3

swananan:

#2174 mentioned an issue where there is a chance of blocking the ngx.read_body API when using HTTP/2. As a result, #2174 disabled the _ngx.read_body_ API in HTTP/2. However, #2211 also suggested that perhaps we should not completely disable the _ngx.read_body_ API in HTTP/2, as many HTTP/2 requests could still benefit from this.

To address this, I attempted to define the minimum limitations for using ngx.read_body in HTTP/2. From what I understand, ngx.read_body is based on the Nginx core to handle the client's body, and the Nginx core typically reads the entire client body before calling the callback registered by the lua-nginx-module. There are two scenarios in which HTTP/2 can trigger this callback. First, when the Content-Length header is used, Nginx core can determine the exact number of bytes to read from the client side. The second scenario is when an HTTP/2 request sends the entire body and immediately closes the HTTP/2 stream. Additionally, HTTP/2 does not support the chunked transfer encoding.

The first scenario is relatively straightforward to detect in the lua-nginx-module. However, the second scenario is challenging to identify because Nginx core takes over the remaining part, and the lua-nginx-module can only wait to be called by the Nginx core. Therefore, in this pull request, I propose to only allow HTTP/2 requests that carry the Content-Length header. So in the second scenario, If someone want to use _ngx.read_body_, need to add the Content-Length header.

I also considered adding a timeout option for the read_body API. However, I decided against it for two reasons:

Nginx already has a timeout option for reading the client's body, which is configured using _client_body_timeout_. This configuration terminates the HTTP request when a timeout occurs. Introducing another timeout option, which would keep the rest of the Lua thread logic running, seemed redundant.

Secondly, I believe that scenarios causing the blocking of the _ngx.read_body_ API should not use _ngx.read_body_. Adding a timeout option may not immediately expose the real problem.

It's worth noting that HTTP/3 faces a similar problem, and this pr implemented the same solution for it.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -2722,7 +2722,7 @@ lua_need_request_body
 
 **phase:** *depends on usage*
 
-Due to the stream processing feature of HTTP2, it does not support HTTP2 connection.
+Due to the stream processing feature of HTTP/2 or HTTP/3, this configuration could potentially block the entire request. Therefore, this configuration is effective only when HTTP/2 or HTTP/3 requests send content-length header. For requests with versions lower than HTTP/2, this configuration can still be used without any problems.
 
 Determines whether to force the request body data to be read before running rewrite/access/content_by_lua* or not. The Nginx core does not read the client request body by default and if request body data is required, then this directive should be turned `on` or the [ngx.req.read_body](#ngxreqread_body) function should be called within the Lua code.
 
@@ -5426,7 +5426,7 @@ Reads the client request body synchronously without blocking the Nginx event loo
  local args = ngx.req.get_post_args()
 ```
 
-Due to the stream processing feature of HTTP2, it does not support HTTP2 connection.
+Due to the stream processing feature of HTTP/2 or HTTP/3, this api could potentially block the entire request. Therefore, this api is effective only when HTTP/2 or HTTP/3 requests send content-length header. For requests with versions lower than HTTP/2, this api can still be used without any problems.
 
 If the request body is already read previously by turning on [lua_need_request_body](#lua_need_request_body) or by using other modules, then this function does not run and returns immediately.
 
--- a/src/ngx_http_lua_accessby.c
+++ b/src/ngx_http_lua_accessby.c
@@ -136,11 +136,26 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
         return NGX_DONE;
     }
 
-/* http2 read body may break http2 stream process */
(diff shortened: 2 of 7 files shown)


src/ngx_http_lua_accessby.c
```diff
@@ -136,11 +136,26 @@ ngx_http_lua_access_handler(ngx_http_request_t *r)
...
+                          "http2 read_body may break http2 stream process");
+            goto done;
+        }
+#endif
+
+#if (NGX_HTTP_V3)
+        if (r->http_version == NGX_HTTP_VERSION_30
+            && r->headers_in.content_length_n < 0)
+        {
+            ngx_log_error(NGX_LOG_WARN, r->connection->log, 0,
```
Appla, line 154: http3?

README.markdown
```diff
@@ -2724,7 +2724,7 @@ lua_need_request_body
 
 **phase:** *depends on usage*
 
-Due to the stream processing feature of HTTP2, it does not support HTTP2 connection.
+Due to the stream processing feature of HTTP/2 or HTTP/3, this configuration could potentially block the entire request. Therefore, this configuration is effective only when HTTP/2 or HTTP/3 requests send content-length header. For requests with versions lower than HTTP/2, this configuration can still be used without any problems.
```
oowl, line 2727: We have discussion about this problem and reached an agreement that enables `read_body` API in the HTTP2 request to contain content-length header 
And why `chunk-transfer` not enabled?  from RFC
```
In https://greenbytes.de/tech/webdav/rfc7540.html#rfc.section.8.1.p.4:

    HTTP/2 uses DATA frames to carry message payloads. The chunked transfer encoding defined in Section 4.1 of [RFC7230] MUST NOT be used in HTTP/2.
```

oowl (approved):

src/ngx_http_lua_accessby.c
swananan: Yes, HTTP/3 has same problem.

README.markdown
oowl: @zhuizhuhaomeng Hi, Can you help us take a look at this PR?

zhuizhuhaomeng (approved):

PR #2241, opened 2023-10-18
bugfix: remove extra spaces at the end of the request field name

### LINKED ISSUE #2140: ngx.req.set_header(key, value ) in kong plugin should throw error for BWS ("bad" whitespace) in header key
### Current Behavior

Adding header `ngx.req.set_header("X-cnn-Service ", "")` to a plugin is causing the request to throw HTTP 400 error due to the additional trailing whitespace in header's Key. 
reference- [field-parsing](https://greenbytes.de/tech/webdav/rfc7230.html#field.parsing) according to rfc7230 **No whitespace is allowed between the header field-name and colon** . 
Due to a small leading/trailing whitespace mistake , much of the developers time is wasted to find this issue .

### Expected Behavior

Two ideas can be proposed
1) Trim the leading/trailing whitespace of the header before plugin execution.
2) Log the  error , that plugin execution failed for the request due to invalid header

### Steps To Reproduce

```markdown
Create a Plugin with the following in access.lua
`
local BasePlugin = require "kong.plugins.base_plugin"
local priorities = require "kong.plugins.priorities"

local plugin_name = "headers-plugin"

local KongHeadersHandler = BasePlugin:extend()

function KongHeadersHandler:new()
   KongHeadersHandler.super.new(self, plugin_name)
end

function KongHeadersHandler:access(conf)
  KongHeadersHandler.super.access(self)
  ngx.req.set_header("X-cnn-Service ", "")
end

KongHeadersHandler.PRIORITY = priorities[plugin_name]
KongHeadersHandler.VERSION = "0.0.1"

return KongHeadersHandler`
```

### Anything else?

kong version 1.1.2
lua version  Lua 5.1.4 
LuaJIT 2.1.0-beta3
nginx version: openresty/1.13.6.2
LuaRocks main command-line interface /opt/kong/luajit/bin/luarocks 3.2.1

lynch1981:
According to RFC7230 second paragraph of section-[3.2.4] (https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4).  
No whitespace is allowed between the header field-name and colon.  

I just remove the extra space at the end of the request  field-name for user convenience.

Nginx won't complain about headers like this, so the response part does not need to be processed.

fix  #2140 Thanks @ram2012k.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,9 +62,9 @@ env:
     - MALLOC_PERTURB_=9
   jobs:
     #- NGINX_VERSION=1.21.4 OPENSSL_VER=1.1.0l OPENSSL_PATCH_VER=1.1.0d
-    #- NGINX_VERSION=1.21.4 OPENSSL_VER=1.1.1s OPENSSL_PATCH_VER=1.1.1f
+    - NGINX_VERSION=1.21.4 OPENSSL_VER=1.1.1s OPENSSL_PATCH_VER=1.1.1f
     #- NGINX_VERSION=1.25.1 OPENSSL_VER=1.1.0l OPENSSL_PATCH_VER=1.1.0d
-    #- NGINX_VERSION=1.25.1 OPENSSL_VER=1.1.1s OPENSSL_PATCH_VER=1.1.1f USE_PCRE2=Y
+    - NGINX_VERSION=1.25.1 OPENSSL_VER=1.1.1s OPENSSL_PATCH_VER=1.1.1f USE_PCRE2=Y
     - NGINX_VERSION=1.25.1 BORINGSSL=1 TEST_NGINX_USE_HTTP3=1 USE_PCRE2=Y
     #- NGINX_VERSION=1.25.1 OPENSSL_VER=1.1.1s TEST_NGINX_USE_HTTP2=1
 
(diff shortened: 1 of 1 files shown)


zhuizhuhaomeng (approved):

ram2012k:
@lynch1981 i already created pull request for it long ago https://github.com/openresty/lua-nginx-module/pull/2149

lynch1981:
> @lynch1981 i already created pull request for it long ago #2149
Yes, i saw it. I found a simpler solution. can you help me review the codes? thanks so much

src/ngx_http_lua_headers.c
```diff
@@ -1002,6 +1010,14 @@ ngx_http_lua_ffi_req_set_header(ngx_http_request_t *r, const u_char *key,
         return NGX_DECLINED;
     }
 
+    for (i = key_len; i > 0; i--) {
```
ram2012k, line 1013: Both leading and trailing spaces must be removed

lynch1981: I read the RFC document and did not find the mandatory requirement that no spaces be allowed before field-name

ram2012k: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4

lynch1981: Sorry, did you mean obs-fold?

ram2012k: **A field value might be preceded and/or followed by optional
   whitespace (OWS); a single SP preceding the field-value is preferred
   for consistent readability by humans.  The field value does not
   include any leading or trailing whitespace:**

lynch1981: sorry, we are talking about field name not field value

ram2012k: Hmm , approving it . problem was leading leading whitespace in field name was also failing in kong

ram2012k (approved):

PR #2254, opened 2023-11-11
doc: module function instead of require

decadence:

Isn't `module` an old form of requiring modules?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -731,7 +731,7 @@ instead of the old deprecated form:
 
 ```lua
 
- require('xxx')
+ module('xxx')
 ```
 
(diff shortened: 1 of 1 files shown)


lynch1981:
You can add doc: prefix to the title to fix Lint error ;)

zhuizhuhaomeng:
@decadence  where did you get the old style?
From the stackoverflow, the original doc is right.
https://stackoverflow.com/questions/7248877/in-lua-how-do-you-import-modules

decadence:
@zhuizhuhaomeng it is listed in the text after that:

> The Lua module is actually loaded by the first Nginx request handler and is cached by the require() built-in in the package.loaded table for later reference, and the **module()** builtin used by some Lua modules has the side effect of setting a global variable to the loaded module table

Yes, I think I got it mixed up. `module` is used inside the module file itself, and not when module is required.

PR #2273, opened 2023-12-21
change: don't move the item with expire time to the head when lookup.

zhuizhuhaomeng:


xiaocang:
Is there a problem with this strategy: if there are some keys that have been expired for a long time, will they cause a pileup at the end of the lru queue that can't be cleaned up?
It's better to change the policy to only target expired keys and not update them to the lru queue head.

xiaocang (approved):

PR #2278, opened 2024-01-17
feature: Add ngx_http_lua_ffi_parse_der_cert and ngx_http_lua_ffi_parse_der_key functions

devicenull:
Per some [discussion](https://groups.google.com/g/openresty-en/c/VEXG8w0sbp0) on the mailing list, parsing SSL certs once and caching the result can significantly improve performance over having to parse them each time.  There are existing functions to do this for PEM, but our existing systems were passing around DER files instead.

This just adds the same parsing functions for DER certs.  Patch for the Lua side of this is in the works.

From our testing, caching already parsed certs let us handle 45% more SSL sessions, and decreased 99% latency by about 20%.  These are best-case numbers, and are likely only representative of traffic while under attack.


devicenull:
Hm, those build failures appear to be entirely unrelated to my changes

t/140-ssl-c-api.t
```diff
@@ -1323,3 +1329,152 @@ SNI is test.com
...
+            local rc = ffi.C.ngx_http_lua_ffi_set_priv_key(r, pkey, errmsg)
+            if rc ~= 0 then
+                ngx.log(ngx.ERR, "failed to set cdata priv key: ",
+                        ffi.string(errmsg[0]))
+                return
+            end
+
+            ffi.C.ngx_http_lua_ffi_free_priv_key(pkey)
+        }
+        }
```
zhuizhuhaomeng, line 1395: Is the second bracket redundant?

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1209,6 +1279,35 @@ ngx_http_lua_ffi_parse_pem_priv_key(const u_char *pem, size_t pem_len,
...
+ngx_http_lua_ffi_parse_der_priv_key(const char *data, size_t len,
+    char **err)
+{
+    BIO               *bio = NULL;
+    EVP_PKEY          *pkey = NULL;
+
+    bio = BIO_new_mem_buf((char *) data, len);
+    if (bio == NULL) {
+        *err = "BIO_new_mem_buf() failed";
+        BIO_free(bio);
```
zhuizhuhaomeng, line 1292: bio is NULL, we don't need to call the BIO_free.

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1172,6 +1172,76 @@ ngx_http_lua_ffi_parse_pem_cert(const u_char *pem, size_t pem_len,
...
+void *
+ngx_http_lua_ffi_parse_der_cert(const char *data, size_t len,
+    char **err)
+{
+    BIO             *bio;
+    X509            *x509;
+    u_long           n;
+    STACK_OF(X509)  *chain;
+
+    bio = BIO_new_mem_buf((char *) data, len);
```
zhuizhuhaomeng, line 1184: need to test if the data is NULL

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1209,6 +1279,35 @@ ngx_http_lua_ffi_parse_pem_priv_key(const u_char *pem, size_t pem_len,
...
 
 
+void *
+ngx_http_lua_ffi_parse_der_priv_key(const char *data, size_t len,
+    char **err)
+{
+    BIO               *bio = NULL;
+    EVP_PKEY          *pkey = NULL;
+
+    bio = BIO_new_mem_buf((char *) data, len);
```
zhuizhuhaomeng, line 1289: test if the data is null

zhuizhuhaomeng:
```
index 6ed8f69d..d775f2b3 100644
@@ -1178,13 +1178,12 @@ ngx_http_lua_ffi_parse_der_cert(const char *data, size_t len,
 {
     BIO             *bio;
     X509            *x509;
-    u_long           n;
     STACK_OF(X509)  *chain;
 
     if (data == NULL || len == 0) {
-        *err = "NULL data passed in";
+        *err = "invalid argument";
         ERR_clear_error();
-        return NULL
+        return NULL;
     }
 
     bio = BIO_new_mem_buf((char *) data, len);
@@ -1292,9 +1291,8 @@ ngx_http_lua_ffi_parse_der_priv_key(const char *data, size_t len,
     BIO               *bio = NULL;
     EVP_PKEY          *pkey = NULL;
 
-    if (data == NULL || len == 0)
-    {
-        *err = "NULL data passed in";
+    if (data == NULL || len == 0) {
+        *err = "invalid argument";
         ERR_clear_error();
         return NULL;
     }
diff --git a/t/140-ssl-c-api.t b/t/140-ssl-c-api.t
index 19410043..6b5ff2f8 100644
--- a/t/140-ssl-c-api.t
+++ b/t/140-ssl-c-api.t
@@ -54,10 +54,10 @@ ffi.cdef[[
         size_t pem_len, char **err);
 
     void *ngx_http_lua_ffi_parse_der_cert(const char *data, size_t len,
-        char **err)
+        char **err);
 
     void *ngx_http_lua_ffi_parse_der_priv_key(const char *data, size_t len,
-        char **err)
+        char **err);
 
     int ngx_http_lua_ffi_set_cert(void *r,
         void *cdata, char **err);
@@ -1330,6 +1330,8 @@ SNI is test.com
 [error]
 [alert]
 
+
+
 === TEST 11: DER cert + private key cdata
 --- http_config
     server {
@@ -1467,7 +1469,7 @@ received: Server: nginx
 received: Content-Type: text/plain
 received: Content-Length: 4
 received: Connection: close
-received:
+received: 
 received: foo
 close: 1 nil

```

zhuizhuhaomeng:
I have merged with the above patch.

PR #2302, opened 2024-04-08
bugfix: the connection won't be closed normally when set arg[1] = "" before arg[2] = true

### LINKED ISSUE #2301: bug: the connection won't be closed normally when set arg[1] = ""
When i set `arg[1] = ""` followed by `arg[2] = true`, the connection won't close normally.

image

Here is the example
```
=== TEST 1: set resp body nil with ngx.arg[1] first
--- config
    location /t {
        content_by_lua_block {
            ngx.say("Hello World!")
        }

        body_filter_by_lua_block {
            ngx.arg[1] = ""
            ngx.arg[2] = true
        }
    }
--- request
GET /t
--- response_body
--- no_error_log
[error]
```
But if i set `arg[1] = ""` before `arg[2] = true`, it will work.

image

Here is the example
```
=== TEST 2: set resp body nil with ngx.arg[2] first
--- config
    location /t {
        content_by_lua_block {
            ngx.say("Hello World!")
        }

        body_filter_by_lua_block {
            ngx.arg[2] = true
            ngx.arg[1] = ""
        }
    }
--- request
GET /t
--- response_body
--- no_error_log
[error]
```

This problem still exists in the latest version

monkeyDluffy6017:
fix #2301

The reason is that when we set  `arg[1] = ""` first, the `lmcf->body_filter_chain` will be null
image
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_bodyfilterby.c#L633-L646

image
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_bodyfilterby.c#L668-L700

Then we can't set the right end flag
image
https://github.com/openresty/lua-nginx-module/blob/master/src/ngx_http_lua_bodyfilterby.c#L526-L549

### DIFF
--- a/src/ngx_http_lua_bodyfilterby.c
+++ b/src/ngx_http_lua_bodyfilterby.c
@@ -546,7 +546,8 @@ ngx_http_lua_body_filter_param_set(lua_State *L, ngx_http_request_t *r,
                 lmcf->body_filter_chain = in;
             }
 
-            /* we set the "last_buf" or "last_in_chain" flag in the last buf of "in" */
+            /* we set the "last_buf" or "last_in_chain" flag
+             * in the last buf of "in" */
             for (cl = in; cl; cl = cl->next) {
                 if (cl->next == NULL) {
                     if (r == r->main) {
--- a/src/ngx_http_lua_bodyfilterby.c
+++ b/src/ngx_http_lua_bodyfilterby.c
@@ -536,8 +536,8 @@ ngx_http_lua_body_filter_param_set(lua_State *L, ngx_http_request_t *r,
              * before arg[2] = true*/
             if (in == NULL) {
                 in = ngx_http_lua_chain_get_free_buf(r->connection->log,
-                                                    r->pool,
-                                                    &ctx->free_bufs, 0);
+                                                     r->pool,
+                                                     &ctx->free_bufs, 0);
                 if (in == NULL) {
                     return luaL_error(L, "no memory");
                 }
--- a/src/ngx_http_lua_bodyfilterby.c
+++ b/src/ngx_http_lua_bodyfilterby.c
@@ -532,9 +532,20 @@ ngx_http_lua_body_filter_param_set(lua_State *L, ngx_http_request_t *r,
         if (last) {
             ctx->seen_last_in_filter = 1;
 
-            /* the "in" chain cannot be NULL and we set the "last_buf" or
-             * "last_in_chain" flag in the last buf of "in" */
+            /* the "in" chain cannot be NULL except that we set arg[1] = "" before arg[2] = true*/
+            if (in == NULL) {
+                in = ngx_http_lua_chain_get_free_buf(r->connection->log,
+                                                    r->pool,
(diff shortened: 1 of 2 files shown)


monkeyDluffy6017:
@zhuizhuhaomeng @spacewander please help to review

t/082-body-filter-2.t
```diff
@@ -225,3 +225,43 @@ GET /t
...
+        body_filter_by_lua_block {
+            ngx.arg[1] = ""
+            ngx.arg[2] = true
+        }
+    }
+--- request
+GET /t
+--- response_body
+--- no_error_log
+[error]
```
zhuizhuhaomeng, line 247: [alert]
[crit]

t/082-body-filter-2.t
```diff
@@ -225,3 +225,43 @@ GET /t
...
+        body_filter_by_lua_block {
+            ngx.arg[2] = true
+            ngx.arg[1] = ""
+        }
+    }
+--- request
+GET /t
+--- response_body
+--- no_error_log
+[error]
```
zhuizhuhaomeng, line 267: [alert]
[crit]

zhuizhuhaomeng (changes requested):

t/082-body-filter-2.t
monkeyDluffy6017: Done

t/082-body-filter-2.t
monkeyDluffy6017: Done

t/082-body-filter-2.t
```diff
@@ -34,7 +34,7 @@ log_level('debug');
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 5);
+plan tests => repeat_each() * (blocks() * 5 - 4);
```
spacewander, line 37: Better to use a positive number, so the plan number won't be negative when `--- ONLY` is given.

monkeyDluffy6017: Done

src/ngx_http_lua_bodyfilterby.c
```diff
@@ -532,9 +532,22 @@ ngx_http_lua_body_filter_param_set(lua_State *L, ngx_http_request_t *r,
         if (last) {
             ctx->seen_last_in_filter = 1;
 
-            /* the "in" chain cannot be NULL and we set the "last_buf" or
-             * "last_in_chain" flag in the last buf of "in" */
+            /* the "in" chain cannot be NULL except that we set arg[1] = ""
+             * before arg[2] = true*/
```
zhuizhuhaomeng, line 536:
```suggestion
             * before arg[2] = true
             */
```

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_bodyfilterby.c
monkeyDluffy6017: Done

zhuizhuhaomeng (approved):

PR #2319, opened 2024-06-04
bugfix: fix config test for signalfd with gcc 11

setnicka:
Without the initialization, the feature test ends with:

```
error: 'set' may be used uninitialized [-Werror=maybe-uninitialized]
```

It is because of the [-Wmaybe-uninitialized](https://gcc.gnu.org/onlinedocs/gcc-11.1.0/gcc/Warning-Options.html#index-Wmaybe-uninitialized) enhancements in the gcc 11.

---


spacewander (approved):

spacewander:
The failure of CI isn't caused by this change.

PR #2329, opened 2024-07-04
feature: add ssl trusted certificate

theweakgod:
This mainly adds the ssl_trusted_certificate feature to avoid issues where CA authentication issued by secondary certificates cannot succeed in mTLS. Added the trusted_cert parameter option for verify_client.
I have made the corresponding modifications at the lua-resty-core:
https://github.com/openresty/lua-resty-core/pull/473


zhuizhuhaomeng:
Will you show us a detailed example of the mTLS failed

theweakgod:
> Will you show us a detailed example of the mTLS failed

The scenario is as follows: verify_client will send the subject names of all CA. However, we do not want to expose the root subject name, so verify_client only wants to send the subject names of the intermediate CA to the client. So we only set intermediate CA in verify_client, but this will cause the verification to fail because the root CA is not trusted.

We have tried configuring nginx ssl_trusted_certificate and simultaneously calling verify_client to send the intermediate CA, but found it doesn't work. After reading the source code, I discovered that verify_client overrides the cert->verify_store configured in nginx, and this operation is implicit in verify_client. So, I created this pull request.

This scenario can be resolved by configuring nginx as follows:

ssl_client_certificate client-server.crt;
ssl_trusted_certificate client-server-root.crt;
ssl_verify_depth 2;
ssl_verify_client on;

More details can be found in Test 13(140-ssl-c-api.t).

zhuizhuhaomeng (approved):

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1468,8 +1468,8 @@ ngx_http_lua_ssl_verify_callback(int ok, X509_STORE_CTX *x509_store)
 
 
 int
-ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
-    int depth, char **err)
+ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
```
zhuizhuhaomeng, line 1471: please remove the trailing space

src/ngx_http_lua_ssl_certby.c
```diff
@@ -1535,54 +1536,76 @@ ngx_http_lua_ffi_ssl_verify_client(ngx_http_request_t *r, void *ca_certs,
...
+            /* clean subject name list, and set it for send to client */
+            SSL_set_client_CA_list(ssl_conn, name_chain);
+        }
 
-            /* add to trusted CA store */
-            if (X509_STORE_add_cert(ca_store, x509) == 0) {
-                *err = "X509_STORE_add_cert() failed";
-                goto failed;
+        if (trusted_chain != NULL) {
+
```
zhuizhuhaomeng, line 1588: remove the extra blank line

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_ssl_certby.c
theweakgod: done

zhuizhuhaomeng (approved):

PR #2345, opened 2024-08-01
bugfix: compatible-lower-ssl-version

Masterlvng:
When I compile nginx that does not support tls1.3, it will report  errors

it show NGX_SSL_TLSv1_3 not define.
 


zhuizhuhaomeng:
fixed in https://github.com/openresty/lua-nginx-module/commit/5777a36a93fe7520d958dd512a14d553e50b807b

PR #2351, opened 2024-08-21
feature: support udp large packet segmentation to send multiple times…

coffeei:
add new udp send method like this:
local ok, err = udpsock:sendbuf(str1,str2,str3,...,big_str,start,end)

This method can send udp message without copy or split sub string in Lua land. Usually, in the udp sending scenario, you may have to split the message and send for many times when the string needed to sent is too long, this method will copy all the string elements piece by piece to the underlying Nginx socket send buffers,the copy order is like this: str1,str2,str3,...,big_str[start,end],which usually may helps to improve your system performance.


README.markdown
```diff
@@ -3641,160 +3641,282 @@ The pool is global at Nginx worker level. And it is used to reuse Lua VMs betwee
...
-* [coroutine.status](#coroutinestatus)
-* [ngx.run_worker_thread](#ngxrun_worker_thread)
+- [Name](#name)
+- [Table of Contents](#table-of-contents)
+- [Status](#status)
+- [Version](#version)
+- [Videos](#videos)
+- [Synopsis](#synopsis)
+- [Description](#description)
+- [Typical Uses](#typical-uses)
```
zhuizhuhaomeng, line 3651: please don't change the format of the doc.

src/ngx_http_lua_socket_udp.c
```diff
@@ -961,6 +965,139 @@ ngx_http_lua_socket_udp_send(lua_State *L)
...
+    buf_len = size;
+    if (stack_size > 3 + stack_diff)
+    {
+        for (int i = (-1 * stack_size) + stack_diff; i <= -4; i++)
+        {
+            luaL_checklstring(L, i, &head_len);
+            size += head_len;
+        }
+    }
+    // copy msg
```
zhuizhuhaomeng, line 1052: you should use the string.buffer instead of implement this feature.

zhuizhuhaomeng (changes requested):

src/ngx_http_lua_socket_udp.c
coffeei: how to use  string.buffer  instead of implement this feature on  lua land ,can you give some example? the copy occurs on the c land,but i need to split the big string on the  lua land ,and also i need to make the header of the udp package on lua land. if i use the origin udpsock:send() method , sub string ( big_msg )  is inevitable ,.usually i can send big string 
 use udp like below，there is no need to make a new string  pass to c land.
```
for idx = 1, totalPackage do
        local header_binary = struct.pack(">i2c1c32i4c1", packageSize + 42, '1', msg_id, totalPackage, zip_type or '0')
        local serial_binary
        if totalPackage == 1 then
            serial_binary = struct.pack(">i4", 0)
        else
            msgEndIndex = msgStartIndex + packageSize - 1
            serial_binary = struct.pack(">i4", idx - 1)
        end
        ok, err = udpsock:sendbuf(header_binary, serial_binary, big_msg, msgStartIndex, msgEndIndex)
        msgStartIndex = msgEndIndex + 1
    end
```

README.markdown
coffeei: ok,i just know this ,sorry

src/ngx_http_lua_socket_udp.c
zhuizhuhaomeng: https://luajit.org/ext_buffer.html

zhuizhuhaomeng: gotcha

PR #2381, opened 2024-12-02
fix: support tcp binding ip:port or ip of ipv4 or ipv6

oyaya:
bugfix: make it possible to bind ip:port or ip from support only binding ip.

### DIFF
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -867,7 +867,7 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
 
     text = (u_char *) luaL_checklstring(L, 2, &len);
 
-    local = ngx_http_lua_parse_addr(L, text, len);
+    local = ngx_http_lua_parse_addr_port(L, text, len);
     if (local == NULL) {
         lua_pushnil(L);
         lua_pushfstring(L, "bad address");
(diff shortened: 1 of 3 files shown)


chensunny:
1. Which cases will use the` ip: port` ,It looks very strange.
2.It may cause a Breaking Change in some cases

Because, as a TCP client, the port will be assigned by the Linux kernel；



@oyaya

src/ngx_http_lua_socket_tcp.c
```diff
@@ -3430,7 +3430,7 @@ ngx_http_lua_socket_tcp_handler(ngx_event_t *ev)
 static ngx_int_t
 ngx_http_lua_socket_tcp_get_peer(ngx_peer_connection_t *pc, void *data)
 {
-    /* empty */
+    pc->type = SOCK_STREAM;
```
chensunny, line 3433: SOCK_STREAM  is by Default

[    type = (pc->type ? pc->type : SOCK_STREAM);](https://github.com/firebase/nginx/blob/4dbf1fbb2c8b229b881719a4438f28e99050171e/src/event/ngx_event_connect.c#L39)

src/ngx_http_lua_util.c
```diff
@@ -4555,4 +4555,59 @@ ngx_http_lua_parse_addr(lua_State *L, u_char *text, size_t len)
...
+    plen = last - p;
+
+    port = ngx_atoi(p, plen);
+    if (port < 1 || port > 65535) {
+        return NULL;
+    }
+
+    len -= plen + 1;
+
+	addr = ngx_http_lua_parse_addr(L, text, len);
```
chensunny, line 4602: Repeated parsing of addr from above

chensunny:
Just have a discussion:
Maybe  sock:bind(ip,port) is better for Breaking Change?

PR #2412, opened 2025-04-25
feature: support tcp binding ip:port or ip of ipv4 or ipv6

chensunny:
## Brackgroud

ref: https://github.com/openresty/lua-nginx-module/pull/2381

## Key Changes 

Original PR uses `sock:bind(IP:Port)` API will cause a Breaking Change.
New design uses `sock:bind(IP, Port)`  API.

To implement the function, it needs to pass 2 params in Lua VM;
Parse the params to make  `addr` struct  and inject into SOCKET_BIND_INDEX cache;
When users  use  `sock:connect` API , it  will get the `addr ` struct from SOCKET_BIND_INDEX cache and set into `u->peer.local` ;

Last will use the peer.local  the make peer connection is other nginx module;

## Test Coverage
This PR adds a new test case (TEST 7) to verify the behavior of upstream sockets when binding to a specific IP and port. The test ensures that the socket binding either succeeds with the expected IP and port (bind: 127.0.0.1:12345) or fails with the error message failed to connect: address already in use.

New Test Case: TEST 7 is added to validate socket binding with a specific IP and port.
Response Validation: The test checks for two possible outcomes in the response body:
Successful binding: bind: 127.0.0.1:12345
Failure due to address conflict: failed to connect: address already in use
Error Log Check: Ensures that the error log contains the expected message when binding fails.

The test case will run 4 times, Expect:
The first time will be successful,
Other times will fail to bind because the TCP connection is still in TIME-WAIT, which the port is still be used.

Why design the case above?
I was trying to use https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksetoption
to set SO_REUSEADDR  to the bind, but the setting is for the connection that will be ESTABLISHED.
But to reuse the TIME-WAIT port, it needs to be set before binding, which can not inject nginx code before bind now;


src/ngx_http_lua_socket_tcp.c
```diff
@@ -843,11 +843,13 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
     u_char               *text;
     size_t                len;
     ngx_addr_t           *local;
+    int                   port;
 
     n = lua_gettop(L);
 
-    if (n != 2) {
-        return luaL_error(L, "expecting 2 arguments, but got %d",
+    // 修正参数检查，允许2或3个参数
```
zhuizhuhaomeng, line 850: Please use English

zhuizhuhaomeng: Please use the C style comment /* */

src/ngx_http_lua_socket_tcp.c
```diff
@@ -865,9 +867,26 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
 
     luaL_checktype(L, 1, LUA_TTABLE);
 
+    port = 0;
+    /* most popular suit: host:port */
```
zhuizhuhaomeng, line 871: In the most cases, we didn't specify the port number.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -865,9 +867,26 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
...
+            lua_pushnil(L);
+            lua_pushfstring(L, "bad port number: %d", port);
+            return 2;
+        }
+    } 
+
     text = (u_char *) luaL_checklstring(L, 2, &len);
 
-    local = ngx_http_lua_parse_addr(L, text, len);
+    local = ngx_http_lua_parse_addr_port(L, text, len, port);
```
zhuizhuhaomeng, line 889: This function is wired.
Actually, it does not parse the port.

src/ngx_http_lua_socket_tcp.c
```diff
@@ -876,7 +895,7 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
 
     /* TODO: we may reuse the userdata here */
     lua_rawseti(L, 1, SOCKET_BIND_INDEX);
-
+
```
zhuizhuhaomeng, line 898: extra blank

src/ngx_http_lua_socket_tcp.c
```diff
@@ -1128,7 +1147,11 @@ ngx_http_lua_socket_tcp_connect(lua_State *L)
     dd("lua peer connection log: %p", pc->log);
 
     lua_rawgeti(L, 1, SOCKET_BIND_INDEX);
-    local = lua_touserdata(L, -1);
+        local = lua_touserdata(L, -1);
```
zhuizhuhaomeng, line 1150: extra blanks

src/ngx_http_lua_socket_tcp.c
zhuizhuhaomeng: use the ngx-releng tool from [openresty-devel repo](https://github.com/openresty/openresty-devel-utils)

zhuizhuhaomeng (changes requested):

zhuizhuhaomeng:
> ## Brackgroud
> ref: #2381
> 
> ## Key Changes
> Original PR uses `sock:bind(IP:Port)` API will cause a Breaking Change. New design uses `sock:bind(IP, Port)` API.
> 
> To implement the function, it needs to pass 2 params in Lua VM; Parse the params to make `addr` struct and inject into SOCKET_BIND_INDEX cache; When users use `sock:connect` API , it will get the `addr ` struct from SOCKET_BIND_INDEX cache and set into `u->peer.local` ;
> 
> Last will use the peer.local the make peer connection is other nginx module;
> 
> ## Test Coverage
> This PR adds a new test case (TEST 7) to verify the behavior of upstream sockets when binding to a specific IP and port. The test ensures that the socket binding either succeeds with the expected IP and port (bind: 127.0.0.1:12345) or fails with the error message failed to connect: address already in use.
> 
> New Test Case: TEST 7 is added to validate socket binding with a specific IP and port. Response Validation: The test checks for two possible outcomes in the response body: Successful binding: bind: 127.0.0.1:12345 Failure due to address conflict: failed to connect: address already in use Error Log Check: Ensures that the error log contains the expected message when binding fails.
> 
> The test case will run 4 times, Expect: The first time will be successful, Other times will fail to bind because the TCP connection is still in TIME-WAIT, which the port is still be used.
> 
> Why design the case above? I was trying to use https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksetoption to set SO_REUSEADDR to the bind, but the setting is for the connection that will be ESTABLISHED. But to reuse the TIME-WAIT port, it needs to be set before binding, which can not inject nginx code before bind now;

 `sock:bind(IP:Port)`   does not case a breaking change.  The original  `sock:bind(IP)`  is also compatible.
But  `sock:bind(IP, Port)`  is more efficient indeed.

src/ngx_http_lua_socket_tcp.c
chensunny: use `ngx_http_lua_parse_address_with_port` is better?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -865,9 +867,26 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
 
     luaL_checktype(L, 1, LUA_TTABLE);
 
+    port = 0;
+    /* handle case: host:port */
+    if (n == 3 && lua_isnumber(L, 3)) {
```
zhuizhuhaomeng, line 872: what if n == 3 and it is not a number ?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -865,9 +867,26 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
...
+            lua_pushnil(L);
+            lua_pushfstring(L, "bad port number: %d", port);
+            return 2;
+        }
+    }
+
     text = (u_char *) luaL_checklstring(L, 2, &len);
 
-    local = ngx_http_lua_parse_addr(L, text, len);
+    local = ngx_http_lua_parse_address_with_port(L, text, len, port);
```
zhuizhuhaomeng, line 889: set the port outside this function.

zhuizhuhaomeng (changes requested):

chensunny:
Here are two ways:

- Uniformly process the param first, and then handle the logic
- Handle the logic while processing parameters 


I think following others ' hands, which was used first, is better?

src/ngx_http_lua_socket_tcp.c
```diff
@@ -865,6 +867,25 @@ ngx_http_lua_socket_tcp_bind(lua_State *L)
...
     luaL_checktype(L, 1, LUA_TTABLE);
 
+    port = 0;
+    /* handle case: host:port */
+    /* Hit the following parameter combination:
+     * sock:bind("127.0.0.1", port)     */
+    if (n == 3) {
+        if (!lua_isnumber(L, 3)) {
+            lua_pushnil(L);
+            lua_pushliteral(L, "bad port number: need int type");
```
zhuizhuhaomeng, line 877:
```suggestion
            lua_pushliteral(L, "expecting port to be a number but got %s", xxxx);
```

zhuizhuhaomeng (changes requested):

zhuizhuhaomeng:
> ## Brackgroud
> ref: #2381
> 
> ## Key Changes
> Original PR uses `sock:bind(IP:Port)` API will cause a Breaking Change. New design uses `sock:bind(IP, Port)` API.
> 
> To implement the function, it needs to pass 2 params in Lua VM; Parse the params to make `addr` struct and inject into SOCKET_BIND_INDEX cache; When users use `sock:connect` API , it will get the `addr ` struct from SOCKET_BIND_INDEX cache and set into `u->peer.local` ;
> 
> Last will use the peer.local the make peer connection is other nginx module;
> 
> ## Test Coverage
> This PR adds a new test case (TEST 7) to verify the behavior of upstream sockets when binding to a specific IP and port. The test ensures that the socket binding either succeeds with the expected IP and port (bind: 127.0.0.1:12345) or fails with the error message failed to connect: address already in use.
> 
> New Test Case: TEST 7 is added to validate socket binding with a specific IP and port. Response Validation: The test checks for two possible outcomes in the response body: Successful binding: bind: 127.0.0.1:12345 Failure due to address conflict: failed to connect: address already in use Error Log Check: Ensures that the error log contains the expected message when binding fails.
> 
> The test case will run 4 times, Expect: The first time will be successful, Other times will fail to bind because the TCP connection is still in TIME-WAIT, which the port is still be used.
> 
> Why design the case above? I was trying to use https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#tcpsocksetoption to set SO_REUSEADDR to the bind, but the setting is for the connection that will be ESTABLISHED. But to reuse the TIME-WAIT port, it needs to be set before binding, which can not inject nginx code before bind now;

Please change the DOC of lua-resty-core.

PR #2416, opened 2025-05-08
feature: use the actual ngx_connection_t * in ssl_client_hello_by

climagabriel:
Do we really need to create a fake connection here? Sure, the request struct wasn't even created yet but the conneciton does exist.
I personally need it because I need to allocate memory that will live for as long as the connection exists, not until the Lua chunk returns.

image

Also, surely there must be a better way to do this.
This line already does exist at the very top of ngx_http_lua_ssl_client_hello_handler().

If the real connection exists, can't we make it available in the Lua code itself?
The same way we make the fake request available by resty.core.base.get_request

### DIFF
--- a/src/ngx_http_lua_ssl_client_helloby.c
+++ b/src/ngx_http_lua_ssl_client_helloby.c
@@ -670,8 +670,7 @@ ngx_http_lua_ffi_ssl_get_client_hello_ext_present(ngx_http_request_t *r,
     int                      got_extensions;
     size_t                   ext_len;
     int                     *ext_out;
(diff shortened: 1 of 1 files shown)


zhuizhuhaomeng (dismissed):

zhuizhuhaomeng (approved):

climagabriel:
If the real connection exists, can't we make it available in the Lua code itself. The same way we make the fake request available by resty.core.base.get_request

climagabriel:
I read a bit more about openresty and realized that data persisting beyond the lifetime of a request goes against the design principles so I'm closing this PR.

PR #1962, opened 2021-11-03
feature: ngx_http_lua_ffi_ssl_ciphers

bjne:
Add ngx_http_lua_ffi_ssl_ciphers that returns a uint16_t array of
tls_protocol_id supported (and enabled) by both server and client

consider this pull request a request for comment
example usage from resty-core:

```lua
ffi.cdef[[
int ngx_http_lua_ffi_ssl_ciphers(ngx_http_request_t *r, uint16_t *ciphers,
        uint16_t *nciphers, char **err);

typedef struct {
    uint16_t nciphers;
    uint16_t ciphers[?];
} ngx_lua_ssl_ciphers;
]]

do
    local tls_proto_id = {
        [0x1305] = {
            iana_name = "TLS_AES_128_CCM_8_SHA256",
            tls_version = 1.3,
            kex = "none",
            auth = "none",
            enc = "AES 128 CCM 8",
            hash = "SHA256"
        },
        ...
   }

      local unknown_cipher = {
            iana_name = "UNKNOWN",
            tls_version = 0,
            kex = "UNKNOWN",
            auth = "UNKNOWN",
            enc = "UNKNOWN",
            hash = "UNKNOWN"
    }

    setmetatable(tls_proto_id, {
        __index = function(t,k)
            t[k] = unknown_cipher

            return unknown_cipher
        end
    })

    local iterate_ciphers = function(ciphers, n)
        if ciphers.nciphers > n then
            return n+1, tls_proto_id[ciphers.ciphers[n]]
        end
    end

    local ciphers_t = {}

    ffi.metatype('ngx_lua_ssl_ciphers', {
        __ipairs = function(ciphers)
            return iterate_ciphers, ciphers, 0
        end,
        __tostring = function(ciphers)
            for n,c in ipairs(ciphers) do
                ciphers_t[n] = type(c) == "table" and c.iana_name or format("0x%.4x", c)
            end

            return concat(ciphers_t, ":", 1, ciphers.nciphers)
        end
    })
end

local ciphers_typ = ffi.typeof("ngx_lua_ssl_ciphers")


local ciphers_buf = ffi.new("uint16_t [?]", 256)

_M.ciphers = function()
    local r = get_request()
    if not r then
        ngx.log(ngx.ERR, "no request found")
    end

    ciphers_buf[0] = 255
    local rc = C.ngx_http_lua_ffi_ssl_ciphers(r, ciphers_buf+1, ciphers_buf, errmsg)
    if rc ~= FFI_OK then
[...]

PR #2419, opened 2025-05-08
feature: ngx_http_lua_ffi_ssl_get_client_hello_ciphers()

climagabriel:

https://github.com/openresty/lua-resty-core/pull/498


chensunny:
The same like https://github.com/openresty/lua-nginx-module/pull/1962

climagabriel:
> The same like #1962

Partially. He uses a different openssl api calls under the hood, and his intention is to fetch the client and server ciphers from the session once it's established. 
I aim merely to fetch the ciphers offered by the client in the clienthello message, before the SSL session has even been established.

PR #2432, opened 2025-06-28
bugfix: improve HTTP/3 SSL Lua callback yield handling

swananan:

https://github.com/openresty/openresty/pull/1063
https://github.com/openresty/openresty-devel-utils/pull/58

### Problem Description
When using HTTP/3, yield Lua APIs (such as ngx.sleep, cosocket, ngx.thread.*, etc.) in ssl_certificate_by_lua* and ssl_client_hello_by_lua* directives cannot work properly, causing HTTP/3 request connections to be abnormally disconnected.

### Root Cause
HTTP/3 in nginx is implemented through QUIC code, and its SSL handling mechanism is significantly different from traditional HTTP/1.x and HTTP/2. 

### Solution
Currently, nginx QUIC SSL has three implementation approaches:

1. OpenSSL traditional interface compatibility mode (target of this fix)
2. BoringSSL implementation
3. OpenSSL 3.5+ official external QUIC library interface

Since OpenResty has not yet been upgraded to adapt to the latest nginx version, currently only the first compatibility mode is supported by default.

This fix specifically targets the first approach - the OpenSSL traditional interface compatibility mode. The implementation requires some simple modifications to nginx to ensure that nginx does not treat callback yield errors during ssl_do_handshake as fatal errors https://github.com/openresty/openresty/pull/1063
 . Additionally, the lua-nginx-module needs to explicitly call ssl_do_handshake after successful Lua script execution to further advance the QUIC handshake interaction.

### TODO:

1. Future nginx version adaptation: Support for OpenSSL 3.5 latest standard interface when adapting to the newest nginx versions
2. Session fetch limitation: This fix currently does not include ssl_session_fetch_by_lua Lua functionality, as nginx QUIC implemented through the first approach (OpenSSL traditional interface) does not support Zero RTT, making the fix unnecessary at this time. However, future support will be considered.
3. OpenResty has incomplete support for BoringSSL, but adding full support may be considered in the future.

### DIFF
--- a/src/ngx_http_lua_util.c
+++ b/src/ngx_http_lua_util.c
@@ -49,6 +49,10 @@
 #if (NGX_THREADS)
 #include "ngx_http_lua_worker_thread.h"
 #endif
+#if (NGX_HTTP_V3)
+#include <ngx_event_quic.h>
+#include <ngx_event_quic_connection.h>
+#endif
 
 
 #if 1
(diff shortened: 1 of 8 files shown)


oowl (approved):

src/ngx_http_lua_util.c
```diff
@@ -4561,4 +4564,34 @@ ngx_http_lua_ffi_bypass_if_checks(ngx_http_request_t *r)
     r->disable_not_modified = 1;
 }
 
+
+#if (NGX_HTTP_V3)
+void
+ngx_http_lua_resume_quic_ssl_handshake(ngx_connection_t *c)
```
oowl, line 4570: Do we need to comment here to describe why quic need extra resume SSL_do_handshake() function call?

zhuizhuhaomeng (changes requested):

Issue #2253, opened 2023-11-09
why not support lua phase in pre_content phase


akf00000:
I have som lua function need to be call after access phase，if I implement it in the last of access phase functions，the subrequest will not take effect, why not support lua phase in pre_content phase?

HanadaLee:
> would you please give us an detailed example?

I think the operations of getting upstream peers can be performed in the pre_content phase (if pre_content can support consocket).

Since balancer_by_lua_block does not support consocket, getting upstream peers from the outside is usually performed in the access_by_lua phase.

However, since the access phase is mainly responsible for access control, many subrequests usually skip the access phase, resulting in subrequests being unable to get upstream peers in the access phase.

Currently, I have solved this problem by replacing access_by_lua with the logic of getting upstream peers in rewrite_by_lua, but it is not an elegant solution after all.

In nginx, this phase is mainly used for try_files and mirrors. If possible, I hope to implement lua hook in the pre_content phase, i.e. pre_content_by_lua_xxx.

PR #2472, opened 2026-01-11
feature: add precontent_by_lua directives

HanadaLee:

Since balancer_by_lua_block does not support consocket, getting upstream peers from the outside is usually performed in the access_by_lua_*. (refer to https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#balancer_by_lua_block)

However, since the access phase is mainly responsible for access control, many subrequests usually skip the access phase, resulting in subrequests being unable to get upstream peers in the access phase.  The current solution is to identify sub-requests and then move the relevant processing logic forward to rewrite_by_lua_*. But it is ultimately an unreasonable solution, as it inappropriately utilizes the rewrite or access phase to provide functionality that does not belong to them.

In nginx, there is actually a precontent phase after the post_access phase. Initially, nginx named this phase `NGX_HTTP_TRY_FILES_PHASE`, and it was a dedicated phase for the try_files module (other module handlers could not be registered). As nginx's functionality expanded, it was renamed `NGX_HTTP_PRECONTENT_PHASE`, which also allowed the registration of new module handlers.

By injecting Lua code into the precontent phase, we can implement proxy/upstream-related preparation functions, including upstream header rewriting, proxy authentication, proxy URI rewriting, and upstream information retrieval (such as obtaining the upstream peer list via DNS). Furthermore, the precontent phase will not be bypassed in sub-requests (one of the biggest problems with registering non-access control related functions in the access phase).

This PR added 3 directives:

```
precontent_by_lua_block
precontent_by_lua_file
precontent_by_lua_no_postpone
```

The PR for lua-resty-core: https://github.com/openresty/lua-resty-core/pull/519

close https://github.com/openresty/lua-nginx-module/issues/2253

### DIFF
--- a/t/189-precontent.t
+++ b/t/189-precontent.t
@@ -19,7 +19,7 @@ our $StapScript = $t::StapThread::StapScript;
 
 repeat_each(2);
 
-plan tests => repeat_each() * (blocks() * 3 + 10);
+plan tests => repeat_each() * (blocks() * 3 + 6);
 
 #log_level("info");
 #no_long_string();
@@ -32,7 +32,7 @@ __DATA__
 --- config
     location /lua {
         precontent_by_lua_block {
-            ngx.log(ngx.ERR, "precontent_by_lua_block executed")
+            ngx.log(ngx.INFO, "precontent_by_lua_block executed")
         }
         content_by_lua_block {
             ngx.say("content phase executed")
@@ -53,7 +53,7 @@ precontent_by_lua_block executed
 --- config
     location /lua {
         precontent_by_lua_block {
-            ngx.log(ngx.ERR, "precontent phase executed")
+            ngx.log(ngx.INFO, "precontent phase executed")
             ngx.exit(ngx.OK)
         }
         content_by_lua_block {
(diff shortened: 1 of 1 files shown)


src/ngx_http_lua_precontentby.c
```diff
@@ -0,0 +1,412 @@
+
+/*
+ * Copyright (C) Xiaozhe Wang (chaoslawful)
+ * Copyright (C) Yichun Zhang (agentzh)
```
zhuizhuhaomeng: change to your email

src/ngx_http_lua_precontentby.c
```diff
@@ -0,0 +1,412 @@
...
+        cur_ph = &ph[r->phase_handler];
+        last_ph = &ph[cur_ph->next - 1];
+
+        dd("ph cur: %d, ph next: %d", (int) r->phase_handler,
+           (int) (cur_ph->next - 2));
+
+#if 0
+        if (cur_ph == last_ph) {
+            dd("XXX our handler is already the last precontent phase handler");
+        }
```
zhuizhuhaomeng: don't keep the dead code copy from other file

PR #2477, opened 2026-03-06
feature: callback on async push messages recieved on socket in the pool

deem0n:

Current implementation forcefully closes opened connection in the pool in case such connection receives "unexpected" incoming packet from the over side.

There are some use cases when we need to process such messages and keep socket open. For example, NATS (https://nats.io) protocol has health heartbeat (ping/pong) requirement to keep peers connected. If openresty will open connection to the NATS server and will not reply on async PING message, NATS will close unhealthy client.

Also we may want to subscribe to NATS, Redis and other pub/sub servers to receive async incoming messages.

This patch allows such processing with new config option `on_push` where we need to provide callback function. Here is working example with callback  implementation:

```lua
    local sock = ngx.socket.tcp()
    sock:settimeout(3000)

    local ok, err = sock:connect("nats", 4222, {
        on_push = function(data)
            if data == "PING\r\n" then
                return "PONG\r\n", true
            end
            return nil, false
        end
    })
```

This patch is fully backward compatible with old behavior, so if you do not provide on_push handler to the conect() function then connection will be closed on unexpected incoming packets, as before.

### Regression tests

Added 11 Test::Nginx integration cases with a Python standard-library TCP
fixture. The fixture binds an OS-assigned loopback port and uses a separate
control connection to send data after Lua has returned the socket to nginx's
keepalive pool. No NATS installation or other external service is needed.

The cases cover exact binary data, PING/PONG replies, single-byte input,
fragmented messages, pushes larger than the read buffer, keeping or closing
the connection, callback errors, EOF, unchanged behavior without `on_push`,
callback replacement/collection, and callbacks after request completion.
Reuse checks also verify bidirectional IO on the original TCP connection.

These tests exposed and now cover fixes for truncated callback data,
single-byte pushes closing the socket, incomplete draining of available data,
and old callback references retained after pool checkout. Callbacks receive
arbitrary TCP chunks; application message framing belongs in the callback.

Validation was completed on the separate Luxms `test/on-push-tcp` branch
before updating this PR: [four passing Linux CI jobs](https://github.com/luxms/lua-nginx-module/actions/runs/35450432685),
covering nginx 1.27.1 and 1.29.4 on Ubuntu 22.04 and 24.04. Each job runs the
new cases twice (110 assertions), plus the existing sanity/content tests
(390 assertions total per job).

Local reproduction instructions: [`doc/socket-on-push-tests.md`](https://github.com/luxms/lua-nginx-module/blob/25f5ca3924c7bdf6414b87f9b1e171b0b195bf16/doc/socket-on-push-tests.md).

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -64,8 +64,8 @@ Version
 =======
 
 This document describes ngx_lua
-[v0.10.28](https://github.com/openresty/lua-nginx-module/tags), which was released
-on 17 Jan, 2025.
+[v0.10.29](https://github.com/openresty/lua-nginx-module/tags), which was released
+on Oct 24, 2025.
 
 Videos
 ======
--- a/src/ngx_http_lua_socket_tcp.c
+++ b/src/ngx_http_lua_socket_tcp.c
@@ -6028,11 +6028,10 @@ ngx_http_lua_socket_keepalive_close_handler(ngx_event_t *ev)
     item = c->data;
 
     if (n > 0 && item->on_push_cb_ref != LUA_NOREF) {
-        char        rbuf[4096];
-        ssize_t     nread;
-        lua_State  *L;
-        int         close_conn;
-        int         nret;
+        unsigned char rbuf[4096];
+        ssize_t       nread;
+        lua_State     *L;
+        int           close_conn;
 
         ngx_log_debug0(NGX_LOG_DEBUG_HTTP, ev->log, 0,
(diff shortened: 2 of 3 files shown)


zhuizhuhaomeng:
awesome

deem0n:
@zhuizhuhaomeng Added tests and fixed some bugs!

PR #2520, opened 2026-08-27
bugfix: fix heap-use-after-free in receiveuntil after read error

waterWang:
Closes #1146.

`ngx_http_lua_socket_tcp_finalize_read_part` clears `cp->upstream`
but leaves `u->input_filter_ctx` pointing at the compiled pattern.
After a read error the iterator userdata is garbage-collected; its
`__gc` handler sees `cp->upstream == NULL` and does not clear
`u->input_filter_ctx`, leaving a dangling pointer. A later
`sock:close()` dereferences it in `ngx_http_lua_socket_tcp_finalize`,
causing a heap-use-after-free.

Clear `u->input_filter_ctx` on both sides of the relationship when
finalizing the read part, matching what `tcp_finalize` already does.

Reported-by: BrookPanZM


zhuizhuhaomeng:
@waterWang We need to add test cases to reproduce this issue althrough it seem harmness.

Issue #2517, opened 2026-08-11
heap-use-after-free in ngx_http_lua_socket_tcp_finalize after receiveuntil userdata GC


BrookPanZM:
I have report this issue on OpenResty. Here I paste the issue again. 

https://github.com/openresty/openresty/issues/1146

lilyonnamyran26:
Have you verified if the issue occurs consistently with specific versions of OpenResty or only under certain conditions in the `ngx_http_lua_socket_tcp_finalize` function?

PR #2504, opened 2026-05-26
bugfix: SIGSEGV in receiveuntil __gc on aborted multipart upload.

climagabriel:
read_error_retval_handler calls finalize_read_part directly when the receiveuntil iterator's recv errors. That clears u->buf_in but leaves cp->upstream live with cp->state > 0. Later GC fires cleanup_compiled_pattern -> read_prepare, which derefs the now-NULL u->buf_in.

Mirror tcp_finalize's cp->upstream = NULL detach so __gc's existing `if (u != NULL)` guard short-circuits.

Backtrace:

  ngx_http_lua_socket_tcp_read_prepare
  ngx_http_lua_socket_cleanup_compiled_pattern
  lj_BC_FUNCC
  gc_call_finalizer
  gc_finalize
  gc_onestep
  lj_gc_fullgc
  lua_gc
  lj_cf_collectgarbage
  lj_BC_FUNCC
  ngx_http_lua_run_thread
  ngx_http_lua_socket_tcp_resume_helper
  ngx_http_lua_access_handler
  ngx_http_core_access_phase
  ngx_http_core_run_phases
  ngx_http_lua_socket_tcp_read
  ngx_http_request_handler
  ngx_epoll_process_events
  ngx_process_events_and_timers
  ngx_worker_process_cycle
  ngx_spawn_process
  ngx_start_worker_processes
  ngx_master_process_cycle
  main

I hereby granted the copyright of the changes in this pull request
to the authors of this lua-nginx-module project.

PR #2526, opened 2026-09-16
tests: add an ASan job running the whole test suite.

u5surf:
Adds an AddressSanitizer job that runs the whole test suite. Following @zhuizhuhaomeng's note that we could run all the test cases, this covers everything rather than a socket subset.

The key part is building LuaJIT with `LUAJIT_USE_SYSMALLOC`: without it, LuaJIT's allocator never returns memory to the C allocator, so ASan cannot see use-after-free on Lua objects at all.

## What the job does

- Builds nginx with `NGX_BUILD_ASAN=1`, which `ngx-build` already supports, and LuaJIT with `LUAJIT_USE_SYSMALLOC`.
- Preloads the ASan runtime ahead of mockeagain. ASan refuses to start behind another library, and some test files (`t/065-tcp-socket-timeout.t`, `t/131-duplex-req-socket.t`) put mockeagain in `LD_PRELOAD` themselves.
- Allows a longer `TEST_NGINX_TIMEOUT` than the other jobs, since ASan slows every request down.
- Dumps `asan-logs/` on failure. An ASan abort otherwise surfaces only as a connection error from the test harness, which is hard to read.

## It finds real defects

Before #2525 landed, this job reported a heap-use-after-free in `t/067-req-socket.t` TEST 20 — the existing regression test from #2504 — on unmodified master:

```
==nginx==ERROR: AddressSanitizer: heap-use-after-free
WRITE of size 8
    #0 ngx_http_lua_socket_tcp_finalize  src/ngx_http_lua_socket_tcp.c:4592
```

That was #2517. With the fix merged, the job is clean.


zhuizhuhaomeng:
We clould run all the test cases.

.github/workflows/build_and_test.yml
```diff
@@ -271,7 +286,39 @@ jobs:
...
+          # ASan runtime trips ASan's initial library order check.
+          mkdir -p asan-logs
+          export ASAN_OPTIONS=detect_leaks=0,log_path=$PWD/asan-logs/asan,log_exe_name=true
+          python3 ./util/nc_server.py &
+          # The socket tests, which is where the Lua GC and the request
+          # lifetime interact. Running the whole suite under ASan takes too long.
+          # t/065-tcp-socket-timeout.t and t/131-duplex-req-socket.t are left out:
+          # they preload mockeagain themselves, which ASan's library order check
+          # rejects.
+          /usr/bin/env perl $(command -v prove) -I. -Itest-nginx/inc -Itest-nginx/lib \
```
zhuizhuhaomeng: We can skip the failed test like  t/065-tcp-socket-timeout.t and t/131-duplex-req-socket.t.
Running the whole suite is still needed if we want to run asan mode.

u5surf:
@zhuizhuhaomeng Done — the job now runs the whole suite.

- 231 files, 25,054 tests, no ASan reports.
- `t/146-malloc-trim.t` is the one file left out: ASan replaces the allocator, so `malloc_trim()` returns 0 there.
- The ASan runtime is preloaded ahead of mockeagain, so the files that preload it themselves run too.
- `TEST_NGINX_TIMEOUT` is higher for this job than for the others.
- The job takes about 25 minutes, against 17–19 for the other jobs.

The remaining red job is `t/189-http2-subreq-error-wakeup.t`, the flake #2523 addresses; it is unrelated to this change.

